PR #108, opened 2026-03-24
Test InMemoryEntityStore: lifecycle, queries, relationships

DominicBurkart:
## Summary

- Adds 22 new unit tests for `InMemoryEntityStore` in `harness/src/entities/mod.rs`, which previously had only 3 trivial tests that never exercised any `EntityStore` trait methods
- Covers the full entity lifecycle (`store` → `exists` → `update` → `delete`), error paths (duplicate ID rejection, not-found on update/delete), query filtering (by entity type, tags, time range, case-insensitive text search, limit, relevance scoring), and relationship management (create, bidirectional retrieval, type-specific deletion)
- Uses concrete `GitRepository` and `FileEntity` types to exercise real serialization paths through the query engine

## Motivation

`InMemoryEntityStore` contains the core entity management logic — store CRUD, multi-dimensional query filtering, and a relationship graph — but had zero test coverage of these code paths. This is the primary in-memory backing store used in development and testing, so its invariants should be continuously validated.

## Key invariants now under test

| Category | Invariant |
|----------|-----------|
| Store lifecycle | `store()` makes entity findable via `exists()`; `delete()` reverses this |
| Uniqueness | `store()` rejects duplicate entity IDs with `AlreadyExists` |
| Update guards | `update()` on nonexistent ID returns `NotFound` |
| Query: type filter | Filtering by `EntityType` returns only matching entities |
| Query: tag filter | Only entities with at least one matching tag are returned |
| Query: time range | Entities outside the time window are excluded |
| Query: text search | Case-insensitive substring match on serialized JSON; relevance = 0.8 |
| Query: limit | Result set is truncated to the requested limit |
| Query: relevance | No text query yields relevance 1.0 |
| Relationships | Creation requires both endpoints to exist; retrieval is bidirectional |
| Relationship deletion | Only the specified relationship type is removed |

## Test plan

- [x] All 25 entity tests pass locally (`cargo test --lib -p harness entities::tests`)
- [x] Full workspace unit tests pass (`cargo test --workspace --lib`)
- [x] `cargo fmt` and `cargo clippy` clean
- [ ] CI passes (patch-level and project-level coverage)

https://claude.ai/code/session_019ye7KRxjwCHc43FfQne9HF

### DIFF
--- a/harness/src/entities/mod.rs
+++ b/harness/src/entities/mod.rs
@@ -484,6 +484,8 @@ impl EntityStore for InMemoryEntityStore {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::entities::ast::types::{FileEntity, FileType};
+    use crate::entities::git::types::GitRepository;
 
     #[test]
     fn test_entity_metadata_creation() {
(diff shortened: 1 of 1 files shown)


DominicBurkart:
Closing: superseded by #113, which has more tests (30 vs 22) in a cleaner structure (dedicated test file with helper functions). Existing review flagged weak assertions and missing coverage here.

---

PR #111, opened 2026-03-27
Refactor SharedModelPool with RAII guards and improve container health checks

DominicBurkart:
## Summary
This PR refactors the `SharedModelPool` to use RAII-style `ModelGuard` handles instead of manual reference counting, adds proper HTTP health checks for container startup, and improves the MCP server's lifecycle management.

### Container Management
- **Health check improvements**: Added `health_check_host()` function that polls Ollama's HTTP API (`/api/tags`) instead of just sleeping, with configurable timeout and exponential backoff
- **URL parsing**: Implemented `parse_ollama_url()` in `model/src/ollama.rs` to properly extract host and port from Ollama URLs, with support for stripping `/v1` suffix
- **Host URL helper**: Added `ContainerConfig::ollama_host_url()` method to generate the correct Ollama endpoint based on port mappings

### SharedModelPool Refactoring
- **RAII guards**: Replaced manual `release()` calls with `ModelGuard` that automatically decrements ref count on drop
- **Simplified API**: `get_or_start()` now returns `ModelGuard` directly instead of requiring separate `release()` calls
- **Serialized startup**: Added `tokio::sync::Mutex` to ensure only one task starts the container when multiple concurrent requests arrive
- **Removed atomic operations**: Replaced `AtomicUsize` with standard `Mutex<PoolInner>` for simpler, safer reference counting
- **Test provider support**: Added `new_with_provider()` constructor for testing without actual container startup

### MCP Server Updates
- **Guard ownership**: `NannaMcpServer` now holds the `ModelGuard` to keep the container alive for the server's lifetime
- **Flexible I/O**: Refactored `run_stdio()` to use new `run_on()` method accepting generic `AsyncRead`/`AsyncWrite` for better testability
- **Container lifecycle**: Added `--host-port` and `--ollama-url` CLI flags to `mcp-serve` command for flexible deployment options

### Testing
- Added comprehensive unit tests for `ollama_host_url()`, `parse_ollama_url()`, and `ModelGuard` lifecycle
- Added integration tests for concurrent `get_or_start()` calls and guard cleanup
- Added stdio round-trip tests for MCP protocol with duplex I/O
- Added ignored integration test for real container lifecycle with `SharedModelPool`
- Updated CI workflow to properly start Ollama container before running ignored tests

## Implementation Details
- Health checks use a 2-second polling interval with remaining timeout clamping to avoid overshooting
- The `start_lock` in `SharedModelPool` prevents thundering herd during concurrent startup attempts
- `ModelGuard` implements `Drop` to safely decrement ref count and clean up resources when the last guard is dropped
- URL parsing handles both explicit ports and defaults to 11434 for Ollama

https://claude.ai/code/session_01UuzdDpdt21rFrvzV2gi4Gb

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,34 @@ nix develop
 nix build
 ```
 
+### LLM Setup (Ollama)
+
+The agent requires a running [Ollama](https://ollama.ai/) instance with a model installed:
+
+```bash
+# Install Ollama (see https://ollama.ai/download)
+curl -fsSL https://ollama.ai/install.sh | sh
+
+# Pull the default model
+ollama pull qwen3:0.6b
+
+# Verify Ollama is running
+nix develop --command cargo run --bin harness -- health
+```
+
+### Running the Agent
+
+```bash
+# Enter development environment
+nix develop
+
+# Run the agent with tools enabled (recommended)
+cargo run --bin harness -- agent --prompt "Your task description" --tools
+
+# Run with a specific model and verbose output
+cargo run --bin harness -- agent --prompt "Your task" --model qwen3:0.6b --tools --verbose
+```
+
 ### Using Cachix (Optional but Recommended)
 
 Cachix provides a public binary cache for faster builds. No account needed to pull pre-built artifacts.
--- a/harness/src/agent/mod.rs
+++ b/harness/src/agent/mod.rs
@@ -280,6 +280,7 @@ pub struct AgentLoop {
     tool_registry: Option<ToolRegistry>,
     conversation_history: Vec<ChatMessage>,
     progress_counter: Option<Arc<AtomicUsize>>,
+    state_history: Vec<AgentState>,
 }
 
 impl AgentLoop {
@@ -296,6 +297,7 @@ impl AgentLoop {
             tool_registry: None,
             conversation_history: Vec::new(),
             progress_counter: None,
+            state_history: Vec::new(),
         }
     }
 
@@ -312,6 +314,7 @@ impl AgentLoop {
             tool_registry: None,
             conversation_history: Vec::new(),
             progress_counter: None,
+            state_history: Vec::new(),
         }
     }
 
@@ -332,6 +335,7 @@ impl AgentLoop {
             tool_registry: None,
             conversation_history: Vec::new(),
             progress_counter: None,
+            state_history: Vec::new(),
         }
     }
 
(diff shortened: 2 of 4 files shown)


harness/src/container.rs
```diff
@@ -581,80 +625,146 @@ pub fn exec_in_container(
...
+/// Dropping this guard decrements the pool's reference count and stops the
+/// container when no more guards exist.
+pub struct ModelGuard {
+    pool: Arc<SharedModelPool>,
+    provider: Arc<dyn ModelProvider>,
+}
+
+impl ModelGuard {
+    pub fn provider(&self) -> &Arc<dyn ModelProvider> {
+        &self.provider
```
DominicBurkart: **Potential issue: `unwrap()` on `Mutex::lock()` in `Drop`.** If any code panics while holding `self.pool.inner`, the mutex becomes poisoned and this `unwrap()` will panic during `Drop`, which is a double-panic and causes an abort. Consider using `lock().unwrap_or_else(|e| e.into_inner())` to recover from poisoning in the `Drop` impl, since decrementing the ref count is important for cleanup even after a panic.

---

.github/workflows/ci.yml
```diff
@@ -147,11 +161,11 @@ jobs:
       shell: bash
       run: |
         nix develop --command cargo nextest run --workspace --test '*' --all-features --test-threads=1
-        if curl -s --max-time 5 http://localhost:11434/api/tags > /dev/null 2>&1; then
-          nix develop --command cargo nextest run --workspace --test '*' --all-features --run-ignored ignored-only --test-threads=1
-        else
-          echo "Ollama not reachable on port 11434, skipping model integration tests"
-        fi
+        nix develop --command cargo nextest run --workspace --test '*' --all-features --run-ignored ignored-only --test-threads=1
+
```
DominicBurkart: **CI failure -- `integration-container (ubuntu-latest)` is failing.** This step unconditionally runs ignored tests now (removed the `if curl ...` guard), but the CI Ollama container is mapped to port 11434 while the integration test `test_shared_model_pool_lifecycle_real_container` uses port 11436. If the ignored test tries to start its own container on 11436 but the image or runtime isn't available in CI, this will fail.

The old code gracefully skipped ignored tests when Ollama was unreachable. The new approach is correct in principle (start Ollama explicitly, then run all tests), but you need to ensure the ignored tests' port expectations match the CI setup, or that the tests can discover the correct port. This is likely the cause of the current CI failure.

---

PR #112, opened 2026-03-27
Add unit tests for InMemoryEntityStore, model types, and image validation

DominicBurkart:
## Summary

- **`harness/src/entities/store_tests.rs`** -- 22 new tests for `InMemoryEntityStore`, covering the full CRUD lifecycle (`store`, `exists`, `update`, `delete`), all query filter dimensions (entity type, tags, time range, text search, limit, combined filters), relationship management (create, get from both sides, delete, selective delete by type), and error paths (duplicate store, update/delete of nonexistent entities, relationship with missing entity).

- **`model/src/types_extended_tests.rs`** -- 18 new tests for `model::types`, covering `ChatMessage::assistant_with_tools`, `ChatRequest::with_tools` (verifying it sets `ToolChoice::Auto`), serde round-trips for `ToolChoice` (all four variants), `MessageRole`, `FinishReason`, `SchemaType`, `ToolCall`/`FunctionCall`, `ChatResponse` with `Usage`, `ModelInfo` with and without optional fields, `ToolDefinition` with nested `PropertySchema`, and chained builder usage.

- **`image-builder/src/validate_tests.rs`** -- 5 new tests for `validate_image` edge cases: non-empty directory, empty directory, non-JSON file content, empty file, and JSON array file (first byte `[` vs `{`).

## Motivation

The `InMemoryEntityStore` is core domain infrastructure with complex query logic (type filtering, tag filtering, time range filtering, case-insensitive text search, relevance scoring, limit truncation, relevance sorting) and relationship management, yet had only a single trivial test checking initial emptiness. The model types had gaps around builder methods and serde correctness for key enums. The image validator had untested branches for directory and non-JSON-file inputs.

## Key invariants validated

- Store rejects duplicate entity IDs; update/delete reject missing IDs
- Query filters compose correctly (type + tags)
- Text search is case-insensitive and assigns 0.8 relevance vs 1.0 for unfiltered
- Time range filtering uses inclusive bounds
- Limit truncates after sorting
- Relationships are bidirectionally queryable and selectively deletable by type
- `ToolChoice::Auto` is the default and is set by `with_tools()`
- All serde-derived types round-trip correctly through JSON
- `validate_image` returns false for empty files/dirs and non-JSON content

## Test plan

- [ ] CI passes `cargo nextest run --workspace --lib`
- [ ] No regressions in existing tests
- [ ] All new tests pass locally with `cargo test`

https://claude.ai/code/session_01NuHBW36UvTi7akwASQ5ZcH

PR #113, opened 2026-03-28
test: comprehensive unit tests for InMemoryEntityStore

DominicBurkart:
## Summary\n\nAdds 30 focused unit tests for `InMemoryEntityStore` in `harness/src/entities/store_tests.rs`, exercising every `EntityStore` trait method and query filter path.\n\n## Why this component?\n\n`InMemoryEntityStore` is a core component used by the agent loop for entity management, yet it had **only one test** (`test_in_memory_store_basic_operations`) that verified the store starts empty. All CRUD operations, query filtering, and relationship management were entirely untested despite containing ~150 lines of non-trivial logic (type filtering, tag matching, case-insensitive text search, time range filtering, relevance scoring, limit truncation, relationship referential integrity checks).\n\n## Current test disposition\n\n- **Before**: 1 trivial test (`assert_eq!(store.entities.len(), 0)`)\n- **After**: 31 tests covering every code path through the `EntityStore` implementation\n\n## Key invariants being validated\n\n- **Store**: returns correct entity ID; rejects duplicate IDs with `AlreadyExists`\n- **Exists**: true after store, false for unknown IDs\n- **Update**: replaces entity content; fails with `NotFound` for missing entities\n- **Delete**: removes entity; fails with `NotFound` for missing entities\n- **Query by entity type**: filters correctly; excludes non-matching types\n- **Query by tags**: matches entities with at least one overlapping tag\n- **Query by text**: case-insensitive substring match on serialized JSON; returns relevance 0.8; empty result for non-matches\n- **Query by time range**: includes entities within range; excludes entities outside\n- **Query limit**: truncates results to requested count\n- **Query combined filters**: type + tag filters compose correctly\n- **Query default**: returns all entities with relevance 1.0\n- **Relationships**: create/get bidirectional lookup; referential integrity (rejects nonexistent source or target); delete by type preserves other relationship types; empty result for unknown entity\n- **Metadata**: stable IDs, version initialized to 1, timestamps set on creation\n- **Enum coverage**: all `RelationshipType` variants are distinct; `Custom` holds its value\n- **Default impls**: `InMemoryEntityStore::default()` and `EntityQuery::default()` behave as expected\n\n## Test design approach\n\n- Uses `ContextEntity` as a lightweight concrete `Entity` implementation (already available, no new dependencies)\n- Each test is self-contained with its own store instance (no shared mutable state)\n- Async tests use `#[tokio::test]` matching the project's existing pattern\n- Helper functions (`make_entity`, `make_tagged_entity`) reduce boilerplate while keeping tests readable\n- No external service dependencies -- all tests run purely in-memory"

### DIFF
--- a/harness/src/entities/store_tests.rs
+++ b/harness/src/entities/store_tests.rs
@@ -524,7 +524,7 @@ mod tests {
 
     #[test]
     fn all_relationship_types_are_distinct() {
-        let types = vec![
+        let types = [
             RelationshipType::Contains,
             RelationshipType::Calls,
             RelationshipType::Imports,


harness/src/entities/store_tests.rs
```diff
@@ -0,0 +1,564 @@
...
+            .await
+            .unwrap();
+
+        let rels = store.get_relationships(&id_a).await.unwrap();
+        assert_eq!(rels.len(), 1);
+        assert_eq!(rels[0].relationship_type, RelationshipType::References);
+    }
+
+    #[tokio::test]
+    async fn get_relationships_for_unknown_entity_returns_empty() {
```
DominicBurkart: Good test -- this is the only PR of the three that validates that `delete_relationship` only removes the matching type while preserving other relationship types between the same entities. PR #117 also tests this, but this version is slightly more thorough with the assertion on the surviving relationship's type.

---

DominicBurkart (commented):
## Review: test: comprehensive unit tests for InMemoryEntityStore

### CI Status: All green (17/17 checks pass)

### Quality

Well-structured tests with good coverage. Helper functions (`make_entity`, `make_tagged_entity`) keep tests readable. The test for selective relationship deletion by type is particularly good. Case-insensitive text search and relevance score assertions pin down important invariants.

### Duplicate coverage problem

This PR overlaps heavily with PR #112 (same file path, same test patterns) and PR #117 (different path but same entity store coverage plus model types and judge scoring).

Since this PR is entity-store-only and #117 is a strict superset (entity store + model types + judge scoring), recommend closing this in favor of #117.

### If this PR is chosen instead

- The tests are well-written and would be ready to merge after a rebase onto main (`mergeable_state: behind`).
- The `#[cfg(test)] mod tests` double-wrapping (the file is already `#[cfg(test)]` gated in `mod.rs`) is harmless but redundant.

---

Issue #20, opened 2025-10-05
Entity Management


DominicBurkart:
Now that we have the global project structure in place, we can begin with the meaningful implementation. Essentially all of the domain complexity of Nanna is in its entity management approach­: what it collects and surfaces as context to the model, and the structure it provides for the model to understand and ask about the relationships between the pieces of information it shows. 

The objective for this issue is to split and track the different domains of implementation necessary for entity management.

# 1. Document the nanna architecture (ARCHITECTURE.md, linked in README.md and AGENTS.md): 

## Harness control flow
```mermaid
---
config:
  theme: redux-dark
  layout: dagre
---
flowchart TD
    A(["Application State 1"]) --> n6["Entity Enrichment"]
    n10(["User Prompt"]) --> n4["Plan Entity Modification"]
    B{"Task Complete?"} --> C["Yes"] & D["No"]
    D --> n1["Entity Modification Decision"]
    n1 --> n3["Query Entities (RAG)"] & n4
    n4 --> n7["Perform Entity Modification"]
    C --> n9(["Application State 2"])
    n3 --> n1
    n7 --> n11["Update Entities"]
    n11 --> B
    n6 --> n4
    n6@{ shape: rect}
    n4@{ shape: rect}
    n1@{ shape: diam}
    n3@{ shape: rect}
    n7@{ shape: rect}
    n11@{ shape: rect}
     A:::Rose
     A:::Aqua
     n10:::Aqua
     n9:::Aqua
    classDef Rose stroke-width:1px, stroke-dasharray:none, stroke:#FF5978, fill:#FFDFE5, color:#8E2236
    classDef Aqua stroke-width:1px, stroke-dasharray:none, stroke:#46EDC8, fill:#DEFFF8, color:#378E7A
```

## Containers Topology
```mermaid
---
config:
  theme: redux-dark
  layout: elk
---
flowchart TD
    B(["Harness Container"]) -- Modifies --> C(["Dev Container"])
    B -- Queries --> n1(["Model"])
    C -- Can compile binary for --> n2(["Sandbox"])
    n2 -- Can be promoted to --> n3(["Release"])
```

# 2. Create one subissue per entity type to track individual implementation

Here are the classes of entities about the application to surface to the model.

[...]

PR #117, opened 2026-03-28
Add unit tests for entity store, model types, and judge scoring

DominicBurkart:
## Summary\n\n- **`harness/tests/entity_store_tests.rs`** — 15 tests covering `InMemoryEntityStore` CRUD lifecycle (store, exists, update, delete with error cases), query filtering (by entity type, tags, text search, time range, limit), and relationship management (create, referential integrity enforcement, bidirectional retrieval, delete). Previously the store had a single trivial test checking empty creation.\n\n- **`model/tests/types_serde_tests.rs`** — 17 tests verifying serialization contracts for the model API types. Covers all `ToolChoice` variants (including `Specific`), `MessageRole` rename_all, `FinishReason` snake_case, `SchemaType` lowercase, `Usage` round-trip, `ChatResponse` with multiple choices, `ChatRequest::with_tools` builder behavior, `assistant_with_tools` constructor, `tool_response` round-trip, `ToolCall`/`FunctionCall` serde, and nested `PropertySchema` with array items.\n\n- **`model/tests/judge_unit_tests.rs`** — 20 tests for the judge scoring helpers and config edge cases. Covers `calculate_coherence_score` boundaries (empty, single word, long text, paragraph bonus, word variety), `calculate_relevance_score` edge cases (no required keywords, all present, forbidden penalty, too-short response, empty prompt, clamping to [0,1]), `JudgeConfig::calculate_retry_delay` determinism with zero jitter and max cap, `ValidationMetrics` score clamping, `ValidationResult` accessor methods, and `ValidationCriteria` preset builder invariants.\n\n## Why these areas\n\nThese three areas had the largest gap between code complexity and test coverage:\n1. The entity store is the backbone of Nanna's entity management but had no meaningful tests\n2. The model types define the LLM API contract and need serde fidelity guarantees\n3. The judge scoring functions are pure logic with many edge cases worth pinning down\n\n## Test plan\n\n- [ ] `cargo nextest run --workspace --all-features` passes\n- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes\n- [ ] No existing tests broken\n\nhttps://claude.ai/code/session_011cihHmotavd4Y496r9Tevt"


DominicBurkart (commented):
## Review: Add unit tests for entity store, model types, and judge scoring

### CI Status: All green (17/17 checks pass)

### Overall

This is the strongest of the three overlapping test PRs (#112, #113, #117). It covers three distinct areas (entity store, model types serde, judge scoring) with 52 total tests across 883 lines. The tests are well-structured, each is self-contained, and they pin down important invariants.

### Duplicate coverage

This PR subsumes the entity store tests in #113 and the model types tests in #112. **Recommend merging this PR and closing #112 and #113.** The only content unique to #112 is `image-builder/src/validate_tests.rs` (5 image validation tests) -- that could be extracted into a small separate PR.

### Issues (see inline comments)

1. **Test name/assertion mismatch**: `coherence_very_long_text_penalized` doesn't assert long text scores lower -- either rename or add the comparison.
2. **Jitter test fragility**: The upper bound in `retry_delay_with_jitter_bounded` is tightly coupled to the jitter factor math. Add a comment explaining the derivation.
3. Minor: verify `GitRepository::new` produces unique IDs when called with identical args (the `query_respects_limit` test depends on this).

### What looks good

- Judge scoring tests are excellent -- the boundary conditions for coherence (empty, single word, long text, paragraph bonus, word variety) and relevance (forbidden keywords, empty prompt, clamping) are exactly the kind of edge cases that prevent regressions.
- Model types serde tests thoroughly cover all enum variants and the builder API.
- Entity store tests use `GitRepository` (vs `ContextEntity` in #113), providing integration-level coverage with a different entity type.
- `ValidationCriteria` preset builder tests (`technical_documentation_criteria_stricter_than_default`, `creative_writing_criteria_more_lenient`) are a good contract test pattern.

Needs rebase onto main (`mergeable_state: behind`).

---

DominicBurkart (commented):
## Code Review — PR #117 (head `ba79c30`)

CI: 17/17 checks green (unit on Linux/macOS/Windows, lint, integration, integration-container, security, cross-platform builds). Codecov: +0.28 pp to 60.96%. Branch is behind main — rebase required before merge.

Previous review (`79b7b2e`) identified three issues. Assessing whether they were addressed in the two subsequent commits, plus a full re-read of the diff.

---

### Carry-over from prior review: status

**Issue 1 — `coherence_very_long_text_penalized` name/assertion mismatch (previously raised, NOT fixed)**

`harness/tests/entity_store_tests.rs` is not the file; this is in `model/tests/judge_unit_tests.rs`. The test body still only bounds both scores to `(0, 1]` — it never asserts `long_score < short_score`. The test name claims long text is penalized but the assertion does not verify that claim. Either:
- rename to `coherence_very_long_text_valid_score_range`, or
- add `assert!(long_score < short_score, "long text ({}) should score lower than short text ({})", long_score, short_score);`

This is a correctness hole: the scoring function's length penalty could be silently removed and this test would not catch it.

**Issue 2 — `retry_delay_with_jitter_bounded` upper bound unexplained (previously raised, NOT fixed)**

The magic constant `150` (base 100 + 50% of 100) is still uncommented. Add a single-line comment:
```rust
// Upper bound: base_delay_ms + (base_delay_ms * jitter_factor) = 100 + 100*0.5 = 150ms
assert!(d <= Duration::from_millis(150));
```
Without this, a future agent adjusting `jitter_factor` in the fixture will not understand why 150 is the correct ceiling.

**Issue 3 — `query_respects_limit` relies on `GitRepository::new` producing unique IDs (previously raised, NOT fixed)**

Five entities are created with identical args `("u", "m")`. If `GitRepository::new` generates IDs from content (url + branch) rather than a UUID/timestamp, all five `store()` calls after the first will return `EntityError::AlreadyExists` and the loop will silently swallow errors because the `unwrap()` will panic on the second iteration — crashing the test rather than producing a useful failure. Verify that the ID is a UUID (not a hash of url+branch), or change the loop to use distinct URLs:
```rust
for i in 0..5 {
    store.store(Box::new(GitRepository::new(format!("u{i}").into(), "m".into()))).await.unwrap();
}
```

---

### New findings on head `ba79c30`

**[BUG] `store_duplicate_returns_error` — unused variable `_id`**

```rust
let _id = repo.id().to_string();
```
`_id` is captured but never used in any assertion. If the intent is to verify the stored ID matches the returned ID from the first `store()` call, the assertion is missing. If it is purely bookkeeping, remove the binding. As written, a future refactor moving ID generation into `store()` instead of the constructor would not be caught.

**[LOGIC] `update_existing_entity` does not assert the update persisted**

The test calls `update()` then asserts only `store.exists(&id)` — which was already true before the update. The mutation (`repo.default_branch = "develop"`) is never verified to have been persisted. Add a `get()` call and assert `default_branch == "develop"`, or this test provides zero confidence that `update()` actually writes changes through.

**[LOGIC] `delete_relationship` does not check the reverse index**

`create_and_get_relationship` verifies both directions are populated. `delete_relationship` only checks `get_relationships(&repo_id)` after deletion; it does not verify that the reverse entry in `ctx_id`'s relationship list is also removed. If `InMemoryEntityStore` maintains a bidirectional index and `delete_relationship` only removes one direction, this test will not catch the stale reverse entry.

**[LOGIC] `relevance_too_short_response_penalized` has a vacuous assertion**

```rust
let score = calculate_relevance_score("Short.", "explain something", &criteria);
assert!(score <= 1.0);
```
`score <= 1.0` is always true regardless of whether the short-response penalty is applied. The test name promises a penalty is observable but the assertion does not verify it. Compare against a long-enough response:
```rust
let long_response = "word ".repeat(25); // >= 100 chars
let long_score = calculate_relevance_score(&long_response, "explain something", &criteria);
assert!(score < long_score, "short response ({}) should score lower than adequate response ({})", score, long_score);
```

**[MINOR] `assistant_with_tools_none_content` double-consumes `msg.tool_calls`**

```rust
assert!(msg.tool_calls.unwrap().is_empty());
```
`unwrap()` moves out of `msg.tool_calls`. This compiles only because `tool_calls` is not used again, but the pattern is fragile — any future assertion on `msg` after this line will not compile. Prefer `assert!(msg.tool_calls.as_ref().unwrap().is_empty())` for consistency with the surrounding test style.

**[MINOR] `coherence_multiline_gets_paragraph_bonus` uses `>=` not `>`**

The assertion `multi_score >= single_score` permits equal scores, meaning the paragraph bonus could be 0.0 and this test would still pass. If the implementation guarantees a strict bonus for double-newline paragraphs, use `>`. If equality is intentional (e.g., the bonus is only applied above a word-count threshold), add a comment explaining why `>=` is correct.

**[COVERAGE GAP] `get()` / `retrieve()` method not tested**

The entity store tests cover `exists`, `store`, `update`, `delete`, `query`, `create_relationship`, `get_relationships`, and `delete_relationship` — but there is no test calling `get()` (or whatever the single-entity retrieval method is named). Per issue #20, the harness control flow is: Query Entities -> Entity Modification Decision -> Perform Entity Modification -> Update Entities. The `get()` path (fetch a specific entity by ID for inspection/modification) is a primary hot path and should be pinned.

**[COVERAGE GAP] No test for `get_relationships` on an entity with no relationships**

`create_and_get_relationship` starts from a populated store. The zero-relationship case (a freshly stored entity) should be a separate test verifying `get_relationships` returns an empty `Vec` rather than an error.

**[COVERAGE GAP] `ToolChoice::Specific` wire format not pinned**

`tool_choice_specific_round_trip` asserts `json.contains("my_tool")` but does not assert the full wire format. The OpenAI/Anthropic wire format for a specific tool is `{"type":"function","function":{"name":"my_tool"}}`. If the serde representation drifts from the provider contract, this test will not catch it. Pin the exact JSON:
```rust
assert_eq!(json, r#"{"type":"function","function":{"name":"my_tool"}}"#);
```
(adjust to match your actual serde representation — but it should be exact, not substring).

**[COVERAGE GAP] No negative serde test for unknown enum variants**

`FinishReason`, `MessageRole`, `SchemaType` round-trips are tested, but there is no test verifying that an unknown string (e.g., `"stop_sequence"`) either deserializes gracefully to an `Unknown` variant or returns a clean error. Provider APIs add new finish reasons; the current tests do not pin the failure mode.

---

### No issues found

- No PII or secrets in any test fixture.
- No production code modified (3 new test files, 0 changes to `src/`).
- No bloat: 883 lines for 52 tests across three distinct modules is proportionate.
- Issue #20 (Entity Management) alignment is good — entity store CRUD and query tests directly exercise the "Query Entities (RAG)" and "Update Entities" steps in the harness control flow diagram.
- `ValidationCriteria` preset builder contract tests (`technical_documentation_criteria_stricter_than_default`, `creative_writing_criteria_more_lenient`) are well-structured and directly enforce the scoring invariants downstream agents rely on.

---

### Summary for the implementing agent

Priority order for fixes before merge:

1. **Must fix** — `coherence_very_long_text_penalized`: add the comparison assertion or rename. The penalty logic is untested.
2. **Must fix** — `relevance_too_short_response_penalized`: vacuous `<= 1.0` assertion. Add a comparative assertion.
3. **Must fix** — `update_existing_entity`: assert the mutation persisted via a `get()` call.
4. **Must fix** — `delete_relationship`: also assert reverse index is cleared.
5. **Should fix** — `query_respects_limit`: use distinct URLs or verify ID generation is UUID-based.
6. **Should fix** — `retry_delay_with_jitter_bounded`: comment the magic constant derivation.
7. **Should add** — `get()` by ID test; `get_relationships` empty case; `ToolChoice::Specific` exact wire format assertion.
8. **Nice to have** — Unknown variant deserialization behavior test for `FinishReason`/`MessageRole`.

Rebase onto main is also required (`mergeable_state: behind`).

---

Issue #2, opened 2025-09-24
ci: Investigate and resolve cache-nix-action HTTP 400 errors


DominicBurkart:
## Problem
The CI cache system is experiencing persistent HTTP 400 errors that are being bypassed with version prefixes (`nix-v2-`) rather than properly resolved. This leads to poor build performance and potential storage waste.

## Current Issues
- HTTP 400 errors from GitHub Actions cache service: `Cache service responded with 400`
- Tar compression failures: `"/usr/bin/tar" failed with error: The process '/usr/bin/tar' failed with exit code 2`
- Builds running from scratch without cache benefits
- Band-aid solution with cache key versioning instead of root cause fix

## Current Workaround
```yaml
primary-key: nix-v2-build-containers-${{ hashFiles('**/flake.lock') }}
```

## Expected Solution
- Identify root cause of HTTP 400 responses from cache service
- Fix underlying cache corruption or configuration issues  
- Restore reliable caching with proper performance metrics
- Remove version prefix workarounds

## Acceptance Criteria
- [ ] Root cause of HTTP 400 errors identified and documented
- [ ] Cache system working without HTTP 400 errors
- [ ] Build times reduced by >50% with effective caching
- [ ] Remove all `nix-v2-` prefix workarounds
- [ ] Add cache health monitoring and alerting

## Priority
🚨 **CRITICAL** - This significantly impacts CI performance and maintainability.

PR #140, opened 2026-04-02
feat: add OpenAI-compatible gateway provider

### LINKED ISSUE #124: add AI gateway-friendly layer
Currently, the agent calls ollama essentially directly. 

Instead, nanna should be configurable via command line argument or config file to use a different openapi-compliant API as the model providers.

The goals here are: 
1. Functionality: nanna can be used to defer work to dedicated local servers, or potentially some second provider.
2. Extensibility: the move from ollama to vllm is being considered; improving modularity would help avoid code churn.

https://www.ibm.com/think/topics/ai-gateway

DominicBurkart:
## Summary

- Adds `OpenAICompatProvider` implementing `ModelProvider` for any OpenAI-compatible endpoint (vLLM, LiteLLM, OpenRouter, Ollama `/v1`)
- Introduces `GatewayConfig` enum with serde-tagged dispatch between Ollama and OpenAI-compat providers
- Adds CLI flags (`--provider`, `--api-base`, `--api-key`) for runtime provider selection
- `OllamaProvider` preserved unchanged for backward compatibility
- Includes unit tests for request/response serialization and config serde roundtrip

## Test plan

- [x] All existing tests pass (OllamaProvider unchanged)
- [x] New unit tests for OpenAI-compat serialization
- [ ] `cargo clippy` clean
- [ ] CI green

Closes #124

https://claude.ai/code/session_01ShDykYG2muaykPpmqbzv4A

### DIFF
--- a/harness/src/main.rs
+++ b/harness/src/main.rs
@@ -240,7 +240,11 @@ fn generate_swebench_report(
 fn build_provider_from_cli(
     cli: &Cli,
 ) -> Result<Arc<dyn ModelProvider>, Box<dyn std::error::Error>> {
-    build_provider_arc(cli.provider.as_str(), cli.api_base.as_deref(), cli.api_key.as_deref())
+    build_provider_arc(
+        cli.provider.as_str(),
+        cli.api_base.as_deref(),
+        cli.api_key.as_deref(),
+    )
 }
 
 fn build_provider_arc(
@@ -270,40 +274,6 @@ fn build_provider_arc(
         }
     };
     Ok(gateway.build_provider()?)
-}
-
-fn generate_swebench_report(
-    input: &std::path::Path,
-    output_dir: Option<&std::path::Path>,
-    compare: Option<&std::path::Path>,
-) -> Result<(std::path::PathBuf, Option<std::path::PathBuf>), Box<dyn std::error::Error>> {
-    use harness::eval::swebench_report::SweBenchReport;
-    use harness::eval::swebench_results::SweBenchRunResult;
-
-    let json = std::fs::read_to_string(input)?;
-    let run_result: SweBenchRunResult = serde_json::from_str(&json)?;
-
-    let owned_default;
-    let base_dir: &std::path::Path = match output_dir {
-        Some(p) => p,
-        None => {
-            owned_default = std::env::current_dir()?.join("results");
-            owned_default.as_path()
-        }
-    };
-
-    let report = SweBenchReport::new("SWE-bench Report", run_result);
-    let report_path = report.write_to_directory(base_dir)?;
-
-    let comparison_path = if let Some(compare_path) = compare {
-        let compare_json = std::fs::read_to_string(compare_path)?;
-        let compare_result: SweBenchRunResult = serde_json::from_str(&compare_json)?;
-        Some(report.write_comparison_to_directory(&compare_result, base_dir)?)
-    } else {
-        None
-    };
-
-    Ok((report_path, comparison_path))
 }
 
 fn create_tool_registry(workspace_root: &std::path::Path) -> ToolRegistry {
@@ -597,7 +567,10 @@ async fn health_check(provider: &dyn ModelProvider) -> Result<(), Box<dyn std::e
 
     match provider.health_check().await {
         Ok(()) => {
-            println!("\u{2713} Health check passed. {} is running and accessible.", name);
+            println!(
+                "\u{2713} Health check passed. {} is running and accessible.",
+                name
+            );
             info!("Health check successful");
         }
         Err(e) => {
--- a/harness/src/main.rs
+++ b/harness/src/main.rs
@@ -5,6 +5,7 @@ use harness::entities::{EntityStore, InMemoryEntityStore};
 use harness::tools::ToolRegistry;
 use model::prelude::*;
 use std::io::{self, Write};
+use std::sync::Arc;
 use tracing::{error, info};
 
 // NOTE: `main.rs` binds the workspace entity store to `InMemoryEntityStore`
@@ -16,6 +17,18 @@ use tracing::{error, info};
 #[command(name = "harness")]
 #[command(about = "A CLI tool for interacting with language models")]
 struct Cli {
+    /// Provider backend: "ollama" or "openai-compat"
+    #[arg(long, default_value = "ollama")]
+    provider: String,
+
+    /// Base URL for the API (overrides provider default)
+    #[arg(long)]
+    api_base: Option<String>,
+
+    /// API key for authentication (openai-compat only)
+    #[arg(long)]
+    api_key: Option<String>,
+
     #[command(subcommand)]
     command: Commands,
 }
@@ -94,8 +107,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let cli = Cli::parse();
 
-    let config = OllamaConfig::default();
-    let provider = OllamaProvider::new(config)?;
+    let provider = build_provider_from_cli(&cli)?;
 
     let workspace_root = std::env::current_dir()?;
     let tool_registry = create_tool_registry(&workspace_root);
@@ -111,7 +123,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
             if let Some(initial_prompt) = prompt {
                 single_chat(
-                    &provider,
+                    provider.as_ref(),
                     &tool_registry,
                     &model,
                     &initial_prompt,
@@ -121,7 +133,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 .await?;
             } else {
                 interactive_chat(
-                    &provider,
+                    provider.as_ref(),
                     &tool_registry,
                     &model,
                     tools,
@@ -132,13 +144,13 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             }
         }
         Commands::Models => {
-            list_models(&provider).await?;
+            list_models(provider.as_ref()).await?;
         }
         Commands::Tools => {
             list_tools(&tool_registry);
         }
         Commands::Health => {
-            health_check(&provider).await?;
+            health_check(provider.as_ref()).await?;
         }
         Commands::Agent {
             prompt,
@@ -148,6 +160,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             tools,
         } => {
             run_agent(
+                &cli.provider,
+                cli.api_base.as_deref(),
+                cli.api_key.as_deref(),
                 &prompt,
                 &model,
                 max_iterations,
@@ -161,7 +176,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             model,
             max_iterations,
         } => {
-            run_mcp_server(&model, max_iterations).await?;
+            run_mcp_server(
+                &cli.provider,
+                cli.api_base.as_deref(),
+                cli.api_key.as_deref(),
+                &model,
+                max_iterations,
+            )
+            .await?;
         }
         Commands::SweBenchReport {
             input,
@@ -212,6 +234,69 @@ fn generate_swebench_report(
     };
 
     Ok((report_path, comparison_path))
+}
+
+fn build_provider_from_cli(
+    cli: &Cli,
+) -> Result<Arc<dyn ModelProvider>, Box<dyn std::error::Error>> {
+    let gateway = match cli.provider.as_str() {
+        "ollama" => {
+            let mut cfg = OllamaConfig::default();
+            if let Some(base) = &cli.api_base {
+                cfg = cfg.with_base_url(base);
+            }
+            GatewayConfig::Ollama(cfg)
+        }
+        "openai-compat" => {
+            let base_url = cli
+                .api_base
+                .clone()
+                .unwrap_or_else(|| "http://localhost:8080".to_string());
+            GatewayConfig::OpenaiCompat(OpenAICompatConfig {
+                base_url,
+                api_key: cli.api_key.clone(),
+                default_model: "default".to_string(),
+                timeout: std::time::Duration::from_secs(30),
+            })
+        }
+        other => {
+            return Err(format!(
+                "Unknown provider: {}. Use 'ollama' or 'openai-compat'.",
+                other
+            )
+            .into());
+        }
+    };
+    Ok(gateway.build_provider()?)
+}
+
+fn build_provider_arc(
+    provider_name: &str,
+    api_base: Option<&str>,
+    api_key: Option<&str>,
+) -> Result<Arc<dyn ModelProvider>, Box<dyn std::error::Error>> {
+    let gateway = match provider_name {
+        "ollama" => {
+            let mut cfg = OllamaConfig::default();
+            if let Some(base) = api_base {
+                cfg = cfg.with_base_url(base);
+            }
+            GatewayConfig::Ollama(cfg)
+        }
+        "openai-compat" => {
+            let base_url = api_base.unwrap_or("http://localhost:8080").to_string();
+            GatewayConfig::OpenaiCompat(OpenAICompatConfig {
+                base_url,
+                api_key: api_key.map(|s| s.to_string()),
+                default_model: "default".to_string(),
+                timeout: std::time::Duration::from_secs(30),
+            })
+        }
+        other => {
+            return Err(format!("Unknown provider: {}", other).into());
+        }
+    };
+    Ok(gateway.build_provider()?)
 }
 
 fn create_tool_registry(workspace_root: &std::path::Path) -> ToolRegistry {
@@ -294,7 +379,7 @@ async fn store_repo_guidance_entity(
 }
 
 async fn single_chat(
-    provider: &OllamaProvider,
+    provider: &dyn ModelProvider,
     tool_registry: &ToolRegistry,
     model: &str,
     prompt: &str,
@@ -362,7 +447,7 @@ async fn single_chat(
 }
 
 async fn interactive_chat<S: EntityStore + Send>(
-    provider: &OllamaProvider,
+    provider: &dyn ModelProvider,
     tool_registry: &ToolRegistry,
     model: &str,
     enable_tools: bool,
@@ -461,7 +546,7 @@ async fn interactive_chat<S: EntityStore + Send>(
     Ok(())
 }
 
-async fn list_models(provider: &OllamaProvider) -> Result<(), Box<dyn std::error::Error>> {
+async fn list_models(provider: &dyn ModelProvider) -> Result<(), Box<dyn std::error::Error>> {
     println!("Available models:");
     let models = provider.list_models().await?;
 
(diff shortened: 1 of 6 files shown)


DominicBurkart (commented):
Automated code review — head `1cb8ac1` (no prior review found since last commit).

## Summary

Five areas examined: API key safety, `GatewayConfig` serde dispatch, `--api-key` CLI flag safety, `OllamaProvider` backward-compat, and unit-vs-integration test classification.

**Blocking security issues (2):**
1. `--api-key` is a plain CLI flag — it will appear in shell history and `ps` output on every platform. See inline comment on `harness/src/main.rs`.
2. `OpenAICompatConfig` derives `Debug` (and `Serialize`) unconditionally, meaning the raw key is emitted any time the config is printed via `{:?}` or serialised to disk/logs. See inline comment on `model/src/config.rs`.

**Non-blocking issues (4):** raw error-body echo could reflect a key if the upstream API ever echoes auth material; `GatewayConfig` serde tag has a naming mismatch risk; two builder functions duplicate logic; the `test_response_deserialization` test has a coverage gap for `finish_reason: null`.

Inline comments follow.

---

harness/src/main.rs
```diff
@@ -16,6 +17,18 @@ use tracing::{error, info};
...
 struct Cli {
+    /// Provider backend: "ollama" or "openai-compat"
+    #[arg(long, default_value = "ollama")]
+    provider: String,
+
+    /// Base URL for the API (overrides provider default)
+    #[arg(long)]
+    api_base: Option<String>,
+
+    /// API key for authentication (openai-compat only)
```
DominicBurkart: **[SECURITY — BLOCKING] `--api-key` exposes the secret in shell history and process listings.**

Passing a secret as a plain CLI argument means:
- It lands verbatim in `~/.bash_history` / `~/.zsh_history` on every shell that records history.
- It is visible in `/proc/<pid>/cmdline` and `ps aux` output to any local user for the lifetime of the process.

Recommended mitigations (pick one or combine):
1. **Environment variable fallback** — read from `NANNA_API_KEY` (or `OPENAI_API_KEY`) when `--api-key` is absent, and document that as the preferred path. clap supports `#[arg(long, env = "NANNA_API_KEY")]` for this directly.
2. **File-path flag** — accept `--api-key-file <path>` and read the secret from that file at startup.
3. At minimum, add a doc comment on this field warning users not to pass real keys on shared or logged systems.

The `env = "..."` approach is the lowest-friction fix:
```rust
/// API key for authentication (openai-compat only).
/// Prefer setting NANNA_API_KEY in the environment to avoid shell history exposure.
#[arg(long, env = "NANNA_API_KEY")]
api_key: Option<String>,
```

---

model/src/config.rs
```diff
@@ -100,6 +102,90 @@ impl Default for ModelDefaults {
     }
 }
```
DominicBurkart, line 103: **[SECURITY — BLOCKING] `#[derive(Debug, Serialize)]` on `OpenAICompatConfig` emits the raw `api_key` value.**

`#[derive(Debug)]` means any `{:?}` or `dbg!()` of this struct — including test failure output, tracing spans, and panic messages — will print the secret in plain text. `#[derive(Serialize)]` means the full key is included if this config is ever written to a config file, audit log, or serialised for any other purpose.

**Minimum required fix — redact in `Debug`:**
```rust
// Remove Debug from the derive and implement it manually:
#[derive(Clone, Serialize, Deserialize)]
pub struct OpenAICompatConfig { ... }

impl std::fmt::Debug for OpenAICompatConfig {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("OpenAICompatConfig")
            .field("base_url", &self.base_url)
            .field("api_key", &self.api_key.as_ref().map(|_| "[REDACTED]"))
            .field("default_model", &self.default_model)
            .field("timeout", &self.timeout)
            .finish()
    }
}
```

**Also recommended — skip in `Serialize`:**
```rust
#[serde(skip_serializing)]
pub api_key: Option<String>,
```
The key should be supplied at runtime via env var / flag, not persisted through serde round-trips.

---

model/src/openai_compat.rs
```diff
@@ -0,0 +1,529 @@
...
+        if status == reqwest::StatusCode::UNAUTHORIZED || status == reqwest::StatusCode::FORBIDDEN {
+            return Err(ModelError::Authentication);
+        }
+        if status == reqwest::StatusCode::TOO_MANY_REQUESTS {
+            return Err(ModelError::RateLimit);
+        }
+        if !status.is_success() {
+            let body = response.text().await.unwrap_or_default();
+            return Err(ModelError::Unknown {
+                message: format!("OpenAI-compat API error {}: {}", status, body),
```
DominicBurkart, line 295: **[SECURITY — INFO] Non-2xx error body is echoed verbatim into `ModelError::Unknown`.**

Some OpenAI-compatible gateways (e.g. OpenRouter) reflect authentication context or partial token material in 4xx response bodies. If the `ModelError` is later logged via `tracing::error!` or surfaced to a user, this could leak sensitive data.

Mitigations:
- Truncate `body` to a safe length before including it in the error message:
  ```rust
  let body = response.text().await.unwrap_or_default();
  let body = if body.len() > 512 { &body[..512] } else { &body };
  ```
- Consider not forwarding the body for 401/403 responses (those are already mapped to `ModelError::Authentication` above; this path only fires for other non-success codes, but a 5xx body could still contain echoed request context on some proxies).

---

model/src/config.rs
```diff
@@ -100,6 +102,90 @@ impl Default for ModelDefaults {
...
+        if self.timeout.is_zero() {
+            return Err("Timeout must be greater than 0".to_string());
+        }
+        Ok(())
+    }
+}
+
+/// Unified gateway configuration for provider selection.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "provider", rename_all = "kebab-case")]
```
DominicBurkart, line 230: **[NON-BLOCKING] `GatewayConfig` serde tag value: verify `rename_all = "kebab-case"` on `OpenaiCompat`.**

With `#[serde(tag = "provider", rename_all = "kebab-case")]`:
- `GatewayConfig::Ollama` → `"provider":"ollama"` ✓
- `GatewayConfig::OpenaiCompat` → `"provider":"openai-compat"` ✓ (the existing roundtrip test confirms this)

The dispatch itself is correct. One gap: an externally-authored config file using `"provider":"openai_compat"` (underscore, which is an easy mistake) will produce an opaque `unknown variant` serde error rather than a helpful message. Consider adding:
```rust
#[serde(rename = "openai-compat", alias = "openai_compat")]
OpenaiCompat(OpenAICompatConfig),
```
so either spelling is accepted, or at least document the expected exact string in the config file format.

---

harness/src/main.rs
```diff
@@ -147,13 +162,83 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                "Unknown provider: {}. Use 'ollama' or 'openai-compat'.",
+                other
+            )
+            .into());
+        }
+    };
+    Ok(gateway.build_provider()?)
+}
+
+fn build_provider_arc(
```
DominicBurkart: **[NON-BLOCKING] `build_provider_from_cli` and `build_provider_arc` are near-duplicates.**

Both functions perform the same `match provider_name` dispatch with identical `GatewayConfig` construction logic and defaults. The only difference is where the three inputs come from. A future provider addition requires editing both.

The minimal fix is one line — have `build_provider_from_cli` delegate to `build_provider_arc`:
```rust
fn build_provider_from_cli(
    cli: &Cli,
) -> Result<Arc<dyn ModelProvider>, Box<dyn std::error::Error>> {
    build_provider_arc(cli.provider.as_str(), cli.api_base.as_deref(), cli.api_key.as_deref())
}
```
This removes ~35 lines of duplicated match logic.

---

model/src/openai_compat.rs
```diff
@@ -0,0 +1,529 @@
...
+                        role: MessageRole::Assistant,
+                        content: c.message.content,
+                        tool_calls,
+                        tool_call_id: None,
+                    },
+                    finish_reason,
+                }
+            })
+            .collect();
+
```
DominicBurkart, line 159: **[NON-BLOCKING — TEST GAP] `finish_reason: null` / unknown values silently map to `Stop`.**

The catch-all `_ => Some(FinishReason::Stop)` handles both `None` (field absent / JSON `null`) and any unrecognised string (e.g. `"stop"` with a capital S from a non-standard endpoint). Mapping unknown values to `Stop` is a reasonable default, but it makes it impossible for the caller to distinguish a genuine `"stop"` from a server error or future new value.

There is no test covering this branch. Consider adding:
```rust
#[test]
fn test_response_finish_reason_null_becomes_stop() {
    let raw_json = r#"{"choices":[{"index":0,"message":{"role":"assistant","content":"hi"},"finish_reason":null}],"usage":null}"#;
    let raw: OpenAIRawResponse = serde_json::from_str(raw_json).unwrap();
    let resp = OpenAICompatProvider::parse_response(raw).unwrap();
    // Documents current behaviour; update if you choose to return None instead
    assert_eq!(resp.choices[0].finish_reason, Some(FinishReason::Stop));
}
```
This makes the intentional choice explicit and will catch any future regression.

---

model/src/config.rs
```diff
@@ -233,4 +319,57 @@ mod tests {
...
+        assert!(cfg.validate().is_ok());
+
+        cfg.base_url = "".to_string();
+        assert!(cfg.validate().is_err());
+
+        cfg.base_url = "ftp://bad".to_string();
+        assert!(cfg.validate().is_err());
+
+        cfg.base_url = "http://localhost:8080".to_string();
+        cfg.default_model = "".to_string();
```
DominicBurkart, line 405: **[NON-BLOCKING — TEST / FOLLOW-UP] This assertion will need updating after the `Serialize` fix.**

`assert_eq!(c.api_key, Some("sk-test".to_string()))` currently passes because `api_key` is serialised and deserialised in full. If `#[serde(skip_serializing)]` is added to `api_key` as recommended in the `OpenAICompatConfig` security comment, deserialization of the round-tripped JSON will yield `api_key: None`, and this assertion will start failing.

After applying the security fix, update this test to:
```rust
// api_key is not serialised; it must be supplied at runtime
assert_eq!(c.api_key, None);
```
This makes the intentional no-persist-of-key contract visible in the test suite.

---

DominicBurkart (commented):
Automated code review — head `1cb8ac1` (no prior review found since last commit).

## Summary

Six files changed, 787 additions. Five areas examined: API key handling, `GatewayConfig` serde dispatch, `--api-key` CLI flag safety, `OllamaProvider` backward-compat, and unit-vs-integration test classification.

---

### Security findings (2 blocking)

**1. `--api-key` CLI flag exposes secrets in shell history and `ps` output** (`harness/src/main.rs:28`)
Any value passed as `--api-key sk-...` is recorded in `~/.bash_history`, visible in `/proc/<pid>/cmdline`, and shown in `ps aux` output to all local users. The fix is low-effort: add `env = "NANNA_API_KEY"` to the clap attribute so the env-var path is the documented default.

**2. `OpenAICompatConfig` derives `Debug` and `Serialize` unconditionally, emitting the raw key** (`model/src/config.rs:103`)
Any `{:?}` format, `dbg!()`, test-failure dump, or serde serialisation of this struct will include the plaintext API key. Implement `Debug` manually to redact `api_key`, and add `#[serde(skip_serializing)]` to the field.

---

### Non-blocking findings (4)

**3. Raw error-body echoed into `ModelError::Unknown`** (`model/src/openai_compat.rs:285`) — some gateways reflect auth context in 4xx/5xx bodies; truncate to 512 bytes before including in the error message.

**4. `GatewayConfig` serde alias gap** (`model/src/config.rs:144`) — `"openai_compat"` (underscore) fails deserialization with an opaque error; add a `#[serde(alias)]` or document the exact expected string.

**5. Duplicated provider-construction logic** (`harness/src/main.rs:213`) — `build_provider_from_cli` and `build_provider_arc` share ~35 lines of identical match logic; `build_provider_from_cli` should delegate to `build_provider_arc`.

**6. Test coverage gap for `finish_reason: null`** (`model/src/openai_compat.rs:156`) + the roundtrip test assertion on `api_key` will need updating after fix #2 is applied (`model/src/config.rs:372`).

---

### Other checks

**OllamaProvider backward-compat:** confirmed unchanged. The diff touches only `harness/src/main.rs` call-sites (widening parameter types from `&OllamaProvider` to `&dyn ModelProvider`) and `model/src/config.rs` (additive new types). `model/src/ollama.rs` is not in the diff.

**Unit vs. integration tests:** all tests in `model/src/openai_compat.rs` and `model/src/config.rs` operate on pure data-transformation functions and synchronous constructors with no network I/O or `#[tokio::test]`. They are genuine unit tests.

---

DominicBurkart:
**CRON status** ([session](https://claude.ai/code/session_01KbZB2Y25oBcD14Thr1Xs31)): Merge conflicts in `harness/src/main.rs` and `harness/Cargo.toml` resolved (commit `4fd438e1`), incorporating both the PR's provider-selection infrastructure and main's `SweBenchReport` additions. The new review flagged two **security blockers**: `--api-key` on the CLI exposes the API key in shell history and `ps aux` output (fix: `#[arg(env = "NANNA_API_KEY")]`), and `OpenAICompatConfig` derives `Debug` + `Serialize` unconditionally, leaking the raw key in any debug output or serialization. Not ready for promotion until both security issues are fixed.

---

DominicBurkart (commented):
Automated code review — head `4fd438e` (two prior reviews at `1cb8ac1`; new commits present since last review).

## Status of previously-flagged issues

The three commits added since `1cb8ac1` introduce the `SweBenchReport` subcommand, two new binary targets (`harness-eval`, `harness-score`), and route `api_key` through `run_agent` / `run_mcp_server`. **Neither blocking security issue from the previous two reviews has been addressed.**

---

## Security findings (2 blocking — unchanged from prior reviews)

### 1. `--api-key` CLI flag exposes the secret in shell history and `ps` output (`harness/src/main.rs`)

```rust
/// API key for authentication (openai-compat only)
#[arg(long)]
api_key: Option<String>,
```

Any invocation such as `harness --api-key sk-...` is recorded in `~/.bash_history`, visible in `/proc/<pid>/cmdline`, and readable by all local users via `ps aux`. This is a well-known CLI secret-handling hazard.

**Fix:** add `env = "NANNA_API_KEY"` to the `clap` attribute and document the environment variable as the primary path. The flag can remain as an override, but the help text should warn that it is visible to other users.

```rust
#[arg(long, env = "NANNA_API_KEY")]
api_key: Option<String>,
```

### 2. `OpenAICompatConfig` derives `Debug` and `Serialize` without redacting `api_key` (`model/src/config.rs`)

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct OpenAICompatConfig {
    pub base_url: String,
    pub api_key: Option<String>,   // ← emitted verbatim by {:?} and serde
    ...
}
```

Any `{:?}` format, `dbg!()` macro, test-failure output, or serialisation of this struct (e.g. writing config to disk or logs) will include the plaintext key.

**Fix:**
- Implement `Debug` manually to print `api_key: Some("[REDACTED]")` / `None`.
- Add `#[serde(skip_serializing)]` (or a custom serializer that emits `null`) to the `api_key` field.

The existing roundtrip test `test_gateway_config_serde_roundtrip_openai_compat` asserts `c.api_key == Some("sk-test".to_string())` — that assertion will need updating once serialisation is fixed to skip the key.

---

## New finding in this diff

### 3. `api_key` forwarded as explicit function argument into `run_agent` and `run_mcp_server` (non-blocking)

```rust
async fn run_agent(
    provider_name: &str,
    api_base: Option<&str>,
    api_key: Option<&str>,   // ← new parameter
    ...
```

This is in-process forwarding and introduces no additional exposure beyond the CLI flag origin (finding #1). However, the `#[allow(clippy::too_many_arguments)]` attribute added to suppress the clippy warning is a code smell — the root cause is the duplicated `build_provider_from_cli` / `build_provider_arc` logic flagged in earlier reviews (finding #5 from the `2026-05-13` review). Collapsing both into a single helper would eliminate both the duplication and the overly long argument list.

---

## Non-blocking findings (carried over, still unresolved)

### 4. Raw error body echoed unbounded into `ModelError::Unknown` (`model/src/openai_compat.rs`)

```rust
let body = response.text().await.unwrap_or_default();
return Err(ModelError::Unknown {
    message: format!("OpenAI-compat API error {}: {}", status, body),
});
```

Some gateway implementations (LiteLLM, OpenRouter) reflect auth context or upstream headers in 4xx/5xx bodies. Truncate to a safe length before including in the error:

```rust
let body = response.text().await.unwrap_or_default();
let body = body.chars().take(512).collect::<String>();
```

### 5. `finish_reason: null` mapped silently to `FinishReason::Stop` (`model/src/openai_compat.rs`)

```rust
let finish_reason = match c.finish_reason.as_deref() {
    Some("tool_calls") => Some(FinishReason::ToolCalls),
    Some("length")     => Some(FinishReason::Length),
    Some("content_filter") => Some(FinishReason::ContentFilter),
    _ => Some(FinishReason::Stop),  // catches None (null) and unknown strings alike
};
```

`finish_reason: null` is valid in streaming responses. Consider mapping `None` to `None` to preserve the distinction, and logging or returning an error for truly unknown string values.

---

## Checks that passed

- **OllamaProvider backward-compat:** `model/src/ollama.rs` is not touched. Call-sites widened to `&dyn ModelProvider` correctly. Confirmed.
- **`GatewayConfig` serde tag:** `#[serde(tag = "provider", rename_all = "kebab-case")]` produces `"provider":"ollama"` and `"provider":"openai-compat"`. Roundtrip tests confirm both variants deserialise correctly.
- **CLI integration:** `--provider`, `--api-base`, and `--api-key` are wired end-to-end through `build_provider_from_cli` into `main`, and through `build_provider_arc` into `run_agent` and `run_mcp_server`. The integration exists.
- **No secrets or PII committed** in test fixtures or added files.
- **`SweBenchReport` subcommand** is purely additive, correctly gated behind `eval-runner`, and has four unit tests covering happy-path, bad JSON, missing input, and missing compare file.
- **New binary targets** (`harness-eval`, `harness-score`) are gated behind `required-features = ["eval-runner"]` — production builds are unaffected.
- **Unit vs. integration tests:** all new tests operate on pure data transformation or filesystem I/O with no network calls. Correctly classified.

---

**Summary:** The PR is still blocked by the two security issues (#1 and #2) that have been present since the initial submission. The new commits do not resolve them. The `--api-key` flag must read from an environment variable by default, and `OpenAICompatConfig::Debug`/`Serialize` must redact the key before this PR is safe to merge.

https://claude.ai/code/session_01PgHUWy1VwNSg48EME331CG

---

DominicBurkart:
**CRON status** — session https://claude.ai/code/session_01PgHUWy1VwNSg48EME331CG

**Security fixes applied** (commit `700643e`): `--api-key` CLI flag now uses `env = "NANNA_API_KEY"` to prevent key leakage via shell history/`ps aux`; `OpenAICompatConfig` Debug impl now redacts `api_key`; `api_key` marked `#[serde(skip_serializing)]`; error body truncated to 512 chars. Additionally, 4 missing stub files added to fix compile errors (commit `a987939`). **PR remains in draft** — blocked by merge conflict in `.github/workflows/` files requiring a `workflows`-scoped token for manual rebase.

---

DominicBurkart:
CI is substantially green — all Rust unit, integration, lint, security, and cross-platform build jobs pass (`All Checks Passed` gate: ✅), with only the two environment-level bring-up jobs ("macOS M-series + libkrun" and "Windows WSL2 + podman") failing, which are pre-existing infra flakiness unrelated to the changes in this PR. Both security issues have been addressed: the `--api-key` CLI flag now reads from the `NANNA_API_KEY` environment variable (reducing shell-history exposure), `OpenAICompatConfig` uses a custom `Debug` impl that redacts the key as `[REDACTED]`, and `#[serde(skip_serializing)]` prevents the key from appearing in any serialized output, all verified by new unit tests. This PR is not yet ready to merge — it remains in draft and the `mergeable_state` is blocked, but once the draft is lifted and the author confirms the two infra bring-up failures are pre-existing (not regressions), it should be mergeable.

---

PR #141, opened 2026-04-03
test: extract common test harness, entity store unit tests, agent state machine tests

DominicBurkart:
## What this PR does

Three targeted test improvements that reduce duplication and add isolated unit coverage for the two areas that had none.

---

### 1. `harness/tests/common/mod.rs` — shared mock helpers (new file)

`integration_tests.rs` defined `SequenceMockProvider`, `make_tool_call`, `make_stop_response`, and `wrap_with_state_machine_responses` inline. Any future test file that needs a scripted LLM would have had to copy those definitions verbatim.

This file extracts all four into a single `common` module with `#[allow(dead_code)]` so the compiler doesn't warn when only a subset is imported. Test files opt in with:

```rust
#[path = "common/mod.rs"]
mod common;
use common::{SequenceMockProvider, make_stop_response, wrap_with_state_machine_responses};
```

`wrap_with_state_machine_responses` encodes the five-state scaffolding required for every agent run (EnrichingEntities → PlanningEntityModification → PerformingEntityModification → UpdatingEntities → CheckingTaskCompletion), which was previously copy-pasted implicitly every time a test set up its response list.

---

### 2. `harness/tests/entity_unit_tests.rs` — `InMemoryEntityStore` unit tests (new file)

`InMemoryEntityStore` had a single placeholder test (`assert_eq!(store.entities.len(), 0)`). This file adds 12 focused async unit tests covering:

| Invariant | Test |
|---|---|
| Store → exists | `store_and_exists_by_id` |
| Duplicate store → `AlreadyExists` | `store_duplicate_id_returns_already_exists_error` |
| Three entities all queryable | `store_multiple_entities_all_retrievable_via_query` |
| Type filter works | `query_by_entity_type_filters_correctly` |
| `limit` is respected | `query_limit_is_respected` |
| Update succeeds | `update_existing_entity_succeeds` |
| Update unknown → `NotFound` | `update_nonexistent_entity_returns_not_found` |
| Delete removes entity | `delete_removes_entity` |
| Delete unknown → `NotFound` | `delete_nonexistent_entity_returns_not_found` |
| Relationship creation + bidirectional retrieval | `create_and_retrieve_relationship` |
| Relationship requires both entities present | `create_relationship_requires_both_entities_to_exist` |
| Relationship deletion | `delete_relationship_removes_it` |

All tests use real concrete types (`ContextEntity`, `GitRepository`) so nothing is stubbed at the entity level.

---

### 3. `harness/tests/agent_state_machine_tests.rs` — agent state-machine unit tests (new file)

The agent state machine (ARCHITECTURE.md: plan → perform → check → decide → query/re-plan) had integration-level coverage in `integration_tests.rs` but no file focused exclusively on transition correctness. This file adds five tests that use the shared `SequenceMockProvider` to drive the agent with scripted responses:

| Test | What it verifies |
|---|---|
| `agent_loop_transitions_plan_to_perform_to_complete` | Full canonical state sequence appears in `state_history()` in order |
| `agent_loop_recognizes_completion_signal` | `task_completed = true` and `final_state = Completed` when "COMPLETE" is returned |
| `agent_loop_replans_when_incomplete` | `PlanningEntityModification` appears ≥ 2 times after an INCOMPLETE cycle |
| `agent_loop_enters_querying_entities_on_query_decision` | `QueryingEntities` appears when decision response contains "QUERY" |
| `agent_loop_records_tool_calls_in_result` | `AgentRunResult.tool_calls_made` is populated for tool-call responses |

---

## Test plan

- [ ] `cargo test -p harness --test entity_unit_tests` — all 12 entity store tests pass without network/container
- [ ] `cargo test -p harness --test agent_state_machine_tests` — all 5 state-machine tests pass without network/container  
- [ ] `cargo test -p harness --test integration_tests` — existing tests unaffected (no definitions were removed from that file; the common module is purely additive)
- [ ] `cargo test -p harness` — full harness test suite green


DominicBurkart:
All 18 CI check runs are passing (success or expected-skipped), including unit tests on Ubuntu/macOS/Windows, integration tests, lint, security scan, and cross-platform builds for x86_64 and aarch64. This PR received its first-ever code review from this session, covering code quality, correctness, and security. The review agent is posting its findings; this PR should be reviewed and any open comments addressed before promoting it out of draft status.

---

DominicBurkart (commented):
This PR adds three new test files: a shared mock helper module, `InMemoryEntityStore` unit tests, and agent state-machine tests. The direction is correct and the coverage gaps being addressed are real. That said, there are several issues worth resolving before this graduates out of draft.

---

## 1. Duplicate coverage with `entity_store_tests.rs` — the new `entity_unit_tests.rs` adds little net value

`harness/tests/entity_store_tests.rs` already exists and covers the same contract in full: `store_and_exists`, `store_duplicate_returns_error`, `update_existing_entity`/`update_nonexistent_entity_fails`, `delete_existing_entity`/`delete_nonexistent_entity_fails`, `query_filters_by_entity_type`, `query_respects_limit`, `create_and_get_relationship`, `create_relationship_with_missing_entity_fails`, `delete_relationship`. Every test in `entity_unit_tests.rs` maps 1-to-1 to a test in that file.

The PR description says `InMemoryEntityStore` had "a single placeholder test" and this file fills the gap — but `entity_store_tests.rs` already filled it. The net result is duplicate maintenance surface with no new invariants exercised. Three things `entity_store_tests.rs` has that `entity_unit_tests.rs` does not:
- `query_filters_by_tags`
- `query_text_search_matches_json_content`
- `query_filters_by_time_range`

**Recommendation:** Either (a) drop `entity_unit_tests.rs` entirely and note in the PR that the gap was already covered, or (b) make `entity_unit_tests.rs` the canonical file, delete `entity_store_tests.rs`, and port the three missing query-filter tests over.

---

## 2. `integration_tests.rs` still defines its own `SequenceMockProvider` — the extraction is incomplete

The PR description states that `SequenceMockProvider`, `make_tool_call`, `make_stop_response`, and `wrap_with_state_machine_responses` were "defined inline" in `integration_tests.rs` and have been extracted. However, `integration_tests.rs` (82 KB, SHA `a6d7a4334`) is unchanged in this diff — it still imports and defines its own versions of these helpers. The `common` module is only wired up in the two new test files.

If the goal is to eliminate future copy-paste, `integration_tests.rs` needs to be updated to `use` the shared module, otherwise the duplication this PR is supposed to fix will persist in the largest test file in the codebase.

**Recommendation:** Migrate `integration_tests.rs` to import from `common/mod.rs` using the same `#[path = "common/mod.rs"] mod common;` pattern (or a symlink/re-export). If that is intentionally deferred, the PR description should say so and a follow-up issue should be filed.

---

## 3. `#[path = "common/mod.rs"] mod common;` is fragile and non-idiomatic

Using `#[path]` to pull in a sibling module works for integration tests that are each independent binary crates, but it carries risks:
- The `#[path]` attribute is resolved relative to the file that contains it. If a test file is ever moved or the `common/` directory is renamed, the path silently breaks at compile time with a hard-to-diagnose error.
- The pattern forces every consumer to repeat the `#[path]` declaration and the `use` imports. Rust's idiomatic solution for shared integration-test helpers is a `tests/common/mod.rs` that is pulled in via `mod common;` without a path attribute (Cargo resolves `tests/common/mod.rs` automatically when an integration test declares `mod common;`). The `#[path]` attribute here suggests the author may not have been aware of that convention, or there is a reason it was not usable here that should be documented.

**Recommendation:** Remove the `#[path]` attributes and rely on the conventional `mod common;` declaration. Cargo's test runner finds `tests/common/mod.rs` automatically. If there is a reason the convention cannot be used here, add a comment explaining it.

---

## 4. `agent_loop_replans_when_incomplete` hard-codes the internal LLM call sequence and will silently break on agent changes

`harness/tests/agent_state_machine_tests.rs` lines 159–173 construct a raw `Vec<ChatResponse>` that exactly mirrors the agent's internal call sequence: one call for `PlanningEntityModification`, one for `PerformingEntityModification`, none for `UpdatingEntities`, one for `CheckingTaskCompletion`, one for `EntityModificationDecision`, etc. This is also true of `agent_loop_enters_querying_entities_on_query_decision` (lines 197–220).

The problem is that `SequenceMockProvider` pops responses in strict FIFO order. If the agent ever adds an LLM call to `EnrichingEntities` or `UpdatingEntities` (both currently silent), or changes the number of calls inside any state, these tests will panic with `ServiceUnavailable: No more scripted responses` — or worse, silently consume the wrong response and produce a confusing assertion failure. There is no indication in the test that the queue length matters.

The tests that use `wrap_with_state_machine_responses` are better insulated because the scaffolding is hidden, but the two multi-cycle tests bypass the wrapper entirely.

**Recommendation:** At minimum, add an assertion after `agent.run()` that the mock provider's internal queue is empty (i.e., every scripted response was consumed). This makes mismatches fail loudly at the queue-exhaustion point rather than in a downstream assertion. A `remaining_responses()` accessor on `SequenceMockProvider` would suffice.

---

## 5. `SequenceMockProvider` panics inside `lock().unwrap()` under test parallelism

`harness/tests/common/mod.rs` line 60:
```rust
let mut responses = self.responses.lock().unwrap();
```
`unwrap()` on a `Mutex` lock poisons the mutex if any thread that previously held the lock panicked. In a test binary that runs multiple `#[tokio::test]` cases concurrently, a panic in one test can poison the mutex and cause every other test using a `SequenceMockProvider` instance to crash with an `PoisonError`, turning one failure into many and obscuring the root cause.

In practice each `AgentLoop` gets its own `Arc<SequenceMockProvider>` instance so the scope of sharing is limited, but the `unwrap()` should still be replaced with `expect()` that names the context, or — more robustly — `lock().unwrap_or_else(|e| e.into_inner())` to recover from poisoning gracefully in tests.

---

## 6. `agent_loop_transitions_plan_to_perform_to_complete`: `history[0]` assertion is wrong per the production code

`harness/tests/agent_state_machine_tests.rs` lines 110–115:
```rust
assert_eq!(
    history[0],
    AgentState::PlanningEntityModification,
    "first recorded state should be PlanningEntityModification (EnrichingEntities produces no LLM call)"
);
```

Looking at `harness/src/agent/mod.rs` line 552, `transition_to()` pushes the *destination* state onto `state_history`. The agent starts in `EnrichingEntities` (line 309), then calls `transition_to(PlanningEntityModification)` (line 491) — so `history[0]` is indeed `PlanningEntityModification`. So far so good.

However, the comment "EnrichingEntities produces no LLM call" conflates two separate facts: (1) `EnrichingEntities` does not make an LLM call, and (2) `EnrichingEntities` is the *starting* state and is never pushed to history because `transition_to` only records the *arriving* state, not the departing one. The latter is the actual reason `history[0]` is `PlanningEntityModification`. This distinction matters if `enrich_entities` ever gains an LLM call — the test would still pass but the reasoning in the comment would become wrong, masking the now-incorrect assumption.

**Minor:** Update the comment to reflect that `EnrichingEntities` is the initial state and is not recorded in history (it is never a transition destination), not just that it makes no LLM call.

---

## 7. `make_context_entity` signature diverges from `ContextEntity::default()` used in `entity_store_tests.rs`

`harness/tests/entity_unit_tests.rs` lines 19–27 define:
```rust
fn make_context_entity(description: &str) -> ContextEntity {
    ContextEntity::new(
        description.to_string(),
        vec[],
        vec[],
        String::new(),
        "test-model".to_string(),
    )
}
```
Meanwhile `entity_store_tests.rs` uses `ContextEntity::default()` directly, with no constructor arguments. If `ContextEntity::new()` ever changes its signature (e.g., adding a required parameter), only `entity_unit_tests.rs` will break; `entity_store_tests.rs` will continue to compile. The asymmetry makes the codebase harder to maintain. Given that the two test files are now directly redundant (see issue 1 above), this is a secondary concern, but the pattern of using the `new()` constructor where `default()` is available and sufficient is worth noting.

---

## 8. Missing test: `SequenceMockProvider` exhaustion behaviour is unverified

The PR tests that the mock drives the agent correctly when responses are available, but there is no test that verifies the `ServiceUnavailable` error path when the queue is empty mid-run. If the agent silently tolerates a missing response (e.g., falls back to a default), a test that scripts too few responses would pass without exercising the intended code path. A unit test for the mock itself — `sequence_mock_returns_error_when_exhausted` — would prevent regressions in the helper that all other tests depend on.

---

## 9. `agent_loop_records_tool_calls_in_result` does not verify the tool was actually *executed*

`harness/tests/agent_state_machine_tests.rs` lines 246–268: the test asserts that `result.tool_calls_made[0].tool_name == "echo"` but does not assert anything about the tool's output appearing in the conversation history, or that `EchoTool` actually ran (i.e., that a `ToolMessage` response was added to the history). The test would pass even if the agent recorded the *request* for a tool call but the tool executor crashed and was silently swallowed. Consider asserting that `result.conversation_snapshot` contains at least one message with `role == Tool` and the expected echo output.

---

## 10. Draft status and `mergeable_state: "behind"`

The PR is currently marked as a draft and the branch is behind `main`. Before requesting a final review, the branch should be rebased or merged from main to resolve the `behind` status and the draft flag should be removed when the above concerns are addressed. This is noted as a process point, not a blocker.

---

## Summary

| # | Severity | File | Issue |
|---|---|---|---|
| 1 | Medium | `entity_unit_tests.rs` | Wholesale duplication of `entity_store_tests.rs`; no new invariants |
| 2 | Medium | `integration_tests.rs` (unchanged) | Mock extraction is incomplete; the largest test file still has its own copies |
| 3 | Low | `common/mod.rs`, consumers | `#[path]` is non-idiomatic; `mod common;` without path attribute is the Cargo convention |
| 4 | Medium | `agent_state_machine_tests.rs:159,197` | Raw response queues encode implicit call-count coupling; add queue-empty assertion |
| 5 | Low | `common/mod.rs:60` | `lock().unwrap()` will propagate test panic poisoning; use `expect()` or recover |
| 6 | Low | `agent_state_machine_tests.rs:110` | Misleading comment conflates "no LLM call" with "not recorded in history" |
| 7 | Low | `entity_unit_tests.rs:19` | `ContextEntity::new(...)` where `default()` suffices, asymmetric with sibling test file |
| 8 | Low | (missing) | No unit test for `SequenceMockProvider`'s own exhaustion error path |
| 9 | Low | `agent_state_machine_tests.rs:246` | Tool-call test doesn't verify tool execution result, only call recording |

Items 1 and 2 are the most important to resolve. Item 4 is a latent fragility that will cause confusing failures as the agent evolves. The rest are polish-level concerns that are straightforward to fix.

---

PR #142, opened 2026-04-03
arch: remove dead monitoring stub in observability

DominicBurkart:
Remove the call to `MonitoringSystem::start_monitoring` from `ObservabilitySystem::initialize`, as that method spawned a background task whose loop body did nothing except emit a debug log message ("Performing periodic monitoring checks") and never performed any actual work.

### DIFF
--- a/harness/src/observability.rs
+++ b/harness/src/observability.rs
@@ -573,12 +573,9 @@ impl ObservabilitySystem {
                 reason: e.to_string(),
             })?;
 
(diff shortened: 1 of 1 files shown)


DominicBurkart:
Closing — overlaps with the broader #324 effort to drop the monitoring/telemetry/observability subsystem. Whichever direction wins on #324 makes this PR's narrower stub-removal moot. Linked issues remain open.

PR #177, opened 2026-04-11
add all checks passed job

DominicBurkart:


DominicBurkart:
implement these changes please

# Plan: Future-proof `all-checks` job in ci.yml



## Context



PR #177 adds an `all-checks` job to `.github/workflows/ci.yml` intended as a single required branch-protection status check. The current implementation lists only three jobs in `needs`:



```yaml

needs: [test-matrix, build-matrix, build-containers]

```



But the workflow has six other jobs beyond `all-checks` itself: `test-matrix`, `build-matrix`, `build-containers`, `security-scan`, `cache-maintenance`, `release`, `ci-summary`. The current `all-checks` runs before `security-scan`, `cache-maintenance`, `release`, and `ci-summary` have a chance to complete, and won't catch failures in them.



The user wants `all-checks` to cover everything that runs on a given commit or PR without manual upkeep.



## Constraint



GitHub Actions `needs` requires explicit job names — there is no wildcard or "all other jobs" syntax. However:



- Jobs listed in `needs` that are **skipped** (via their own `if:` condition) will report `result: 'skipped'`, which does NOT trigger the existing `failure` / `cancelled` check.

- A job with `if: always()` will still run even if some of its `needs` are skipped.



So we can safely add all conditional jobs (`security-scan`, `cache-maintenance`, `release`) to `needs`, and they'll just show as `skipped` when their conditions don't apply.



The `needs.*.result` wildcard already checks results dynamically across whatever is listed — the only thing that needs updating when a new job is added is the `needs` list itself (one place, one line).



## Approach



Expand the `needs` list of `all-checks` to include **every other job in the workflow**:



```yaml

all-checks:

  name: All Checks Passed

  runs-on: ubuntu-latest

  if: always()

  needs:

    - test-matrix

    - build-matrix

    - build-containers

    - security-scan

    - cache-maintenance

    - release

    - ci-summary

  steps:

    - name: All checks passed

      if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}

      run: echo "All CI checks passed"

    - name: Some checks failed

      if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}

      run: exit 1

```



No changes to the step logic — `needs.*.result` already fans out dynamically across everything in `needs`. The only change from the PR diff is the expanded `needs` list.



### Why this covers "everything run on a commit or PR"



- On a **PR**: `security-scan`, `cache-maintenance`, `release` are skipped → `all-checks` sees `skipped` for them, does not fail.

- On a **push to main**: `release` is skipped → same.

- On a **release**: `cache-maintenance` is skipped → same.

- In all cases, `all-checks` waits for the full terminal layer before reporting.



### Maintenance going forward



When a new job is added to the workflow, add it to `all-checks`'s `needs` list. This is a single, obvious place — the job will fail loudly if missed, making it self-enforcing.



## Critical file



- `.github/workflows/ci.yml` — specifically the `all-checks` job starting around line 614 of the PR head (`feat/all-checks-ci` branch).



## Implementation steps



1. Checkout / switch to branch `claude/future-proof-check-mapping-xnPjd`.

2. Edit `.github/workflows/ci.yml`: replace the `all-checks` job's `needs` line with the full list above.

3. Commit with a clear message.

4. Push to `claude/future-proof-check-mapping-xnPjd`.



## Verification



- Diff shows only the `needs` field of `all-checks` changed.

- YAML is valid (`nix develop --command` is not needed here, but a quick `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` confirms no syntax errors).

- On a subsequent PR, the `all-checks` job should appear after all other jobs complete, and branch-protection can use "All Checks Passed" as the sole required check.

PR #183, opened 2026-04-11
refactor(main): eliminate provider duplication and trivial wrapper

DominicBurkart:
## Summary

This PR makes three targeted improvements to `harness/src/main.rs` with no new dependencies or logic changes.

### Action items

- **Remove `create_tool_registry` wrapper function** — the function was a one-liner that only called `harness::tools::create_tool_registry`. It added an indirection layer with no semantic value (no error handling, no transformation, not even a doc comment). All three call-sites now call the underlying function directly. This removes 4 lines of dead indirection.

- **Eliminate duplicate `OllamaProvider` initialization** — `main()` constructed `OllamaProvider::new(OllamaConfig::default())` for `Chat`/`Models`/`Health` commands, but `run_agent` and `run_mcp_server` each silently constructed their *own* `OllamaProvider::new(OllamaConfig::default())`, ignoring the one from `main()`. This meant the `Agent` and `McpServe` subcommands always paid the construction cost twice. The fix creates a single `Arc<dyn ModelProvider>` at startup and threads it through both functions.

- **Unify `Arc` and `ModelProvider` trait usage** — with the above change, `Arc<dyn ModelProvider>` replaces the concrete `OllamaProvider` in all helper-function signatures. This is also a step toward the architecture's stated goal of provider-agnosticism (the concrete type is now only named once, at the top of `main()`). The `use std::sync::Arc` import moves from inside individual `async fn` bodies to the module level, and the now-unused `use harness::tools::ToolRegistry` import is dropped.

### What was reviewed but not changed

- `monitoring.rs`, `observability.rs`, `telemetry.rs` — These three modules (~107 KB combined) contain stub implementations with mostly-hardcoded values. They are referenced by `agent/eval.rs`, so they cannot be safely removed at this time. The evaluation framework's use of `ObservabilitySystem` is guarded by `collect_observability: false` in all test configurations, so the stubs never cause false signals in CI. These are candidates for a future refactor once the evaluation path matures.
- No changes to `lib.rs` re-exports, entity definitions, MCP protocol implementation, or test files.

https://claude.ai/code/session_018Uk9aCCyjsv4r1aGQr7SV7

### DIFF
--- a/harness/src/main.rs
+++ b/harness/src/main.rs
@@ -2,9 +2,9 @@ use clap::{Parser, Subcommand};
 use harness::entities::ast::WorkspaceScanner;
 use harness::entities::git::GitRepository;
 use harness::entities::{EntityStore, InMemoryEntityStore};
-use harness::tools::ToolRegistry;
 use model::prelude::*;
 use std::io::{self, Write};
+use std::sync::Arc;
 use tracing::{error, info};
 
 #[derive(Parser)]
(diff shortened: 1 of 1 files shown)


DominicBurkart:
Closing — superseded by #268 (provider/AgentLoop consolidation already landed on main as commit 050267b). Linked issues remain open.

PR #184, opened 2026-04-11
test(entities): add comprehensive unit tests for InMemoryEntityStore

DominicBurkart:
## What part of the code was chosen and why?

`harness/src/entities/mod.rs` — specifically the `InMemoryEntityStore` implementation and its `EntityStore` trait.

This is the runtime backbone of the entire entity graph. Every agent loop iteration routes through it: entities are stored, queried (with RAG-style filtering), related to each other, updated, and deleted as the agent works through a task. Despite being central to the system, the existing test coverage was a single stub:

```rust
#[tokio::test]
async fn test_in_memory_store_basic_operations() {
    // Note: Full tests will be added when concrete entity types are implemented
    let store = InMemoryEntityStore::new();
    assert_eq!(store.entities.len(), 0);
}
```

That stub asserts only that a freshly-created store is empty. The non-trivial query engine (type filtering, tag filtering, time-range filtering, text search with case folding, relevance sorting, result limiting) and all relationship lifecycle methods had **zero test coverage**.

## Current test disposition

- **Before this PR**: 1 stub test; every real behaviour path was untested.
- **After this PR**: 30 unit tests; every documented behaviour of `InMemoryEntityStore` is covered.

## Key invariants the new tests validate

| Area | Invariants validated |
|---|---|
| **store / exists** | Stored entity is found by id; unknown id returns false; two distinct entities receive distinct UUIDs |
| **update** | Replacement entity is persisted; updating a non-existent id returns `EntityError::NotFound` |
| **delete** | Entity is gone after deletion; deleting a non-existent id returns `EntityError::NotFound` |
| **query – no filters** | Full scan returns every entity in the store |
| **query – type filter** | Correct entities returned for a single type, multiple types, and a type with no matches |
| **query – tag filter** | Only entities carrying the requested tag are returned; absent tag yields empty results |
| **query – time range** | Entities created within the range are included; entities outside are excluded |
| **query – text search** | Substring match (case-insensitive) finds entity; absent token yields empty results |
| **query – limit** | Result set is capped at the requested limit; a limit larger than the store returns all |
| **query – ordering** | Results are returned in descending relevance order |
| **relationships – create / get** | Relationship is visible from both endpoints; creating a relationship with a missing source or target returns `EntityError::NotFound` |
| **relationships – delete** | The targeted relationship is removed; only the matching `RelationshipType` is deleted, leaving others intact |
| **combined filters** | type + text and type + limit filters compose correctly |

https://claude.ai/code/session_018Uk9aCCyjsv4r1aGQr7SV7

### DIFF
--- a/harness/src/entities/mod.rs
+++ b/harness/src/entities/mod.rs
@@ -19,6 +19,9 @@ pub mod git;
 pub mod telemetry;
 pub mod test;
 
+#[cfg(test)]
+mod store_tests;
+
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
--- a/harness/src/entities/store_tests.rs
+++ b/harness/src/entities/store_tests.rs
@@ -1,0 +1,558 @@
+//! Unit tests for `InMemoryEntityStore`.
+//!
+//! The store is the runtime backbone of the entity graph; all CRUD, query
+//! filtering, relationship management, and error paths are exercised here.
+//!
+//! The existing `mod tests` block inside `mod.rs` only asserts that a freshly
+//! created store is empty — every other behaviour is untested.
+
+#[cfg(test)]
+mod tests {
+    use crate::entities::context::types::ContextEntity;
+    use crate::entities::git::types::{GitBranch, GitRepository};
+    use crate::entities::{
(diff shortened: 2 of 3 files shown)


DominicBurkart:
Closing — duplicate. `harness/tests/entity_store_tests.rs` already exists on main from 4152d66 / 79b7b2e. The PR's 576-line version conflicts add/add. Linked issues remain open.

PR #198, opened 2026-04-12
test(entities): comprehensive InMemoryEntityStore unit tests

DominicBurkart:
## What component was chosen and why

**`InMemoryEntityStore`** in `harness/src/entities/mod.rs`.

The store is the central data structure of the entire harness: every entity the agent loop reads, writes, or relates during a task flows through it. It implements a non-trivial `EntityStore` trait with eight methods, including multi-predicate query filtering (entity type, tags, time range, free-text), optimistic duplicate detection, and referential-integrity guards on relationships.

Despite this breadth, the existing `#[cfg(test)]` block in that file contained exactly **one** test — `test_in_memory_store_basic_operations` — which only asserted `store.entities.len() == 0`. Every meaningful code path was untested.

## Current test disposition

| Area | Pre-PR coverage |
|------|----------------|
| `store()` / `exists()` | Zero tests |
| `update()` / `delete()` | Zero tests |
| `query()` — type filter | Zero tests |
| `query()` — tag filter | Zero tests |
| `query()` — text search | Zero tests |
| `query()` — time range | Zero tests |
| `query()` — limit | Zero tests |
| `create_relationship()` | Zero tests |
| `get_relationships()` | Zero tests |
| `delete_relationship()` | Zero tests |
| Error paths (NotFound, AlreadyExists) | Zero tests |

All integration tests that touch the store (`integration_tests.rs`, etc.) require a live Ollama instance and are excluded from the tarpaulin coverage run via `tarpaulin.toml`.

## Key invariants the new tests validate

1. **Round-trip identity** — `store()` returns the same ID that was assigned to the entity's metadata; `exists()` reflects the live state.
2. **Duplicate guard** — storing an entity whose ID already exists returns `EntityError::AlreadyExists`.
3. **Missing-entity guard** — `update()` and `delete()` on an unknown ID return `EntityError::NotFound`; `delete()` on a previously-deleted ID also returns `NotFound`.
4. **Query type filter** — results contain only entities matching the requested `EntityType`; an empty `entity_types` vec returns everything (no filter).
5. **Query tag filter uses OR semantics** — a result is included if it has *any* of the queried tags, not *all*.
6. **Text search is case-insensitive** and excludes entities whose serialized JSON does not contain the search term.
7. **Time range filter** — entities created outside the window are excluded; entities inside are included.
8. **Limit truncation** — results are capped at `limit`; a limit larger than the result count returns all matches.
9. **Relationship referential integrity** — `create_relationship()` rejects edges whose source or target entity does not exist.
10. **Relationship retrieval is bidirectional** — `get_relationships(id)` returns edges where `id` appears as either `from` or `to`.
11. **Relationship delete is type-specific** — deleting one edge type between a pair leaves other edge types intact.
12. **Full lifecycle** — store → update → text-query → delete sequence behaves correctly end-to-end.
13. **Multi-type coexistence** — entities of different types coexist in the store and are independently queryable without cross-contamination.

## Testing strategy

**Unit tests in a dedicated integration-test file** (`harness/tests/entity_store_tests.rs`).

- **Why unit-style rather than integration**: No network, container, or filesystem dependencies. Tests run in every `cargo test` invocation and are counted by tarpaulin.
- **Why a separate file rather than in-file**: `entities/mod.rs` already has a `#[cfg(test)]` block; a separate file keeps it uncluttered and makes the new tests easy to find, run in isolation, and extend.
- **Why not property tests (proptest)**: The store's correctness properties are structural (presence/absence, count, filter semantics) rather than numerical. Concrete, named scenarios communicate the invariants more clearly and produce actionable failure messages when they break.
- **Realistic data**: Tests use real-looking git URLs, branch names, commit SHAs, and task descriptions rather than placeholder strings like `"test"` or `""`.

https://claude.ai/code/session_01TEgxmgXgnN2ByXi3WWLYcp

### DIFF
--- a/harness/tests/entity_store_tests.rs
+++ b/harness/tests/entity_store_tests.rs
@@ -1,0 +1,628 @@
+//! Unit tests for `InMemoryEntityStore`.
+//!
+//! The store is the backbone of the agent loop — every entity the agent
+//! reads or writes flows through it.  The existing in-file test only checked
+//! that a fresh store is empty; all the filtering, guard-rail, and
+//! relationship logic was untested.  These tests cover every non-trivial
+//! branch in the `EntityStore` implementation.
+
+use chrono::{Duration, Utc};
+use harness::entities::{
+    context::types::ContextEntity,
+    git::types::{GitBranch, GitRepository},
+    EntityQuery, EntityRelationship, EntityStore, EntityType, InMemoryEntityStore,
+    RelationshipType, TimeRange,
+};
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+fn git_repo() -> GitRepository {
+    GitRepository::new(
+        "git@github.com:example/repo.git".to_string(),
+        "main".to_string(),
+    )
+}
+
+fn git_branch(name: &str) -> GitBranch {
+    GitBranch::new_local(name.to_string(), "abc123def456".to_string())
+}
+
+fn context_entity(task: &str) -> ContextEntity {
+    ContextEntity::new(
+        task.to_string(),
+        vec![model::types::ChatMessage::user(task)],
+        vec![],
+        "done".to_string(),
+        "qwen2.5:0.5b".to_string(),
+    )
+}
+
+// ---------------------------------------------------------------------------
+// Basic CRUD
+// ---------------------------------------------------------------------------
+
+#[tokio::test]
+async fn store_returns_the_assigned_id() {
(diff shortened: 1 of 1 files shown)


DominicBurkart:
Closing — duplicate. `harness/tests/entity_store_tests.rs` already exists on main from commits 4152d66 / 79b7b2e ("Add unit tests for InMemoryEntityStore..."); this PR's 628-line version conflicts add/add. Linked issues remain open.

PR #223, opened 2026-04-17
test(nanna-coder): improve coverage and test quality

DominicBurkart:
## Summary

Adds `harness/tests/tool_coverage_tests.rs` to close the most significant test gaps found in `harness/src/tools.rs` and `harness/src/task.rs`.

### Biggest gap: `CalculatorTool` arithmetic operations

The existing test in `integration_tests.rs` only covered `add` and divide-by-zero. `subtract` and `multiply` had zero test coverage, and the `unknown operation` and `missing operand` error paths were never exercised.

### Other gaps closed

| Area | Gap | New test(s) |
|---|---|---|
| `CalculatorTool` | `subtract`, `multiply`, unknown op, missing operand | 5 tests |
| `ToolRegistry::execute` | `ToolError::NotFound` path never tested | 1 test |
| `WriteFileTool` | No path-security test at all (only `ReadFileTool` had one) | 2 tests (`..` traversal + absolute outside workspace) |
| `ReadFileTool` | No IO error test; no `start_line`-only test; line-number prefix format untested | 4 tests |
| `TaskId` | `Display` / `fmt` impl untested | 2 tests |

All tests are pure unit tests (use `tempfile::tempdir()`); no network or container access required.


harness/tests/tool_coverage_tests.rs
```diff
@@ -0,0 +1,248 @@
...
+        err
+    );
+}
+
+#[tokio::test]
+async fn read_file_start_line_only_reads_to_end() {
+    let temp_dir = tempfile::tempdir().expect("tempdir");
+    let path = temp_dir.path().join("lines.txt");
+    std::fs::write(&path, "a\nb\nc\nd\ne").unwrap();
+
```
DominicBurkart: Tautological assertion. `content.contains("1") && content.contains("alpha")` is satisfied by virtually any non-trivial output — e.g. the string `"alpha\nbeta\ngamma"` alone would pass because "alpha" contains an "a" and... wait, actually "1" is what makes this vacuous: the joined content always contains `2` (the total lines value isn't here, but line 2 always appears), and critically, the implementation uses `format!("{:>6}  {}", i + 1, line)` — a right-aligned 6-column prefix followed by two spaces. A broken implementation that emitted `"prefix_1 alpha"` or even `"1alpha\n2beta"` would pass this test. Since the docstring in `tools.rs:422` specifies the exact format, assert the exact format: e.g. `assert!(content.contains("     1  alpha"))` and `assert!(content.contains("     2  beta"))`. That catches formatting regressions instead of just checking string presence.

---

harness/tests/tool_coverage_tests.rs
```diff
@@ -0,0 +1,248 @@
...
+    );
+    // Ensure the file was not created
+    let _ = std::fs::remove_file(&outside); // clean up just in case
+}
+
+// ---------------------------------------------------------------------------
+// ReadFileTool – IO error and line-range edge cases
+// ---------------------------------------------------------------------------
+
+#[tokio::test]
```
DominicBurkart, line 152: This test has a subtle race/pollution risk: `std::env::temp_dir().join("nanna_write_security_test_outside.txt")` is a fixed path shared across parallel test runs and CI jobs. `cargo test` runs tests concurrently by default; if another test (or a re-run) is cleaning that same path via `remove_file(&outside)` while this test runs, or if a previous failed run left the file, behavior becomes order-dependent. Two fixes: (1) randomize the filename (`format!("nanna_write_sec_{}.txt", uuid::Uuid::new_v4())`), or better (2) use a *second* `tempfile::tempdir()` as the "outside" location so cleanup is automatic and isolated. The current unconditional `remove_file` also silently swallows its error, so a stale leftover from a previous run could mask a real regression where the write *did* succeed.

---

harness/tests/tool_coverage_tests.rs
```diff
@@ -0,0 +1,248 @@
...
+    let outside = std::env::temp_dir().join("nanna_write_security_test_outside.txt");
+    let err = tool
+        .execute(json!({ "path": outside.to_str().unwrap(), "content": "pwned" }))
+        .await
+        .expect_err("absolute path outside workspace must be rejected");
+    assert!(
+        matches!(err, ToolError::PathSecurityViolation { .. }),
+        "expected PathSecurityViolation, got {:?}",
+        err
+    );
```
DominicBurkart, line 143: The negative-assertion is good (checks no file was actually written), but it could be stronger. `temp_dir.path().parent().unwrap().join("escaped.txt")` assumes the traversal target is the tempdir's parent, but the rejection may happen before any path resolution — so this check is mostly theater. Consider also testing traversal variants: `"foo/../../escaped.txt"`, `"./../../escaped.txt"`, encoded variants, and symlink traversal (create a symlink inside the workspace that points outside — this is a real bypass vector that the happy-path `..` rejection does not cover). Without a symlink test, `validate_path_within_workspace` coverage is incomplete against the most common sandbox escape.

---

harness/tests/tool_coverage_tests.rs
```diff
@@ -0,0 +1,248 @@
...
+    std::fs::write(&path, "hello").unwrap();
+    std::fs::remove_file(&path).unwrap();
+
+    let err = tool
+        .execute(json!({ "path": "ghost.txt" }))
+        .await
+        .expect_err("missing file must return an error");
+    // The error is produced by std::fs::read_to_string which maps to ToolError::Io
+    assert!(
+        matches!(err, ToolError::Io(_)),
```
DominicBurkart: `read_file_start_line_only_reads_to_end` asserts `lines_shown == 3` and `total_lines == 5` but never asserts *which* lines were returned. A bug where `start_line` is ignored (returns lines 1-3 instead of 3-5) would still yield `lines_shown=3`. Assert on `result["content"]` — e.g. that it contains `"c"`, `"d"`, `"e"` and does NOT contain `"a"` or `"b"` (watch out for line-number prefixes: `"1  a"` would incorrectly appear to contain `"a"`, so check for `" c"`, `" d"`, `" e"` or parse lines).

---

harness/tests/tool_coverage_tests.rs
```diff
@@ -0,0 +1,248 @@
...
+        .await
+        .expect("start_line only must succeed");
+
+    // Lines 3-5 (c, d, e)
+    assert_eq!(result["lines_shown"], 3, "should return 3 lines from line 3 onwards");
+    assert_eq!(result["total_lines"], 5);
+}
+
+#[tokio::test]
+async fn read_file_line_numbers_are_prefixed_in_content() {
```
DominicBurkart, line 198: `read_file_start_line_1_returns_all_lines` duplicates the existing unit test in `tools.rs` (`result["lines_shown"] == 5` with a similar fixture). More importantly, it asserts only the count — not the equivalence with the no-args case. A bug that treated `start_line: 1` as `start_line: 2` (e.g. off-by-one in the `saturating_sub`) would return 2 lines, but an implementation that treats `start_line: 1` as `start_line: 0` (skipping nothing) would also return 3 lines — both match the assertion. To truly pin down behavior, compare the `content` of `start_line:1` to the `content` of the no-arg call and assert they are identical. Alternatively, delete this test — it is the weakest one in the file and largely redundant.

---

harness/tests/tool_coverage_tests.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+#[test]
+fn task_id_new_display_is_non_empty() {
+    let id = TaskId::new();
+    let s = id.to_string();
+    assert!(!s.is_empty(), "TaskId display must not be empty");
+    // UUID v4 is 36 chars (with hyphens)
+    assert_eq!(s.len(), 36, "UUID v4 should be 36 chars, got: {s}");
+}
```
DominicBurkart, line 259: `task_id_new_display_is_non_empty` asserts `s.len() == 36` — this is effectively re-asserting `uuid::Uuid::new_v4().to_string().len() == 36`, which is testing the `uuid` crate, not our code. `TaskId::new` simply calls `Uuid::new_v4().to_string()` (see `task.rs`), so this test will never catch a bug in *our* code; it will only flag an extremely unlikely upstream format change. Consider either removing this test, or replacing it with something meaningful: e.g. assert two successive `TaskId::new()` calls have distinct `Display` outputs (guards against a misguided refactor to a cached/singleton id). As written, this is classic coverage-chasing.

---

harness/tests/tool_coverage_tests.rs
```diff
@@ -0,0 +1,248 @@
...
+        .expect_err("path traversal must be rejected");
+    assert!(
+        matches!(err, ToolError::PathSecurityViolation { .. }),
+        "expected PathSecurityViolation, got {:?}",
+        err
+    );
+    // File must NOT have been created outside the workspace
+    assert!(
+        !temp_dir.path().parent().unwrap().join("escaped.txt").exists(),
+        "file must not be written outside workspace"
```
DominicBurkart, line 124: `read_file_nonexistent_returns_io_error` creates a file solely to pass path validation and then deletes it. If `validate_path_within_workspace` ever switches to a mode that doesn't require the file to exist (reasonable for a read-file tool that reports IO errors itself), this test's create-then-delete dance becomes confusing and may hide behavior changes. Simpler: pass a path like `"ghost.txt"` that was never created in the first place. If current validation rejects non-existent paths with `PathSecurityViolation` rather than `Io`, that's itself worth asserting — the delete-to-recreate-a-gap trick is a code smell that deserves a code comment or (better) a refactor of `validate_path_within_workspace` to not require existence for read operations.

---

harness/tests/tool_coverage_tests.rs
DominicBurkart, line 1: Overall: the majority of these tests ARE meaningful and close real gaps (subtract/multiply/unknown-op, `ToolRegistry::NotFound`, `WriteFileTool` path-security, `start_line`-only). They are not padding for a 100% codecov target — the PR description's rationale (PR #224) is legitimately served by genuine bug-catching coverage here. Concerns, in priority order: (1) `read_file_line_numbers_are_prefixed_in_content` is tautological and should assert the exact 6-col format; (2) `read_file_start_line_only_reads_to_end` asserts counts but not which lines were returned; (3) `task_id_new_display_is_non_empty` is testing the `uuid` crate, not `TaskId`; (4) the absolute-path-rejection test uses a shared well-known tempfile name that is not race-safe across parallel test runs; (5) no symlink-traversal coverage for the `WriteFileTool` sandbox, which is the more realistic escape vector than plain `..`. No PII or secrets in fixtures — good. No over-mocking — tests use real `CalculatorTool`/`ReadFileTool` impls with `tempfile::tempdir()`, which is the right approach.

---

DominicBurkart (commented):
Test-quality review focused on gaps for coding agents. The PR genuinely closes real gaps (subtract/multiply, ToolRegistry::NotFound, WriteFileTool path security, start_line-only) — not codecov padding. Key issues, priority order: (1) the line-number-prefix assertion is tautological — should pin the exact 6-col format; (2) start_line-only test checks counts but not which lines returned; (3) task_id_new_display_is_non_empty essentially tests the uuid crate; (4) shared fixed tempfile path is not race-safe under parallel cargo test; (5) missing symlink-traversal coverage, the more realistic WriteFileTool escape vector; (6) read_file_nonexistent relies on a create-then-delete dance that is a code smell. No PII/secrets; no over-mocking (real impls + tempdir).

---

PR #224, opened 2026-04-17
ci(codecov): raise patch minimum to 100% + add regression guard (#211)

### LINKED ISSUE #211: reset codecov patch minimum from 90% to 100%
SDLC question: is there any way to avoid coding agents modifying this in prs?

DominicBurkart:
## Summary
- Restores 100% patch-coverage floor on `codecov.yml` (was 90%)
- Adds `scripts/check-codecov-guard.sh` — detects silent relaxations (patch target decreases, `ignore:` list growth)
- Override mechanism: any commit in the PR range carrying a `codecov-override: <reason>` trailer whitelists the change

## Why
Per issue #211's SDLC question, the motivating problem is coding agents silently lowering the patch threshold. The guard:
- Fails closed on target decreases (default behavior)
- Allows target raises and comment-only edits unconditionally
- Unlocks decreases only with an explicit commit-message trailer, which forces a human reviewer to notice

## :warning: Manual step required
The sandbox that produced this PR lacks GitHub App `workflows` permission, so the CI workflow could not be pushed. **Please add `.github/workflows/codecov-guard.yml` manually** with these contents:

```yaml
name: codecov-guard

on:
  pull_request:
    paths:
      - codecov.yml
      - scripts/check-codecov-guard.sh
      - .github/workflows/codecov-guard.yml
  push:
    branches: [main]
    paths:
      - codecov.yml

permissions:
  contents: read

jobs:
  guard:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
      - id: base
        shell: bash
        run: |
          if [ "${{ github.event_name }}" = "pull_request" ]; then
            echo "ref=origin/${{ github.base_ref }}" >> "$GITHUB_OUTPUT"
          else
            echo "ref=HEAD~1" >> "$GITHUB_OUTPUT"
          fi
      - shell: bash
        env:
          BASE_REF: ${{ steps.base.outputs.ref }}
        run: bash scripts/check-codecov-guard.sh
```

Consider also wiring the `guard` job into `all-checks.needs` in `.github/workflows/ci.yml` so it's treated as a required check.

## Local reproduction
```bash
BASE_REF=origin/main bash scripts/check-codecov-guard.sh
```
Verified locally:
- 90% → 100% (this PR): exits 0 (increase is fine)
- 100% → 85% without trailer: exits 1 with clear error
- 100% → 85% with `codecov-override:` trailer in commit: exits 0

## Test plan
- [ ] CI tarpaulin run (from the upcoming first PR to land against this branch) does not regress below 100% on the patched code. If it does, add `#[cfg(not(tarpaulin))]` to the genuinely unreachable branches (noted by the previous comment: container-dispatch paths in `harness/src/workspace.rs`).
- [ ] After workflow is added, confirm it fires on PRs touching `codecov.yml`.

## Defense-in-depth follow-ups (out of scope here)
- Add `.github/CODEOWNERS` requiring owner review on `codecov.yml` — complements the guard by blocking even trailer-carrying changes at merge time. Config-only; document for the repo owner.

Closes #211

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,10 +3,10 @@ coverage:
     project: off
     patch:
       default:
-        # Hard floor — do not relax without explicit override.
-        # Agents modifying this file must include `codecov-override: <reason>`
-        # in a commit message trailer; enforced by
-        # .github/workflows/codecov-guard.yml.
+        # Hard floor — no in-repo override. Only a repo admin's force-merge
+        # can bypass this. The regression guard in
+        # .github/workflows/codecov-guard.yml blocks silent relaxations
+        # (target decreases or `ignore:` list growth).
         target: 100%
 
 ignore:
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,11 +3,11 @@ coverage:
     project: off
     patch:
       default:
-        # Relaxed from 100%: the #[ignore] e2e test and the
-        # runtime-gated test_create_with_container_no_runtime exit early
-        # in CI (no container runtime / Ollama available), so some new
-        # container-dispatch branches are genuinely uncoverable there.
-        target: 90%
+        # Hard floor — do not relax without explicit override.
+        # Agents modifying this file must include `codecov-override: <reason>`
+        # in a commit message trailer; enforced by
+        # .github/workflows/codecov-guard.yml.
+        target: 100%
 
 ignore:
   - "harness/src/main.rs"
--- a/scripts/check-codecov-guard.sh
+++ b/scripts/check-codecov-guard.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+# Guard against silent relaxations of codecov.yml patch thresholds.
+#
+# Fails if, between $BASE_REF (default: origin/main) and HEAD, any `target:`
+# value in codecov.yml DECREASES or the `ignore:` list GAINS entries, UNLESS
+# one of the commits in the range contains a `codecov-override: <reason>`
+# trailer. Target raises and comment-only edits always pass. If codecov.yml
+# did not exist on BASE_REF (first-time add), any values are accepted.
+#
+# Usage (local):   BASE_REF=origin/main bash scripts/check-codecov-guard.sh
+# Usage (CI):      invoked by .github/workflows/codecov-guard.yml
+set -euo pipefail
+
+BASE_REF="${BASE_REF:-origin/main}"
+FILE="codecov.yml"
+
+# Extract all integer `target:` percentages, smallest first.
+min_target() {
+  # $1 = file content on stdin; prints the lowest integer percent or nothing.
+  grep -oE 'target:[[:space:]]*[0-9]+%' "$@" 2>/dev/null \
+    | grep -oE '[0-9]+' \
+    | sort -n \
+    | head -1
+}
+
+# Extract the count of entries under the `ignore:` block.
+ignore_count() {
+  awk '/^ignore:/{flag=1; next} /^[^[:space:]-]/{flag=0} flag && /^[[:space:]]*-/{c++} END{print c+0}' "$@"
+}
+
+if ! git rev-parse --verify "$BASE_REF" >/dev/null 2>&1; then
+  echo "guard: base ref '$BASE_REF' not resolvable; skipping check" >&2
+  exit 0
+fi
+
+# If codecov.yml did not exist on base (first-time add), accept.
+if ! git cat-file -e "$BASE_REF:$FILE" 2>/dev/null; then
+  echo "guard: $FILE is new on this branch; no prior values to compare" >&2
+  exit 0
+fi
+
+OLD_CONTENT="$(git show "$BASE_REF:$FILE")"
+NEW_CONTENT="$(cat "$FILE" 2>/dev/null || true)"
+
+# Quick equality check — no diff, no guard.
+if [ "$OLD_CONTENT" = "$NEW_CONTENT" ]; then
+  exit 0
+fi
+
+OLD_TARGET="$(printf '%s\n' "$OLD_CONTENT" | min_target /dev/stdin || true)"
+NEW_TARGET="$(printf '%s\n' "$NEW_CONTENT" | min_target /dev/stdin || true)"
+OLD_IGNORE="$(printf '%s\n' "$OLD_CONTENT" | ignore_count)"
+NEW_IGNORE="$(printf '%s\n' "$NEW_CONTENT" | ignore_count)"
+
+REGRESSION=0
+REASONS=()
+
+if [ -n "$OLD_TARGET" ] && [ -n "$NEW_TARGET" ] && [ "$NEW_TARGET" -lt "$OLD_TARGET" ]; then
+  REGRESSION=1
+  REASONS+=("patch target decreased: $OLD_TARGET% -> $NEW_TARGET%")
+fi
+
+if [ "$NEW_IGNORE" -gt "$OLD_IGNORE" ]; then
+  REGRESSION=1
+  REASONS+=("ignore: list grew from $OLD_IGNORE to $NEW_IGNORE entries")
+fi
+
+if [ "$REGRESSION" -eq 0 ]; then
+  exit 0
+fi
+
+# Look for a codecov-override: trailer anywhere in the commit range.
+RANGE="${BASE_REF}..HEAD"
+if git log --format=%B "$RANGE" 2>/dev/null | grep -qiE '^codecov-override:[[:space:]]*.+'; then
+  echo "guard: regression detected but codecov-override: trailer found — allowing." >&2
+  for r in "${REASONS[@]}"; do echo "  - $r" >&2; done
+  exit 0
+fi
+
+{
+  echo "guard: codecov.yml regression requires a 'codecov-override: <reason>' commit trailer."
+  for r in "${REASONS[@]}"; do echo "  - $r"; done
+  echo "To override, amend a commit in this PR (or add a new commit) with a trailer like:"
+  echo "    codecov-override: temporarily dropping to unblock release"
+} >&2
+exit 1
--- a/scripts/check-codecov-guard.sh
+++ b/scripts/check-codecov-guard.sh
@@ -2,10 +2,14 @@
 # Guard against silent relaxations of codecov.yml patch thresholds.
 #
 # Fails if, between $BASE_REF (default: origin/main) and HEAD, any `target:`
-# value in codecov.yml DECREASES or the `ignore:` list GAINS entries, UNLESS
-# one of the commits in the range contains a `codecov-override: <reason>`
-# trailer. Target raises and comment-only edits always pass. If codecov.yml
-# did not exist on BASE_REF (first-time add), any values are accepted.
+# value in codecov.yml DECREASES or the `ignore:` list GAINS entries. Target
+# raises and comment-only edits always pass. If codecov.yml did not exist on
+# BASE_REF (first-time add), any values are accepted.
+#
+# There is NO in-repo override. Repository admins who need to relax the
+# threshold must do so via GitHub's force-merge / admin-override path, which
+# is logged and auditable. This script must not provide any way for an
+# automated agent (or a commit trailer) to bypass the regression guard.
 #
 # Usage (local):   BASE_REF=origin/main bash scripts/check-codecov-guard.sh
 # Usage (CI):      invoked by .github/workflows/codecov-guard.yml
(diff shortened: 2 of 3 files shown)


DominicBurkart:
Don't provide an override. Repo admins can force merge; there should not be any way for the agent to drop the coverage percentage.

codecov.yml
```diff
@@ -3,11 +3,11 @@ coverage:
     project: off
     patch:
       default:
-        # Relaxed from 100%: the #[ignore] e2e test and the
-        # runtime-gated test_create_with_container_no_runtime exit early
-        # in CI (no container runtime / Ollama available), so some new
-        # container-dispatch branches are genuinely uncoverable there.
-        target: 90%
+        # Hard floor — do not relax without explicit override.
+        # Agents modifying this file must include `codecov-override: <reason>`
```
DominicBurkart: **Blocker — conflicts with owner direction.** The owner commented 3h ago: *"Don't provide an override. Repo admins can force merge; there should not be any way for the agent to drop the coverage percentage."* This PR's entire design centers on a `codecov-override:` commit trailer that an agent can write into its own commit message. Since the sandbox cannot restrict what trailers an agent chooses, the trailer is agent-reachable and therefore exactly the bypass the owner asked you to remove. The correct design is: **no override mechanism in the script at all** — a decrease in `target:` or growth in `ignore:` must unconditionally fail the check, and humans use admin force-merge as the only escape hatch. Please drop the trailer branch from `scripts/check-codecov-guard.sh` and remove the corresponding instructions from this comment block.

---

codecov.yml
```diff
@@ -3,11 +3,11 @@ coverage:
...
+        # in a commit message trailer; enforced by
+        # .github/workflows/codecov-guard.yml.
+        target: 100%
```
DominicBurkart, line 10: Worth calling out explicitly: the prior comment said 100% was relaxed because `#[ignore]`'d e2e tests and `test_create_with_container_no_runtime` exit early in CI, leaving container-dispatch branches in `harness/src/workspace.rs` genuinely uncoverable without a runtime. Raising to 100% without first annotating those branches with `#[cfg(not(tarpaulin))]` (or equivalent) will make CI *unhittable* for any PR that touches those files — even purely legitimate refactors. The PR description acknowledges this as a "test plan" checkbox but the fix isn't in this PR. Recommend either (a) land the `cfg(not(tarpaulin))` annotations in this same PR, or (b) sequence them first and rebase this one on top. Otherwise the next PR to touch workspace.rs will be stuck.

---

scripts/check-codecov-guard.sh
```diff
@@ -0,0 +1,86 @@
...
+
+if [ "$REGRESSION" -eq 0 ]; then
+  exit 0
+fi
+
+# Look for a codecov-override: trailer anywhere in the commit range.
+RANGE="${BASE_REF}..HEAD"
+if git log --format=%B "$RANGE" 2>/dev/null | grep -qiE '^codecov-override:[[:space:]]*.+'; then
+  echo "guard: regression detected but codecov-override: trailer found — allowing." >&2
+  for r in "${REASONS[@]}"; do echo "  - $r" >&2; done
```
DominicBurkart: Remove this whole override block per the owner's comment on the PR thread. An agent that is instructed to raise coverage artificially can just add `codecov-override: <anything>` to its own commit and silently bypass the guard — this is precisely the threat model issue #211 asks about. Without the override, the check becomes "fails-closed, admin-only escape via force-merge," which is what was requested.

---

scripts/check-codecov-guard.sh
```diff
@@ -0,0 +1,86 @@
...
+
+BASE_REF="${BASE_REF:-origin/main}"
+FILE="codecov.yml"
+
+# Extract all integer `target:` percentages, smallest first.
+min_target() {
+  # $1 = file content on stdin; prints the lowest integer percent or nothing.
+  grep -oE 'target:[[:space:]]*[0-9]+%' "$@" 2>/dev/null \
+    | grep -oE '[0-9]+' \
+    | sort -n \
```
DominicBurkart, line 26: `min_target` only catches integer percentages (`[0-9]+%`). An agent circumventing the guard could write `target: 99.9%` or `target: auto` (codecov supports both) and `min_target` would return empty, the `-n` guard on line 62 would be false, and the regression check silently passes. Please broaden the regex to `[0-9]+(\.[0-9]+)?%` and explicitly reject `auto`/non-numeric values as regressions when the old value was numeric.

---

scripts/check-codecov-guard.sh
```diff
@@ -0,0 +1,86 @@
...
+    | head -1
+}
+
+# Extract the count of entries under the `ignore:` block.
+ignore_count() {
+  awk '/^ignore:/{flag=1; next} /^[^[:space:]-]/{flag=0} flag && /^[[:space:]]*-/{c++} END{print c+0}' "$@"
```
DominicBurkart, line 32: `ignore_count` regex `/^[^[:space:]-]/` treats any top-level key that doesn't start with space or `-` as terminating the ignore block — but `#` comment lines inside the block, or lines starting with `?` (YAML complex mapping key), or blank lines, will also match unexpectedly. More importantly, this counts only `- `-prefixed list entries; an agent could move to `ignore: ["a", "b", "c"]` flow-style and `ignore_count` would return 0, *appearing* to shrink the list. Recommend parsing with a real YAML tool (`yq` is already common in CI) rather than awk/grep.

---

scripts/check-codecov-guard.sh
```diff
@@ -0,0 +1,86 @@
...
+}
+
+if ! git rev-parse --verify "$BASE_REF" >/dev/null 2>&1; then
+  echo "guard: base ref '$BASE_REF' not resolvable; skipping check" >&2
```
DominicBurkart, line 36: Silent `exit 0` when the base ref can't be resolved is too permissive — in CI this happens when fetch-depth is wrong, and you'd rather fail loudly than accept any change. The workflow uses `fetch-depth: 0` so this branch shouldn't hit in CI, but for defense-in-depth change to `exit 1` with a clear message, matching the "fail closed" intent of the owner's directive.

---

scripts/check-codecov-guard.sh
```diff
@@ -0,0 +1,86 @@
...
+  exit 0
+fi
+
+# If codecov.yml did not exist on base (first-time add), accept.
+if ! git cat-file -e "$BASE_REF:$FILE" 2>/dev/null; then
+  echo "guard: $FILE is new on this branch; no prior values to compare" >&2
```
DominicBurkart, line 42: Similarly, "first-time add" blanket acceptance is a foot-gun: if an agent deletes `codecov.yml` and re-adds it on the same branch (e.g., via a squash-merge that makes the file look new relative to some base), this branch would accept a relaxed target. For a guard whose threat model is *adversarial agent behavior*, prefer to fail closed here too — a legitimate first add can go in via admin force-merge.

---

scripts/check-codecov-guard.sh
```diff
@@ -0,0 +1,86 @@
...
+  exit 0
+fi
+
+{
+  echo "guard: codecov.yml regression requires a 'codecov-override: <reason>' commit trailer."
+  for r in "${REASONS[@]}"; do echo "  - $r"; done
+  echo "To override, amend a commit in this PR (or add a new commit) with a trailer like:"
+  echo "    codecov-override: temporarily dropping to unblock release"
+} >&2
+exit 1
```
DominicBurkart, line 82: **Workflow file missing from PR.** The PR description says the sandbox lacked `workflows` permission so `.github/workflows/codecov-guard.yml` must be added manually. This means merging this PR as-is installs a guard script that nothing invokes — and it relies on a human remembering to copy-paste the YAML from the PR body. Suggest one of: (a) block merge until the workflow lands (base this PR on a branch that already has it), (b) invoke the guard from an existing workflow (e.g., a step in `ci.yml`) that the sandbox *can* modify, or (c) at minimum, add a pre-commit hook so local runs catch regressions even without CI. As written, there is a real risk the script ships and is never exercised.

---

DominicBurkart, line 82: Also consider: the guard watches only `codecov.yml`. An agent can bypass it entirely by adding `#[cfg(not(tarpaulin))]` or `#[cfg_attr(coverage_nightly, coverage(off))]` to newly-added code, or by marking tests `#[ignore]`, to hide uncovered lines from the patch metric without ever touching `codecov.yml`. A complete defense against issue #211's threat model also needs a guard on net growth of those annotations in source files. Out of scope for this PR, but worth filing as a follow-up so the 100% target actually means what it says.

---

DominicBurkart (commented):
Code review from one of three parallel reviewers. The direction (raise patch floor to 100% + add a regression guard) is correct and well-aligned with issue #211, but the current design conflicts with explicit owner direction posted on this PR 3 hours ago and has several bypass vectors that undermine the stated threat model.

**Must-fix**
1. **Remove the `codecov-override:` trailer mechanism.** Per the owner's comment: *"Don't provide an override. Repo admins can force merge; there should not be any way for the agent to drop the coverage percentage."* An agent-writable trailer is by definition an agent-reachable bypass. Fail closed; admin force-merge is the only escape hatch.
2. **Land the tarpaulin exclusions first (or in this PR).** The prior 90% comment explicitly noted container-dispatch branches in `harness/src/workspace.rs` are uncoverable in CI without a runtime. Merging 100% without first annotating those paths with `#[cfg(not(tarpaulin))]` will make CI unhittable for legitimate refactors of those files.
3. **Workflow is not in this PR.** Merging installs a guard script nothing invokes. Either base on a branch that has the workflow, or wire it into `ci.yml` (which the sandbox *can* edit), or add a pre-commit hook.

**Should-fix (regex/YAML gaps that are themselves bypasses)**
- `min_target` ignores `99.9%` and `auto`; broaden to `[0-9]+(\.[0-9]+)?%` and treat non-numeric as regression.
- `ignore_count` only counts block-style `- ` entries; flow-style `ignore: ["a","b"]` reports 0. Use `yq`.
- `exit 0` on unresolvable base ref / first-time add is too permissive for an adversarial threat model — prefer fail-closed.

**Out-of-scope follow-up**
- Guard only watches `codecov.yml`. An agent can bypass entirely by sprinkling `#[cfg(not(tarpaulin))]` / `#[ignore]` in source. Worth a separate guard on net growth of those annotations.

**Is 100% untenable?** Not inherently — the repo was at 100% before #211's parent relaxation — but it *will* be untenable for `harness/src/workspace.rs` until the tarpaulin cfg-gates ship. Address must-fix #2 and 100% is achievable.

---

DominicBurkart:
Addressed in 1ab73e1 — removed the `codecov-override:` trailer bypass from `scripts/check-codecov-guard.sh` and updated the `codecov.yml` comment. The guard now fails closed on any target decrease or `ignore:` growth with no in-repo escape hatch; relaxations must go through a repo-admin force-merge.

---

codecov.yml
DominicBurkart: Nope. You are describing a test design failure. CI/CD is expected to be functional before merging business logic changes.

scripts/check-codecov-guard.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
```
DominicBurkart, line 1: how is this a guardrail? what stops an agent from changing this too?

most but not all agents working in this repo can't change the github workflows. a big issue is that the codecov configuration doesn't live there, is can be manipulated.

DominicBurkart:
**Iteration attempt under job `DqQQS` timed out before pushing.** PR head SHA unchanged (`f1f2256`). Captured action plan below so the next CRON run (or a human) can pick up without re-analysing the threads.

### Required redesign (per owner comments)

1. **Remove the `codecov-override:` trailer mechanism entirely.** Owner was explicit (twice): agents must not be able to drop the coverage target; admin force-merge is the only escape hatch. Delete the trailer branch from `scripts/check-codecov-guard.sh` and remove the corresponding instructions from `codecov.yml` comments + this PR body.

2. **Fail closed on edge cases** in `scripts/check-codecov-guard.sh`:
   - Missing base ref → `exit 1` (not `exit 0`).
   - First-time-add of `codecov.yml` → `exit 1` (not `exit 0`).
   - `min_target` regex: broaden to `[0-9]+(\.[0-9]+)?%` AND reject `auto` / non-numeric targets as regressions when the prior value was numeric.
   - `ignore_count`: replace awk with `yq` (already available on `ubuntu-latest`) to correctly handle flow-style `ignore: ["a","b"]`.

3. **Cover the second bypass route:** add a diff-scan step (same workflow, separate step) that fails on net new occurrences of `#[cfg(not(tarpaulin))]`, `#[cfg_attr(coverage_nightly, coverage(off))]`, `coverage(off)`, and `#[ignore]` in `*.rs`. Use `git diff --unified=0 ${BASE_SHA}...HEAD -- '*.rs' | grep -E ...`.

4. **Acknowledge the deeper concern** owner raised ("the codecov configuration doesn't live there, it can be manipulated"): the guard lives in-repo and is therefore agent-writable. Mitigations:
   - Inline the guard bash directly into the workflow YAML (no `scripts/` file), since most agents lack `workflows` permission.
   - Suggest `CODEOWNERS` requiring owner review on both `codecov.yml` AND `.github/workflows/codecov-guard.yml` — documented as a follow-up for the owner to install manually.

5. **Do not** rebase the "workspace.rs uncoverable" argument into the PR body — owner explicitly rejected it as "a test design failure."

No code changes were pushed in this cycle. Self-closing as idempotent no-op; next run will regenerate this plan against the same SHA if no human intervention happens.

---

DominicBurkart:
Applied `.github/workflows/codecov-guard.yml` in 489229a. Leaving the inline copy in the PR body as audit trail.

DominicBurkart:
Closing — already on main. The codecov 100% patch minimum and regression guard are in place via `.github/workflows/codecov-guard.yml` (already on main); this PR's script-extraction refactor is stylistic, not substantive. Linked issues remain open.

Issue #200, opened 2026-04-12
mcp-serve bypasses SharedModelPool, requires host Ollama


DominicBurkart:
## Summary

`harness mcp-serve` connects directly to a host Ollama instance at `http://localhost:11434` instead of spinning up the Nanna-managed model container described in `ARCHITECTURE.md`. This forces users installing nanna as a local MCP server to install and run Ollama on the host, which contradicts the documented containerized design.

## Current behavior

`harness/src/main.rs:478-499` (`run_mcp_server`) constructs an `OllamaProvider` from `OllamaConfig::default()` (base URL `http://localhost:11434`) and passes it directly to `NannaMcpServer`. No container is started; `SharedModelPool` is never touched.

```rust
async fn run_mcp_server(model: &str, max_iterations: usize) -> Result<(), Box<dyn std::error::Error>> {
    let config = OllamaConfig::default();
    let provider = Arc::new(OllamaProvider::new(config)?);
    let task_manager = Arc::new(TaskManager::default());
    // ...
    let server = NannaMcpServer::new(task_manager, provider, model.to_string(), max_iterations);
    server.run_stdio().await?;
}
```

## Expected behavior

Per `ARCHITECTURE.md`, the "Nanna Model" container should be managed by Nanna. `SharedModelPool::get_or_start()` (`harness/src/container.rs:584-653`) already implements ref-counted container lifecycle against `ContainerConfig::default()` (image `ollama/ollama:latest`, port 11435, name `nanna-coder-test`). `run_mcp_server` should:

1. Acquire a pool handle via `SharedModelPool::get_or_start()`
2. Point `OllamaConfig` at the pool's container URL (not the host's 11434)
3. Hold the handle for the lifetime of the MCP session so the container stays up
4. Release on shutdown (signal handling for SIGINT/SIGTERM over stdio)

## Related

- #60 — Expose Nanna via MCP (parent)
- #63 — Shared model container lifecycle
- #64 — Dev container creation / workspace isolation (image-builder work likely needed so the container ships with the target model pre-pulled or pulls on first use)
- #127 — Smoke test: Claude.ai/code can install and use nanna
- #195 — Initial local-install PR; ships host-Ollama smoke test, defers container wiring to this issue

## Scope notes

The image-builder side may need work so that the container either (a) ships with models baked in, or (b) pulls on first `get_or_start()`. Signal handling over MCP stdio also needs to flush the pool release so containers don't leak.

PR #195, opened 2026-04-11
feat: local MCP server setup for Claude Code

DominicBurkart:
## Summary
- Increase Ollama timeout from 30s to 300s to support larger models (gemma4:e4b)
- Add one-liner to README for registering nanna as an MCP server in Claude Code
- Gitignore `.mcp.json` (contains machine-specific absolute paths)
- Fix pre-commit hook: unset `GIT_INDEX_FILE`/`GIT_DIR`/`GIT_WORK_TREE` at hook start so integration tests that spawn git subprocesses don't corrupt the parent repo's index

## Test plan
- [x] All 386 tests pass (`cargo nextest run --workspace --all-features`)
- [x] Clippy clean, fmt clean
- [x] MCP handshake verified: `harness mcp-serve --model gemma4:e4b` responds to `initialize` with correct protocol version and capabilities
- [ ] Pull gemma4:e4b (blocked: requires Ollama upgrade from 0.13.5 to newer version)
- [ ] End-to-end: Claude Code assigns a task via nanna MCP and retrieves result

Relates to #127, #129, #60

PR #79, opened 2026-03-19
feat(harness): shared model container lifecycle management (#63)

DominicBurkart:
## Summary

- Fixes `OllamaProvider` to parse host+port from `base_url` instead of hardcoding port 11434
- Adds `ContainerConfig::ollama_host_url()` helper to derive the Ollama URL from the port mapping
- Replaces blind `sleep(startup_timeout)` with a polling health-check loop against `http://localhost:{port}/api/tags`
- Rewrites `SharedModelPool` with `tokio::sync::Mutex` and a RAII `ModelGuard` to eliminate the TOCTOU race and ensure the container is stopped only when the last guard is dropped
- Wires `SharedModelPool` into `run_mcp_server`; adds `--host-port` and `--ollama-url` CLI flags to `mcp-serve`

## Test plan

- [ ] `cargo test -p model` — 35 unit tests pass (parse_ollama_url, provider creation)
- [ ] `cargo test -p harness --lib container` — 13 tests pass (ollama_host_url, health_check_host, SharedModelPool ref counting)
- [ ] `cargo test -p harness --lib mcp` — 10 tests pass
- [ ] `cargo test -p harness --bin harness` — 3 CLI parsing tests pass (default host-port, custom host-port, ollama-url bypass)
- [ ] `cargo clippy --all-features -- -D warnings` — clean
- [ ] `cargo fmt --all -- --check` — clean
- [ ] Manual: `cargo run -- mcp-serve --model qwen3:0.6b` starts container, health check passes, MCP server responds
- [ ] Manual: `cargo run -- mcp-serve --ollama-url http://localhost:11434` connects to external Ollama, no container started

Closes #63

PR #225, opened 2026-04-17
test(harness): integration test spawning mcp-serve over stdio (#201)

### LINKED ISSUE #201: Add integration test that spawns harness mcp-serve binary and speaks MCP over a pipe
## Summary

The MCP stdio framing bug fixed in #195 (follow-up commit) wasn't caught by unit tests because the existing tests call `handle_request` directly or use an in-memory `serve()` helper. A regression in the actual stdio I/O layer — or in the compiled binary's argument parsing, startup, logging-to-stderr, signal handling, etc. — could still surface only when a real MCP client connects.

Add an integration test that:

1. Spawns `target/<profile>/harness mcp-serve --model <tiny-model>` as a child process via `tokio::process::Command` with piped stdin/stdout.
2. Writes newline-delimited JSON-RPC requests (`initialize`, `tools/list`, `notifications/initialized`) to the child's stdin.
3. Reads newline-delimited JSON-RPC responses from stdout and asserts:
   - `initialize` returns `protocolVersion: "2024-11-05"` and `capabilities.tools` object
   - `tools/list` returns the 6 expected tool names
   - notifications produce no response bytes
   - No `Content-Length:` headers anywhere in the output stream
4. Closes stdin and asserts the child exits cleanly within a short timeout.

## Why

Without this, a framing regression (or any other stdio-layer breakage) is invisible to CI and only surfaces when a user tries to connect Claude Code / another MCP client — exactly what happened in #195 before the follow-up fix.

## Notes

- Use a fake/mock provider or a model that doesn't actually require Ollama — the test should not require `ollama serve` to be running. Easiest: add a `--no-provider` test-only flag, or make `mcp-serve` tolerate provider health-check failure at startup so `initialize` / `tools/list` (which don't hit the model) still work. Alternative: point at a mock HTTP server.
- Should live in `harness/tests/mcp_stdio_integration.rs`.
- Keep timeouts generous (a few seconds) to avoid flakes on slow CI runners.

## Related

- #195 — PR where this gap was discovered
- #127 — Parent smoke-test issue
- #200 — Container wiring (separate)

DominicBurkart:
## Summary
Closes the gap flagged by #195: unit tests call `handle_request` directly and miss stdio-framing regressions in the compiled `harness` binary. This PR spawns the real binary as a child process and exchanges JSON-RPC over piped stdin/stdout.

## What it asserts
- `initialize` → `protocolVersion: "2024-11-05"` + `capabilities.tools` is an object
- `initialize` does NOT echo `clientCapabilities` back (regression guard)
- `tools/list` returns exactly the 6 expected tool names (sorted-compare: `assign_task`, `cancel_task`, `get_result`, `list_tasks`, `onboard_repo`, `poll_task`)
- `notifications/initialized` produces no response
- stdout contains no `Content-Length:` header
- exactly 2 non-empty JSON lines on stdout for the session
- child exits cleanly (status 0) within 10s of stdin close

## Provider isolation
Introduces `NANNA_MCP_MOCK_PROVIDER=1` — when set, `run_mcp_server` swaps in `harness::mcp::mock_provider::MockProvider` (satisfies `ModelProvider` without a live backend). `chat` errors loudly; `health_check`/`list_models` succeed cheaply. The test never calls `tools/call`, so `chat` is never invoked.

**PR #79 / #200 coordination**: the env-var bypass is the first code in `run_mcp_server` with a sentinel comment. If PR #79 lands second, the rebase must preserve the bypass above any `SharedModelPool::get_or_start()` call so this test doesn't need Podman.

## Flake resistance
- 10s timeouts on every read + final `wait()`
- `kill_on_drop(true)` on the child
- Background task drains stderr so a chatty logger can't fill the 64KB pipe and deadlock stdout
- `RUST_LOG=warn` minimizes stderr volume
- `OLLAMA_HOST` env-unset on the child so a developer's local Ollama can't be contacted

## Feedback loops
Verified locally under `cargo test -p harness --test mcp_stdio_integration` (0.15s). `cargo fmt --check` + `cargo clippy -p harness -- -D warnings` green. Auto-discovered by the existing workspace nextest run in `.github/workflows/ci.yml:143`, so every PR will exercise it.

## Test plan
- [x] `cargo test -p harness --test mcp_stdio_integration` → `ok`
- [x] `cargo fmt --check` green
- [x] `cargo clippy -p harness --all-targets -- -D warnings` green
- [ ] CI runs it against both the default workspace and on PR #79's branch to confirm coordination contract

## Out of scope
- `tools/call` round-trips (need real model; covered by eval suite)
- Agent loop execution (#200 / eval tests)
- Container startup/health (#200)
- Signal handling (SIGINT/SIGTERM) — separate issue if needed

Closes #201


harness/src/main.rs
```diff
@@ -479,13 +479,23 @@ async fn run_mcp_server(
     model: &str,
     max_iterations: usize,
 ) -> Result<(), Box<dyn std::error::Error>> {
+    // NANNA_MCP_MOCK_PROVIDER test hook (issue #201): swap in a mock provider
+    // so the mcp_stdio_integration test can exercise stdio framing without a
+    // live Ollama or container. Keep at the top of this function — any future
+    // SharedModelPool wiring must preserve this bypass.
     use harness::mcp::NannaMcpServer;
```
DominicBurkart, line 577: Shipping an env-var hook that silently swaps the real provider for a mock in the release binary is a footgun. Anyone (CI misconfiguration, a shell profile, a malicious parent process) setting `NANNA_MCP_MOCK_PROVIDER=1` turns an `mcp-serve` invocation into a no-op model where `chat` errors with a cryptic "MockProvider::chat called" message. Two low-cost mitigations, pick one:

1. Gate the branch behind `#[cfg(debug_assertions)]` (or a `test-hooks` cargo feature that's off in release profiles), so release binaries don't carry the bypass path at all.
2. Keep the env var but emit an unmissable `warn!`/`eprintln!` banner when activated ("MCP running with mock provider — chat will fail"), and refuse to activate if `cfg!(not(debug_assertions))` unless a second confirmation var is also set.

Option 1 is cleaner given `mock_provider` is already `#[doc(hidden)]`. The integration test is compiled in test mode against the binary, so `cfg!(debug_assertions)` will be true there. The only reason to keep this in release is if you expect external contributors to run the test against a `--release` build — which isn't the documented feedback loop.

---

harness/src/mcp/mod.rs
```diff
@@ -1,4 +1,6 @@
 pub mod handlers;
+#[doc(hidden)]
+pub mod mock_provider;
```
DominicBurkart, line 3: `#[doc(hidden)] pub mod mock_provider` still compiles `MockProvider` into every release build and exposes it as a public item (docs hidden is not the same as not-public). Prefer `#[cfg(any(test, feature = "test-hooks"))] pub mod mock_provider;` paired with the same cfg on the bypass branch in `run_mcp_server`. That keeps the symbol out of shipped binaries and removes any chance of a downstream crate depending on `harness::mcp::mock_provider::MockProvider` by accident.

---

harness/tests/mcp_stdio_integration.rs
```diff
@@ -0,0 +1,209 @@
...
+use std::process::Stdio;
+use std::time::Duration;
+
+use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncWriteExt, BufReader};
+use tokio::process::Command;
+use tokio::time::timeout;
+
+const RPC_TIMEOUT: Duration = Duration::from_secs(10);
+
+fn harness_bin() -> std::path::PathBuf {
```
DominicBurkart, line 25: `RPC_TIMEOUT = 10s` is applied individually to every `recv_line`, the final `read_to_string`, and `child.wait()`. That is a theoretical wall-clock of up to 40s per test run, but worse: on a cold CI runner the *first* `recv_line` (which waits for the child to compile-link-dynamic-load + parse args + build the server) is the most likely to blow the 10s budget. Issue #201 explicitly asks for generous timeouts "to avoid flakes on slow CI runners." Consider 30s, or make it configurable via `NANNA_MCP_STDIO_TEST_TIMEOUT_SECS` so CI can bump it without a code change. Also: cargo test by default does not print timeout context — include the elapsed time in the TimedOut error for post-mortem debugging.

---

harness/tests/mcp_stdio_integration.rs
```diff
@@ -0,0 +1,209 @@
...
+    let _ = timeout(RPC_TIMEOUT, reader.read_to_string(&mut remaining))
+        .await
+        .expect("drain remaining stdout");
+
+    let status = timeout(RPC_TIMEOUT, child.wait())
+        .await
+        .expect("child wait timed out")
+        .expect("child wait failed");
+
+    let stderr_contents = stderr_handle.await.unwrap_or_default();
```
DominicBurkart, line 182: Ordering races on the framing invariants: after `drop(stdin)`, the child must (a) finish any in-flight write to stdout, (b) close stdout, (c) exit. `reader.read_to_string(&mut remaining)` returns when stdout EOFs, which happens when the child process exits or explicitly closes the fd. If the child's graceful shutdown drops the stdout handle *after* the stdin-close is observed but *before* `wait()` completes, you might read partial JSON. In practice tokio flushes on drop so it's fine — but this ordering is not documented and is the #1 source of flaky stdio-process tests in my experience. Consider adding a short `tokio::time::sleep(Duration::from_millis(50))` guard, or, better: after `recv_line` for `tools/list`, send a canary `initialize` with id=3 and assert you get the response back before closing stdin. That proves liveness independent of shutdown timing.

---

harness/tests/mcp_stdio_integration.rs
```diff
@@ -0,0 +1,209 @@
...
+        .as_array()
+        .expect("tools/list: result.tools must be an array");
+    assert_eq!(tools.len(), 6, "tools/list: expected exactly 6 tools");
+    let mut names: Vec<&str> = tools.iter().map(|t| t["name"].as_str().unwrap()).collect();
+    names.sort();
+    let expected = [
+        "assign_task",
+        "cancel_task",
+        "get_result",
+        "list_tasks",
```
DominicBurkart, line 163: The hard-coded `expected = [...]` with `len == 6` will break the moment anyone adds a 7th MCP tool (#200 container wiring adds some, future `search_repo`, etc.). That's partially intentional as a regression guard, but future agents landing new tools will hit this failure and may "fix" it by blindly updating the array without understanding *why* it was asserting exactly 6. Recommend one of:
- Add a comment above `expected` explaining "If you're adding a tool, update this list *and* update ARCHITECTURE.md's tool inventory."
- Split into two asserts: a positive `names.contains("assign_task")` (etc.) for the current set (tests the tools are present), and a separate `assert_eq!(names.len(), N, ...)` whose failure message explicitly says "update me plus ARCHITECTURE.md".

Option 2 gives better failure messages and keeps the no-regression property.

---

harness/tests/mcp_stdio_integration.rs
```diff
@@ -0,0 +1,209 @@
...
+            "method": "tools/list"
+        }),
+    )
+    .await
+    .expect("write tools/list");
+
+    let tools_line = recv_line(&mut reader)
+        .await
+        .expect("read tools/list response");
+    let tools_resp: serde_json::Value =
```
DominicBurkart, line 150: `tools.iter().map(|t| t["name"].as_str().unwrap())` — the `unwrap()` on a tool missing a `name` (or with a non-string name) produces a panic with no useful location info. Since this is the *exact* kind of regression the test is meant to catch, a typed message helps future debuggers:

```rust
.map(|t| t["name"].as_str().unwrap_or_else(|| panic!("tool missing string 'name': {t:?}")))
```

Same pattern on line 139 (`tools_resp["result"]["tools"].as_array().expect(...)`) is already good — just mirror it here.

---

harness/tests/mcp_stdio_integration.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+#[tokio::test]
+async fn mcp_stdio_end_to_end_initialize_and_tools_list() {
+    let mut child = Command::new(harness_bin())
+        .args(["mcp-serve", "--model", "mock:test"])
+        .env("NANNA_MCP_MOCK_PROVIDER", "1")
+        .env("RUST_LOG", "warn")
+        // Prevent accidental contact with a developer's local Ollama.
+        .env_remove("OLLAMA_HOST")
```
DominicBurkart, line 60: Only `OLLAMA_HOST` is env_removed. A developer's shell may carry any of: `OLLAMA_MODELS`, `OLLAMA_API_KEY`, `HTTPS_PROXY`, `HTTP_PROXY`, `no_proxy`, plus `NANNA_*` config vars (if any exist). For hermeticity, prefer `.env_clear()` followed by `.env("PATH", std::env::var_os("PATH").unwrap_or_default())` + the two vars you actually set. This also reduces the chance that a maintainer's local `NANNA_MCP_MOCK_PROVIDER=0` (or similar disable) in `.envrc`/direnv sabotages the hook. Minimum PATH + explicit allowlist is the standard hermetic-subprocess recipe.

---

harness/tests/mcp_stdio_integration.rs
```diff
@@ -0,0 +1,209 @@
...
+        .expect("drain remaining stdout");
+
+    let status = timeout(RPC_TIMEOUT, child.wait())
+        .await
+        .expect("child wait timed out")
+        .expect("child wait failed");
+
+    let stderr_contents = stderr_handle.await.unwrap_or_default();
+    assert!(
+        status.success(),
```
DominicBurkart, line 184: `full_stdout = format!("{init_line}{tools_line}{remaining}")` — both `init_line` and `tools_line` come out of `read_line` which keeps the trailing `\n`, so concatenation is fine. But if the server ever sends a response without a trailing newline (framing regression), `read_line` will block until the next `\n` or EOF — which is the exact regression this test tries to catch. It won't fail loudly; it will hang until RPC_TIMEOUT. Consider asserting `init_line.ends_with('\n')` (and same for `tools_line`) right after the read, so a "server forgot to write `\n`" regression shows up as a direct assertion failure rather than a 10s hang. That turns a latent flake signal into an immediate diagnostic.

---

harness/src/mcp/mock_provider.rs
```diff
@@ -0,0 +1,36 @@
...
+use model::types::{ChatRequest, ChatResponse, ModelInfo};
+
+#[doc(hidden)]
+pub struct MockProvider;
+
+#[async_trait]
+impl ModelProvider for MockProvider {
+    async fn chat(&self, _: ChatRequest) -> ModelResult<ChatResponse> {
+        Err(model::provider::ModelError::Unknown {
+            message: "MockProvider::chat called — only initialize/tools/list are supported"
```
DominicBurkart, line 20: Nit: `ModelError::Unknown` is a poor fit — it's semantically "I don't know what went wrong", whereas this is a precise "don't call this". If `model::provider::ModelError` has a `Misconfigured`/`InvalidProvider`/`Unsupported` variant, prefer it. If it doesn't, add one — this same issue will come up for any future test shim. Also, the message should not include an em-dash via non-ASCII if the error ever lands in logs processed by tools that don't handle UTF-8 well (unlikely but free to fix: use `--`).

---

DominicBurkart (commented):
Integration test is well-scoped to #201's intent and the flake-resistance measures (kill_on_drop, stderr drain, per-op timeouts, OLLAMA_HOST unset) show careful thought. Main concerns:

1. **Production bloat / footgun**: `NANNA_MCP_MOCK_PROVIDER` env hook + `mock_provider` module ship in the release binary. Gate behind `#[cfg(debug_assertions)]` or a `test-hooks` cargo feature.
2. **Flakiness**: 10s per-op timeout is tight for cold CI runners (#201 explicitly asks "generous"); consider 30s or env-configurable. Ordering between `drop(stdin)` → stdout EOF → `wait()` is correct but undocumented; a liveness canary pre-shutdown would harden it.
3. **Maintenance**: hard-coded `len == 6` will break on future tool additions without a breadcrumb pointing to ARCHITECTURE.md.
4. **Hermeticity**: only `OLLAMA_HOST` is scrubbed; prefer `env_clear()` + explicit allowlist.

Inline comments have specific suggestions. Nothing blocks merge conceptually but (1) and (2) are worth addressing before this lands on main.

---

Issue #219, opened 2026-04-14
[obs] logs


DominicBurkart:
Allow users to set a file to write nanna logs to using a CLI flag ([currently experimental] / off by default). 

Structural requirements:
- use json structured logs
- ISO 8601

Content:
- User prompt / triggering call to nanna AND session identifiers.
- Every tool call by nanna, including name, arguments, timestamps, duration, and result payload.
- Model decisions that affect control flow, such as branching, retries, and task decomposition.
- Errors, stack traces, timeouts, and cancellation reasons.
- Key environment details, like repo state, branch, commit hash, dependency versions, and working directory.
- Outcome markers, such as success, partial success, or failure, plus any artifacts written.

We're not going to build a backend for these, so there are no immediate off-machine secret/PII concerns. 

Every log should be tagged with the container that it's running in.

DominicBurkart:
Breaking this out into six PRs so the change set stays reviewable and later phases can land in parallel. PR1 is open now as #229 (draft).

**Planned PR boundaries:**

| PR | Scope | Depends on |
|---|---|---|
| **PR1** (#229, open) | deps (`tracing-appender`, `json`/`time` features), global `--log-file` / `--log-file-level` CLI flags, `harness/src/logging.rs` scaffolding, `session.start` + `env.snapshot` events, `{container, service, version}` default fields, regression test for "no flag → zero file writes" | — |
| **PR2** | `SessionId` propagation through `AgentContext` + MCP `TaskId`; top-level `tracing::info_span!("session", ...)` so every downstream event inherits `session_id` and `container`; container-tag `Layer` impl; integration test for `HOSTNAME`-absent fallback | PR1 |
| **PR3** | Agent-loop emissions in `harness/src/agent/mod.rs`: `state.transition`, `llm.retry`, `model.decision`, `run.outcome`, `run.cancelled`. Unit tests per emission site via `tracing-test` | PR1 |
| **PR4** | Tool-dispatch emissions: `tool.call.start` / `tool.call.end` with 4 KB-truncated result previews at both agent and non-agent (chat-mode) dispatch sites; `artifact.written` events inside `WriteFileTool::execute` and post-exec scan in `RunCommandTool`; secret-key redaction | PR1 |
| **PR5** | `harness/benches/logging.rs` with criterion (disabled-overhead, single-field, 2 KB tool-call); `workflow_dispatch` bench job storing artifacts without gating | PR1 |
[...]

PR #229, opened 2026-04-17
feat(harness): JSON structured logging to file via --log-file (#219 PR1)

DominicBurkart:
## Summary

This is **PR1 of 6** for #219 (`[obs] logs`). Scope is deliberately tight: dependency additions, a global `--log-file` CLI flag, a new `harness/src/logging.rs` scaffolding module, and a minimal pair of structured events (`session.start`, `env.snapshot`) so downstream PRs can extend the schema without reshaping existing records.

### What ships in PR1
- `harness/Cargo.toml`: add `tracing-appender = "0.2"`; enable `json` + `time` features on `tracing-subscriber`; enable `env` feature on `clap` so the new flags can read `NANNA_LOG_FILE` / `NANNA_LOG_FILE_LEVEL`.
- `harness/src/logging.rs`:
  - `init(log_file, log_file_level) -> Result<Option<WorkerGuard>>`: always installs the human `fmt` layer on stderr (preserves current UX and coexists with any future stderr spinner). When a log file path is supplied, additionally layers `fmt::layer().json().flatten_event(true).with_timer(UtcTime::rfc_3339()).with_writer(non_blocking)` via `tracing_appender::non_blocking` with `NonBlockingBuilder::lossy(false)` so overflow blocks rather than drops.
  - `session_start(session_id, user_prompt)` and `env_snapshot(workspace_root, git_branch, git_head)` helpers that emit structured events with `{container, service, version}` default fields resolved from `$HOSTNAME` → `$CONTAINER_NAME` → `"host"`.
- `harness/src/main.rs`: new global `--log-file` (env `NANNA_LOG_FILE`) and `--log-file-level` (env `NANNA_LOG_FILE_LEVEL`, default `info`) flags with "EXPERIMENTAL" docstrings; replaces the bare `tracing_subscriber::fmt().init()` with `logging::init(...)` and holds the `WorkerGuard` for the lifetime of `main`. Emits one `session.start` (UUIDv4 session id) + one `env.snapshot` at the top of `run_agent` and `run_mcp_server`.

### Tests
- 7 unit tests in `logging.rs` (JSON field presence for both events, RFC 3339 timestamp regex, container-id fallback chain, path splitting, `WorkerGuard` shape via the non-blocking builder).
- 2 integration tests (plus 1 `#[ignore]`d model-dependent smoke test) in `harness/tests/logging_integration.rs`. The **regression test** `no_log_file_flag_produces_zero_file_writes` asserts that running `harness tools` in an empty tempdir without `--log-file` creates zero files — the invariant reviewers flagged as must-have.

### Feedback loop results (local)
- `cargo fmt --all -- --check`: clean
- `cargo clippy -p harness --all-targets --all-features -- -D warnings`: clean
- `cargo test -p harness`: 292 lib + 33 integration + 2 logging-integration passing; ignored tests untouched

### Out of scope for PR1 (remaining phases)
- **PR2** (session + container propagation): thread `SessionId` through `AgentContext` / MCP `TaskId`, open a top-level `tracing::info_span!` so every downstream event inherits `session_id` and `container`.
- **PR3** (agent emissions): `state.transition`, `llm.retry`, `model.decision`, `run.outcome`.
- **PR4** (tool dispatch): `tool.call.start` / `tool.call.end`, plus `artifact.written` at `WriteFileTool` / `RunCommandTool`.
- **PR5** (benches + CI): `harness/benches/logging.rs` + workflow_dispatch bench job.
- **PR6** (cleanup): remove duplicate `FmtSubscriber` install in `telemetry.rs`, docs, single-subscriber assertion.

Because this PR only delivers the scaffolding and two events out of the full event schedule, it does not satisfy the full acceptance criteria of #219 on its own — using `Refs #219` rather than `Closes`.

Refs #219

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy -p harness --all-targets --all-features -- -D warnings` passes
- [x] `cargo test -p harness` passes (7 new unit + 2 new integration tests, 0 regressions)
- [x] Regression test verifies `--log-file` absent produces zero file writes
- [ ] Manual smoke: `cargo run -p harness -- --log-file /tmp/nanna.log agent --prompt hi` (requires Ollama; gated behind `#[ignore]`)


harness/src/logging.rs
```diff
@@ -0,0 +1,307 @@
...
+            let (dir, filename) = split_log_path(path);
+            let file_appender = tracing_appender::rolling::never(dir, filename);
+            let (non_blocking, guard) = NonBlockingBuilder::default()
+                .lossy(false)
+                .finish(file_appender);
+
+            let file_filter =
+                EnvFilter::try_new(log_file_level).unwrap_or_else(|_| EnvFilter::new("info"));
+
+            let json_layer = fmt::layer()
```
DominicBurkart, line 88: `tracing_appender::rolling::never` writes to a single file forever with no rotation or size cap. Long-running Nanna sessions (especially `run_mcp_server`, which is a daemon) will grow the file unbounded. Issue #219 does not mandate rotation, but since PR5 is about benches/CI rather than rotation, consider either (a) adding a TODO + issue reference here so a future PR picks up rotation, or (b) documenting on `init` that the caller is responsible for external rotation (e.g., logrotate). At minimum, flag this as a known limitation in the module docstring so operators don't get surprised.

---

harness/src/logging.rs
```diff
@@ -0,0 +1,307 @@
...
+        .with_writer(std::io::stderr)
+        .with_filter(stderr_filter);
+
+    let registry = tracing_subscriber::registry().with(stderr_layer);
+
+    match log_file {
+        Some(path) => {
+            // Ensure parent directory exists so the appender doesn't fail silently.
+            if let Some(parent) = path.parent() {
+                if !parent.as_os_str().is_empty() {
```
DominicBurkart, line 74: Latent double-install hazard: `init()` here calls `registry.init()` (which sets the global default subscriber) and `telemetry::TelemetrySystem::initialize()` also calls `tracing::subscriber::set_global_default`. If any future code path initializes both (e.g., a test or a binary that wires up `ObservabilitySystem`), the second one will silently error or panic depending on the path. The PR body notes PR6 will remove the duplicate, which is fine, but a defensive `debug_assert!` or a comment here pointing at `telemetry.rs` would help the next reader. Also: `init()` is not idempotent — calling it twice panics via `init()` on the registry. Worth documenting.

---

harness/src/logging.rs
```diff
@@ -0,0 +1,307 @@
...
+
+/// Emit a `session.start` event capturing the per-run correlation id and the
+/// user prompt. Callers should invoke this once at the top of `run_agent` or
+/// `run_mcp_server`. Later PRs will add span propagation so every downstream
+/// event inherits the `session_id` automatically.
+pub fn session_start(session_id: &str, user_prompt: &str) {
+    info!(
+        event = "session.start",
+        session_id = %session_id,
+        user_prompt = %user_prompt,
```
DominicBurkart, line 128: `user_prompt` is logged verbatim as `%user_prompt`. Issue #219 explicitly states "we're not going to build a backend for these, so there are no immediate off-machine secret/PII concerns," which legitimizes this for PR1. However, user prompts routinely contain API keys, tokens, and file contents pasted by users — and this log file can be copied off-machine (support bundle, CI artifact, etc.). Recommend adding a non-blocking comment noting this known risk and tracking a `redact_secrets` pass (regex for `sk-…`, `ghp_…`, AWS keys, etc.) as a follow-up. Not a blocker per #219's stated scope.

---

harness/src/main.rs
```diff
@@ -483,6 +515,20 @@ async fn run_mcp_server(
...
+    // `user_prompt` field is deliberately a placeholder at server boot — per
+    // request prompts will be logged in later PRs at the task dispatch site.
+    let session_id = uuid::Uuid::new_v4().to_string();
+    logging::session_start(&session_id, "<mcp-server>");
+
+    let workspace_root = std::env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from("."));
+    let git_repo = GitRepository::detect(&workspace_root);
+    logging::env_snapshot(
+        &workspace_root,
+        git_repo.as_ref().and_then(|g| g.current_branch.as_deref()),
```
DominicBurkart, line 619: For `run_mcp_server`, emitting `session.start` with `user_prompt = "<mcp-server>"` at server boot conflates "session" (a per-request concept) with "server lifetime." Later PRs are supposed to thread `SessionId` through `AgentContext`/`TaskId` — consider renaming this boot event to `server.start` (keeping `session.start` reserved for actual per-prompt sessions) so you don't have to change the semantic of `session.start` in PR2. Minor / style.

---

harness/src/logging.rs
```diff
@@ -0,0 +1,307 @@
...
+/// that writes to the named file via `tracing_appender::non_blocking`. The
+/// non-blocking writer is configured with `lossy(false)` so backpressure blocks
+/// the producer rather than dropping records silently — correctness matters
+/// more than throughput for observability data.
+///
+/// Returns `Some(WorkerGuard)` when a log file is configured; the caller MUST
+/// hold the guard until the process is ready to exit so buffered records flush.
+/// Returns `None` when no log file is configured.
+pub fn init(log_file: Option<&Path>, log_file_level: &str) -> std::io::Result<Option<WorkerGuard>> {
+    // Stderr human-readable layer — always installed.
```
DominicBurkart, line 61: `create_dir_all` runs before `tracing_appender::rolling::never`, but the appender itself also handles directory creation. Fine for now, but the `io::Result` you surface here conflates "parent dir couldn't be created" with "file couldn't be opened". Consider wrapping both failure classes in a dedicated error type so the caller's `?` in `main` gives a better UX than a bare `io::Error`.

---

harness/tests/logging_integration.rs
```diff
@@ -0,0 +1,126 @@
...
+    let mut cmd = Command::cargo_bin("harness").expect("locate harness binary");
+    cmd.current_dir(workdir.path())
+        .env_remove("NANNA_LOG_FILE")
+        .env_remove("NANNA_LOG_FILE_LEVEL")
+        .arg("--log-file")
+        .arg(&log_path)
+        .arg("tools");
+
+    let output = cmd.output().expect("spawn harness");
+    assert!(
```
DominicBurkart, line 70: `log_file_flag_produces_jsonl_output_if_any` accepts an empty/missing log file as a pass. That's a weak assertion — it only validates format *if* data happens to appear. Since `session.start` and `env.snapshot` only fire from `run_agent`/`run_mcp_server` (gated on Ollama), the `tools` subcommand path will reliably produce zero lines, making this test effectively a duplicate of the previous "zero writes" invariant. Consider adding a lightweight code path that always fires an event (e.g., a hidden `--log-smoke` subcommand, or reusing existing `tracing::info!` calls triggered by `tools`) so the JSON schema is actually exercised in CI without Ollama.

---

harness/src/logging.rs
```diff
@@ -0,0 +1,307 @@
...
+    layer::SubscriberExt,
+    util::SubscriberInitExt,
+    EnvFilter, Layer,
+};
+
+/// Resolve the container identifier for this process.
+///
+/// Prefers `$HOSTNAME` (which podman/docker set inside containers), falls back
+/// to `$CONTAINER_NAME`, and finally to `"host"` when neither is present.
+pub fn resolve_container_id() -> String {
```
DominicBurkart, line 35: `resolve_container_id()` falls back to `"host"` when neither `$HOSTNAME` nor `$CONTAINER_NAME` is set. On Linux bash, `$HOSTNAME` is a bash-only shell variable and is *not* inherited by child processes unless exported — so when the harness is spawned by a non-bash parent (systemd, cron, `exec(3)` from another binary, or a minimal podman run without `-e HOSTNAME`), this will silently fall through to `"host"` even inside a container. Consider reading `/etc/hostname` or calling `gethostname(2)` (via the `hostname` or `gethostname` crate, or `nix::unistd::gethostname`) before the env fallbacks. Issue #219 explicitly requires "every log should be tagged with the container that it's running in" — the current implementation is best-effort.

---

harness/src/logging.rs
```diff
@@ -0,0 +1,307 @@
...
+    #[test]
+    fn session_start_emits_required_json_fields() {
+        let buf = BufferWriter::default();
+        let layer = fmt::layer()
+            .json()
+            .flatten_event(true)
+            .with_current_span(false)
+            .with_span_list(false)
+            .with_timer(UtcTime::rfc_3339())
+            .with_writer(buf.clone());
```
DominicBurkart, line 238: The unit tests validate JSON field presence via substring matching (`out.contains("\"service\":\"nanna-coder\"")`). This is brittle — any change to field ordering or serialization (e.g., tracing-subscriber updating its JSON layer) will quietly break assertions in surprising ways. Prefer parsing each line with `serde_json::from_str::<serde_json::Value>` and asserting on `.get("service")`, which is robust to field reordering and whitespace.

---

harness/Cargo.toml
```diff
@@ -34,4 +35,5 @@ reqwest = { version = "0.11", features = ["json", "blocking"] }
 [dev-dependencies]
 futures = "0.3"
 tempfile = "3.0"
-serial_test = "3.0"
\ No newline at end of file
+serial_test = "3.0"
+assert_cmd = "2.0"
```
DominicBurkart: Nit: file still lacks a trailing newline (the `\ No newline at end of file` marker in the diff). Worth fixing while you're here — some CI lints flag this.

---

DominicBurkart (commented):
PR1 scope looks tight and well-aligned with #219 — scaffolding + two events + a regression test for the "no flag, no file" invariant is the right shape for a first slice. Dependencies are appropriate (`tracing-appender`, `assert_cmd`, `json`/`time` features) and the stderr layer is preserved so existing UX is untouched. The `WorkerGuard` lifetime handling in `main` is correct.

Main concerns (none are blockers given the explicit "PR1 of 6" scope and #219's stated stance on PII):

1. `resolve_container_id()` relies on `$HOSTNAME`, which is bash-shell-only and not inherited by child processes in many real container launches — falls back to `"host"` silently. Issue #219 explicitly requires every log to be tagged with container, so `gethostname(2)` would be more robust.
2. Log file has no rotation (`rolling::never`) — long-lived MCP server will grow unbounded. Worth a docstring note or a follow-up issue.
3. The `log_file_flag_produces_jsonl_output_if_any` integration test is an `if any` assertion — in practice it produces zero lines on the `tools` path, so it's not actually exercising the JSON schema in CI. Consider a lightweight always-emit path.
4. Unit tests rely on substring matching against JSON output; `serde_json::from_str` would be less brittle.
5. Latent double-subscriber install vs. `telemetry::initialize()` is acknowledged as PR6 work but a comment pointing at the conflict here would help future readers.
6. Prompt logging is legitimized by #219's "no PII concerns" language, but a follow-up redaction pass for secret patterns (`sk-…`, `ghp_…`, AWS keys) would be prudent since log files are easy to ship off-machine.

No bloat in the diff; each file change maps cleanly to the stated scope. Nice regression test design.

---

DominicBurkart:
**Iteration under job `DqQQS` timed out mid-sequence.** One prep commit landed (`9635f15` — `libc` dep + `Cargo.toml` trailing newline). 7 of 9 review threads remain unaddressed. Punch list for the next cycle:

- [ ] **`logging.rs:35`** — replace `$HOSTNAME`/`$CONTAINER_NAME` fallback with `libc::gethostname()` (dep is now in place) or `/etc/hostname`; env only as last resort.
- [ ] **`logging.rs:61`** — introduce `LoggingInitError` enum distinguishing dir-creation vs file-open failures; return from `init()`.
- [ ] **`logging.rs:74`** — add docstring noting double-init hazard w.r.t. `telemetry::TelemetrySystem::initialize` + `debug_assert!` on single subscriber.
- [ ] **`logging.rs:88`** — module docstring flagging unbounded-file-growth; TODO referencing the follow-up issue (still to be filed).
- [ ] **`logging.rs:128`** — known-risk note on `user_prompt` leaking secrets; file a follow-up issue `[obs] redact secrets from log output` and link it here.
- [ ] **`logging.rs:238`** — replace substring-matching test assertions with `serde_json::from_str::<Value>` parsing + `.get(...)`.
- [ ] **`main.rs:528`** — rename the boot event from `session.start` to `server.start` (reserve `session.start` for per-prompt sessions per the PR6 plan).
- [ ] **`logging_integration.rs:70`** — add a lightweight always-fires code path (e.g., hidden `--log-smoke` subcommand or `tracing::info!` in `tools` entry) so the JSON-schema assertion is actually exercised without Ollama.

Each item has a concrete owner quote in the review threads; no new research needed.

---

Issue #129, opened 2026-03-30
first SWE-bench using latest laptop-appropriate model


DominicBurkart:
SWE-BENCH comparison ran with Claude.ai VS Claude.ai + nanna and results (including Claude.ai token usage) recorded.

I want the classic, human curated, and multilingual benches. We'll generate small markdown+mermaid reports in an organized directory structure (/<commit sha>/<bench name>/<scenario>). Scenarios might be claude_code__x86_laptop where left of the __ is the agent and right is where the model nanna uses is hosted.

Issue #89, opened 2026-03-22
eval: runner — execute nanna against single cases


DominicBurkart:
Part of #84.

Build the integration point that executes nanna's agent loop against fixture repositories and captures metrics.

## Scope

- `async fn run_eval()` that:
  - Copies fixture repos to temporary directories
  - Configures and invokes `AgentLoop` with task prompts
  - Runs verification commands post-completion
  - Captures: pass/fail status, duration, iterations, token counts
- `EvalRunnerConfig` struct with model and verbosity settings
- Timeout enforcement via `tokio::time::timeout`
- Integration test marked `#[ignore]` (requires Ollama instance)

## Acceptance criteria

- [ ] `run_eval()` returns structured metrics on completion
- [ ] Temporary directory isolation verified (no cross-test contamination)
- [ ] Timeout enforced correctly
- [ ] `cargo clippy` and `cargo fmt` pass

## Dependencies

Blocked by #87 (report generation) and #88 (eval case format).

PR #236, opened 2026-04-18
fix(eval): wire NANNA_EVAL_MODEL env var through to E2E tests

### LINKED ISSUE #235: ci: ignored model-integration tests never run — Ollama probe on :11434 always fails
## Summary

The `integration-container` matrix leg in `.github/workflows/ci.yml` never actually executes the `--run-ignored ignored-only` test step, so model-backed integration tests (including the `gemma4:e4b` swap from #202 and anything depending on a live model) are silently skipped on every CI run.

## Details

In `.github/workflows/ci.yml:145-154`:

```yaml
- name: Run integration tests (container)
  if: matrix.test-type == 'integration-container'
  shell: bash
  run: |
    if curl -s --max-time 5 http://localhost:11434/api/tags > /dev/null 2>&1; then
      nix develop --command cargo nextest run --workspace --test '*' --all-features --run-ignored ignored-only --test-threads=1
    else
      echo "Ollama not reachable on port 11434, skipping model integration tests"
    fi
```

The guard on line 150 probes `http://localhost:11434/api/tags`, but nothing in the job ever starts an Ollama server on that address. The preceding steps (ci.yml:98-120) only do `nix build .#ollamaImage` and `nix run .#ollamaImage.copyToDockerDaemon` — they load the image into Docker, they don't run it. There is no `ollama serve` or `docker run -p 11434:11434 nanna-coder-ollama:latest` anywhere in this job.

Result: the `curl` fails, the `else` branch prints "Ollama not reachable on port 11434, skipping model integration tests", and the ignored tests — which include every `#[ignore]`-gated E2E test using `E2E_MODEL = "gemma4:e4b"` — never run.

## Why the probe is wrong

The ignored tests don't need a pre-started Ollama on the host. They call `harness::container::start_container_with_fallback` themselves (see `harness/tests/integration_tests.rs`, `dev_container_integration.rs`, `nanna_self_dev_integration.rs`) and manage the container lifecycle inside the test. The probe appears to have been intended as a "skip if no runtime available" guard but is checking the wrong thing — it checks for an *already-running* Ollama server rather than for a container runtime.

## Impact

- PR #202 (merged 2026-04-12, "switch CI E2E model to gemma4:e4b") was merged with its own acceptance boxes unchecked and could not have been validated by CI even if it had waited for results.
- Any future model-swap PR gets the same silent pass.
- Blocker for landing the nanna + gemma4:e4b eval (#129).

## Proposed fix

Pick one:

1. **Remove the probe** — run the ignored tests unconditionally on the `integration-container` leg, since the tests manage their own container. If a runtime isn't available the tests will fail loudly, which is the correct signal.
2. **Fix the probe** — change it to check for a container runtime (`command -v podman || command -v docker`) rather than a live Ollama endpoint.
3. **Actually start Ollama** — add a `docker run -d -p 11434:11434 nanna-coder-ollama:latest` step before the probe. Heavier, and redundant with what the tests already do internally.

(1) is the smallest change and matches test design. (2) is safer if we want to keep a skip path for matrix legs that genuinely can't run containers.

## Acceptance criteria

- [ ] `--run-ignored ignored-only` step runs (or skips for a *correct* reason) on every `integration-container` CI run
- [ ] A deliberately-failing ignored test causes CI to fail (validated by pushing a temporary red test and observing CI)
- [ ] PR #202's test plan boxes can actually be checked by CI output

## Related

- #129 (SWE-bench with gemma4:e4b — blocked)
- #202 (merged without CI validating the model swap)
- #84 (agentic eval suite)

DominicBurkart:
The eval workflow exposes a `model` input but the harness E2E tests
hardcoded `E2E_MODEL = "gemma4:e4b"` and ignored the workflow's env.
Replace the const with an `e2e_model()` helper that reads
`NANNA_EVAL_MODEL` (default: gemma4:e4b) in each of the three E2E
test files so the workflow's choice can actually affect test execution.

Companion workflow changes to eval.yml (add gemma4:e4b choice, export
NANNA_EVAL_MODEL, harden ollama pull with retry/timeout) are staged
locally but cannot be pushed from this session — the GitHub App lacks
`workflows` permission. They need to be applied manually; see the
session transcript for the exact diff.

Related to #129 (gemma4 SWE-bench eval), does not fix #235 (ignored
integration tests never actually run on CI).

### DIFF
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -10,9 +10,10 @@ on:
       model:
         description: 'Model to use for evaluation'
         required: false
-        default: 'qwen3:0.6b'
+        default: 'gemma4:e4b'
         type: choice
         options:
+          - gemma4:e4b
           - qwen3:0.6b
           - llama3.1:8b
       case_filter:
@@ -51,17 +52,33 @@ jobs:
       run: |
         curl -fsSL https://ollama.com/install.sh | sh
         ollama serve &
-        # Wait for Ollama to be ready
-        for i in $(seq 1 30); do
+        # Wait up to 60s for Ollama to be ready
+        for i in $(seq 1 60); do
           ollama list > /dev/null 2>&1 && break
           sleep 1
         done
-        ollama pull "$MODEL"
+        if ! ollama list > /dev/null 2>&1; then
+          echo "Ollama did not become ready within 60s"
+          exit 1
+        fi
+        for attempt in 1 2 3; do
+          echo "Pulling $MODEL (attempt $attempt/3)..."
+          if timeout 600 ollama pull "$MODEL"; then
+            echo "Pulled $MODEL"
+            break
+          fi
+          if [ "$attempt" = "3" ]; then
+            echo "Failed to pull $MODEL after 3 attempts"
+            exit 1
+          fi
+          sleep $((attempt * 10))
+        done
 
     - name: Run eval suite
       id: eval
       env:
         MODEL: ${{ inputs.model }}
+        NANNA_EVAL_MODEL: ${{ inputs.model }}
         CASE_FILTER: ${{ inputs.case_filter }}
       run: |
         echo "## Eval Results" > eval-results.md
--- a/codecov.yml
+++ b/codecov.yml
@@ -11,3 +11,11 @@ coverage:
 
 ignore:
   - "harness/src/main.rs"
+  # These test files contain only #[ignore]-gated E2E tests that run
+  # under --run-ignored ignored-only on the integration-container job
+  # (see .github/workflows/ci.yml). Default CI unit test runs never
+  # execute them, so any new code in these files is genuinely
+  # uncoverable by the codecov patch report. integration_tests.rs is
+  # NOT listed here because it also contains many non-ignored tests.
+  - "harness/tests/dev_container_integration.rs"
+  - "harness/tests/nanna_self_dev_integration.rs"
--- a/codecov.yml
+++ b/codecov.yml
@@ -20,3 +20,8 @@ ignore:
   - "harness/tests/dev_container_integration.rs"
   - "harness/tests/nanna_self_dev_integration.rs"
   - "harness/tests/eval_case_runner_integration.rs"
+  # run_eval_case + run_cargo_in_container require a live container
+  # runtime (podman/docker) and cannot execute in the unit CI job.
+  # The pure helpers they compose live in runner.rs and are covered
+  # by unit tests in that file.
+  - "harness/src/eval/runner_orchestrator.rs"
--- a/codecov.yml
+++ b/codecov.yml
@@ -19,3 +19,4 @@ ignore:
   # NOT listed here because it also contains many non-ignored tests.
   - "harness/tests/dev_container_integration.rs"
   - "harness/tests/nanna_self_dev_integration.rs"
+  - "harness/tests/eval_case_runner_integration.rs"
--- a/harness/src/eval/mod.rs
+++ b/harness/src/eval/mod.rs
@@ -4,8 +4,12 @@
 //! by the agent evaluation framework in [`crate::agent::eval`].
 
 pub mod report;
+pub mod runner;
 
 // Re-export commonly used types from the agent eval module
 pub use crate::agent::eval::{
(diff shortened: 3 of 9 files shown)


DominicBurkart:
## Workflow diff — please apply manually

`harness::eval::run_eval_case` and a live-model integration test landed in `ac6c17d`, and the workflow needs a second step to actually run the ignored integration test (the existing `test(eval)` filter only hits unit tests — see issue #89). The diff below is ready but the session can't push `.github/workflows/*` (GitHub App lacks `workflows` permission, same constraint as b575a40).

```diff
diff --git a/.github/workflows/eval.yml b/.github/workflows/eval.yml
@@ -101,20 +101,57 @@ jobs:
           --workspace --all-features \
           -E "$FILTER" \
           --no-fail-fast 2>&1 | tee eval-output.txt
-        EXIT_CODE=$?
+        UNIT_EXIT_CODE=$?
         set -e

         cat eval-output.txt >> eval-results.md
         echo '```' >> eval-results.md

-        if [ "$EXIT_CODE" -eq 0 ]; then
+        if [ "$UNIT_EXIT_CODE" -eq 0 ]; then
+          echo "" >> eval-results.md
+          echo "**Unit eval status: PASSED**" >> eval-results.md
+        else
+          echo "" >> eval-results.md
+          echo "**Unit eval status: FAILED**" >> eval-results.md
+        fi
+
+        echo "" >> eval-results.md
+        echo "### Live eval cases (EvalCaseRunner \`${MODEL}\`)" >> eval-results.md
+        echo '```' >> eval-results.md
+
+        # Runs harness::eval::run_eval_case against the checked-in fixtures
+        # under evals/cases/**. NANNA_EVAL_MODEL selects the model inside
+        # the integration test (see harness/tests/eval_case_runner_integration.rs).
+        set +e
+        nix develop --command cargo nextest run \
+          --workspace --all-features \
+          --run-ignored only \
+          -E 'test(/run_.*_against_live_model/)' \
+          --test-threads=1 \
+          --no-fail-fast 2>&1 | tee live-eval-output.txt
+        LIVE_EXIT_CODE=$?
+        set -e
+
+        cat live-eval-output.txt >> eval-results.md
+        echo '```' >> eval-results.md
+
+        if [ "$LIVE_EXIT_CODE" -eq 0 ]; then
+          echo "" >> eval-results.md
+          echo "**Live eval status: PASSED**" >> eval-results.md
+        else
+          echo "" >> eval-results.md
+          echo "**Live eval status: FAILED**" >> eval-results.md
+        fi
+
+        # Overall outcome is failure if either layer failed.
+        if [ "$UNIT_EXIT_CODE" -eq 0 ] && [ "$LIVE_EXIT_CODE" -eq 0 ]; then
           echo "outcome=success" >> "$GITHUB_OUTPUT"
           echo "" >> eval-results.md
-          echo "**Status: PASSED**" >> eval-results.md
+          echo "**Overall: PASSED**" >> eval-results.md
         else
           echo "outcome=failure" >> "$GITHUB_OUTPUT"
           echo "" >> eval-results.md
-          echo "**Status: FAILED** (see logs above)" >> eval-results.md
+          echo "**Overall: FAILED** (see logs above)" >> eval-results.md
         fi

@@ -137,4 +174,5 @@ jobs:
         path: |
           eval-results.md
           eval-output.txt
+          live-eval-output.txt
         if-no-files-found: warn
```

Commit message I would have used:

```
ci(eval): run EvalCaseRunner live cases alongside unit tests

Extends eval.yml so the workflow exercises both layers:

1. `test(eval)` (unchanged): unit tests for the case format, the new
   runner helpers (classify_outcome, find_missing_symbols,
   detect_changed_files, EvalRunConfig builders), and the existing
   AgentEvaluator in-memory scenarios.
2. New `--run-ignored only` step filtered to `run_*_against_live_model`:
   runs harness::eval::run_eval_case against checked-in fixtures under
   evals/cases/**/repo with the chosen model (via NANNA_EVAL_MODEL).
```

Without this diff, a manual `workflow_dispatch` run only exercises unit tests — the `gemma4:e4b` selection in the dropdown still propagates (via `NANNA_EVAL_MODEL` from b575a40), but no test actually hits the model, so #129 stays unvalidated.

---

PR #244, opened 2026-04-20
Align MCP task status vocabulary with ARCHITECTURE.md; trim main.rs wrapper

DominicBurkart:
Automated improvement PR — please review before merging.

Two small, surgical changes that align the code with the documented architecture and remove redundancy, without changing any intended behavior.

## Action items

- **Surface `Cancelled` as a distinct MCP status (architectural alignment).** `ARCHITECTURE.md` documents `poll_task` as returning one of `running`, `completed`, `failed`, or `cancelled`, and describes `cancel_task` as transitioning a task to `cancelled` "so subsequent `get_result` calls return a consistent terminal record." The code previously only emitted `"Cancelled"` from `handle_cancel_task`'s own response; once a task was cancelled, `handle_poll_task`, `handle_get_result`, and `handle_list_tasks` all reported it as plain `"Failed"` — because internally cancellation is stored as `TaskStatus::Failed { diagnostics.error_type: "Cancelled" }`. That contract gap is now closed by translating the internal sentinel to `"Cancelled"` at the MCP boundary.

- **Consolidate status mapping into a single helper (redundancy reduction).** All three read handlers in `harness/src/mcp/handlers.rs` now share a tiny `external_status_str` function instead of duplicating a match on `TaskStatus`. This removes the duplicated match arm in `handle_list_tasks` and trims the status/iterations match in `handle_poll_task`, so the external status vocabulary lives in exactly one place. A regression test (`test_poll_cancelled_task_reports_cancelled_status`) exercises `cancel -> poll / get_result / list_tasks`.

- **Drop the redundant `create_tool_registry` wrapper in `main.rs` (complexity reduction).** `harness/src/main.rs` defined a local `create_tool_registry` whose body was a single call to `harness::tools::create_tool_registry`. It added no behavior, just an extra function and a leftover indirection. The public helper is now imported and used directly, matching how `run_agent` was already calling it through its qualified path.

## Why these reduce complexity / improve alignment

- The first item is a real architectural bug: external orchestrators following the documented MCP contract would never observe `cancelled` as a terminal state today. Fixing it keeps `ARCHITECTURE.md` honest.
- The second item deletes duplicated pattern-matching and lowers the chance of a future divergence between how each handler stringifies status.
- The third item removes a zero-value alias, shrinking `main.rs`'s local API surface.

## Test plan

- [ ] `cargo test -p harness mcp::handlers` (new test + existing handler tests pass)
- [ ] `cargo build -p harness` (main.rs still compiles after wrapper removal)
- [ ] `cargo clippy --all-targets -- -D warnings`


DominicBurkart (commented):
## Review — Align MCP task status vocabulary with ARCHITECTURE.md; trim main.rs wrapper

Reviewed: `harness/src/mcp/handlers.rs`, `harness/src/main.rs`, `ARCHITECTURE.md`.
Head commit: `6734fe9`.

---

### Issues that should be fixed before merge

#### 1. `handle_get_result` duplicates the cancellation check instead of calling `external_status_str`

`external_status_str` was introduced specifically to centralise the `"Cancelled"` sentinel logic, but `handle_get_result` bypasses it with an inline `if`:

```rust
// handle_get_result — current code
let status_str = if diagnostics.error_type == "Cancelled" {
    "Cancelled"
} else {
    "Failed"
};
```

This is the exact divergence risk the helper was meant to prevent. If the sentinel string ever changes (e.g. a typo fix, or the internal representation is refactored), `handle_get_result` will silently fall out of sync again — which is precisely the bug this PR is fixing. The fix is to call `external_status_str` here too. One clean way to do it is to compute the status string before the consuming match:

```rust
let status_str = external_status_str(&task.status);
match task.status {
    TaskStatus::Completed { result, finished_at } => { /* use "Completed" directly or status_str */ }
    TaskStatus::Failed { error, diagnostics, finished_at } => {
        Ok(serde_json::json!({ "status": status_str, ... }))
    }
    TaskStatus::Pending => Err(...),
    TaskStatus::Running { .. } => Err(...),
}
```

---

#### 2. Test only exercises `Pending → Cancelled`, not `Running → Cancelled`

`TaskManager::new(0)` prevents the task from ever being scheduled, so the test cancels a task that never left `Pending`. That is a different code path from cancelling an agent loop that is actively running. The documented motivating scenario is:

> `cancel_task` — request termination of a **running** task … Nanna stops its agent loop at the **next safe checkpoint**.

And the PR description itself says the bug was that "once a task was cancelled, `handle_poll_task`, `handle_get_result`, and `handle_list_tasks` all reported it as plain `Failed`" — implying the task had been running. A test that starts the agent loop (even with a `MockProvider` that errors immediately to drive the task to `Failed`, followed by a cancellation of a parallel task that is mid-run) would close this gap. At minimum, the test comment should note that it only covers the Pending-cancel path, not the Running-cancel path.

---

### Non-blocking observations

#### 3. Codecov reports one uncovered line (94.4% patch coverage)

Per the Codecov bot comment, one line in `handlers.rs` is missing coverage. This is likely the `Err("Task {} is still running")` arm in `handle_get_result`, which is unreachable in tests because no test starts a task's agent loop and then immediately calls `get_result`. Not introduced by this PR, but a companion test for that arm would bring patch coverage to 100%.

#### 4. Status string casing diverges from ARCHITECTURE.md (pre-existing)

`ARCHITECTURE.md` documents the values as lowercase: `running`, `completed`, `failed`, `cancelled`. The code emits PascalCase: `"Running"`, `"Completed"`, `"Failed"`, `"Cancelled"`. This PR correctly uses PascalCase to match the existing code convention, but the divergence from the docs remains. Worth a follow-up issue so external orchestrators can rely on the documented vocabulary exactly.

#### 5. `main.rs` wrapper removal — correct

`create_tool_registry` is now imported from `harness::tools` and used directly in both call sites (`main()` and `run_agent()`). No behavior change, clean deletion.

#### 6. `external_status_str` placement — correct

Module-private (`fn`, not `pub`) at the top of `handlers.rs`, above all handlers. Right place, right visibility.

---

### Summary

The architectural gap is real and the approach is right. Two things to address before merging: (a) `handle_get_result` needs to call `external_status_str` instead of inlining the same check — the inline copy defeats the purpose of the helper; (b) the new test covers `Pending → Cancelled` only, leaving the `Running → Cancelled` path (the scenario described in the PR) untested. Everything else looks good.

https://claude.ai/code/session_01KxYQiFucgGCsPzQ2ALJKLD

---

DominicBurkart:
**CRON status update** — The lint CI failure was fixed this cycle: `handle_get_result` now calls `external_status_str` (eliminating the duplicate inline cancellation check the PR intended to consolidate), and `rustfmt` violations in the new test were corrected; a separate commit also added `test_external_status_str_running` to cover the `Running` arm for Codecov. One **remaining blocker from review**: the new test only covers `Pending → Cancelled`, not `Running → Cancelled` (the motivating scenario described in the PR) — a task that's actively executing should be cancelled and then have its status correctly surfaced by `poll_task`, `get_result`, and `list_tasks`. **Not ready for human review** — add a `Running → Cancelled` integration test. [Session](https://claude.ai/code/session_01KxYQiFucgGCsPzQ2ALJKLD)

---

DominicBurkart (commented):
## Review — head `dd5b765`

Reviewed: `harness/src/mcp/handlers.rs`, `harness/src/main.rs`, `ARCHITECTURE.md`.

Previous blockers from the `6734fe9` review are resolved: `handle_get_result` now calls `external_status_str` before the consuming match (clean), and the `Running` arm of the helper is covered by the new synchronous unit test. One blocker remains.

### Must fix before merge

**`Running → Cancelled` path is untested.** The test `test_poll_cancelled_task_reports_cancelled_status` uses `TaskManager::new(0)`, which prevents scheduling, so it only exercises `Pending → Cancelled`. The motivating bug — as stated in the PR description — is that tasks which *ran* and were cancelled were misreported as `Failed`. That path goes through the `TaskStatus::Failed { diagnostics.error_type: "Cancelled" }` sentinel, which is the arm `external_status_str` was specifically written to translate. A test that cancels a task that has actually entered `Running` (and whose internal state is therefore stored as `Failed { error_type: "Cancelled" }`) would confirm the sentinel translation works end-to-end. The existing test and the new `test_external_status_str_running` do not cover this path together.

### Non-blocking

- **Casing divergence vs. ARCHITECTURE.md (pre-existing).** The docs specify `running`, `completed`, `failed`, `cancelled` (lowercase); the code emits PascalCase. This PR correctly matches the existing code convention, but a follow-up issue to align the docs (or the wire format) would prevent confusion for external orchestrators.
- **`main.rs` wrapper removal** — correct and clean.
- **`external_status_str` placement and visibility** — correct.

https://claude.ai/code/session_01RTLA2PX1wUvM9oj8wrAteb

---

DominicBurkart:
**CRON status update** — Both blockers from the previous review are now addressed:

1. **`handle_get_result` calls `external_status_str`** — was already fixed in the prior cycle (`dd5b765`).

2. **Running→Cancelled integration test** — Added `test_running_to_cancelled_reports_cancelled_status` in `mcp::handlers` (commit `db92361`). It injects a task directly into `TaskManager`'s Running state via two new `#[cfg(test)]` helpers (`insert_running_task_for_test`, `insert_dummy_handle_for_test`), cancels it via `handle_cancel_task`, then asserts all three read handlers (`poll_task`, `get_result`, `list_tasks`) surface `Cancelled`. The existing `test_poll_cancelled_task_reports_cancelled_status` gains a comment noting it covers the Pending-cancel path only. All 13 handler tests pass; `cargo clippy --all-targets -- -D warnings` is clean.

[Session](https://claude.ai/code/session_01RTLA2PX1wUvM9oj8wrAteb)

---

DominicBurkart:
**Automated cron status.** Review posted — the two prior-cycle blockers (delegation to `external_status_str`, Running-arm unit coverage) were confirmed fixed on the current head; the one remaining blocker was that `test_poll_cancelled_task_reports_cancelled_status` only exercised `Pending → Cancelled`. Addressed that blocker in commit `db92361`: added `test_running_to_cancelled_reports_cancelled_status` plus two `#[cfg(test)]` helpers (`insert_running_task_for_test`, `insert_dummy_handle_for_test`) so handler tests can inject a `Running` task without spinning up a real agent loop; all 13 handler tests + clippy `-D warnings` pass locally. Branch rebased, CI re-running (prior run was all-green, 92.3 % codecov) — **ready for human review** once the current CI run goes green. [session](https://claude.ai/code/session_01RTLA2PX1wUvM9oj8wrAteb)

---

DominicBurkart (commented):
## Automated review — head `db92361`

Reviewed: `harness/src/mcp/handlers.rs`, `harness/src/main.rs`, `harness/src/task.rs`, `ARCHITECTURE.md`.

All prior blockers are resolved. The new `test_running_to_cancelled_reports_cancelled_status` directly seeds a `Running` task plus a live abort handle via the new `#[cfg(test)]` helpers on `TaskManager`, then drives `cancel → poll / get_result / list_tasks` and asserts `"Cancelled"` on all three. That exercises the `Failed { error_type: "Cancelled" }` sentinel translation end-to-end from the motivating `Running → Cancelled` path — the exact gap flagged on `dd5b765`.

### Correct
- `external_status_str` is the single source of truth used by all three read handlers.
- `main.rs` wrapper removal is clean; `create_tool_registry` imported and used directly.
- `#[cfg(test)]` helpers on `TaskManager` are properly gated and scoped.

### Non-blocking nits
- `insert_dummy_handle_for_test` comment says the dummy task is "intentionally leaked", but the `JoinHandle` is dropped at scope end; since the task is already aborted via `dummy.abort()`, nothing actually leaks. Comment could be tightened, but behaviour is fine.
- Casing divergence vs ARCHITECTURE.md (`Cancelled` vs documented `cancelled`) still pre-existing — worth a follow-up issue to align docs or wire format.

No bugs, no bloat, no secrets/PII. LGTM.

---
_Automated review by Claude Code_

---

DominicBurkart:
**Automated cron status** ([session](https://claude.ai/code/session_01Nokmxow6PWSA7TRChp5o8u))

Review subagent posted a COMMENT confirming prior blockers resolved at `db92361` (`handle_get_result` now calls `external_status_str`; new `test_running_to_cancelled_reports_cancelled_status` seeds a `Running` task + live abort handle via `#[cfg(test)]` helpers and asserts `"Cancelled"` across `poll_task`/`get_result`/`list_tasks`), main.rs wrapper trim clean, minor nits only (misleading "intentionally leaked" comment; pre-existing PascalCase vs ARCHITECTURE.md lowercase divergence is out of scope). CI-fix subagent diagnosed `lint (ubuntu)` fmt failure on the new test and pushed `bd65ca5` with `cargo fmt --all` — **NEEDS CI RECHECK**; if lint goes green this is READY.

---

PR #253, opened 2026-04-21
feat(harness/logging): redact secrets from log output (#232)

### LINKED ISSUE #232: [obs] redact secrets from log output (follow-up to PR #229)
Follow-up to review thread on PR #229 (JSON structured logging).

Context: `harness::logging::session_start` currently emits `user_prompt` verbatim. Issue #219 scopes log files as local-only with no off-machine backend, so this is acceptable for PR1 scaffolding. However, user prompts routinely contain:

- OpenAI-style API keys (`sk-...`)
- GitHub PATs (`ghp_...`, `github_pat_...`)
- AWS keys (`AKIA...`, secret keys)
- Cloud provider tokens (Azure, GCP)
- Pasted file contents (may contain `.env` values)

Log files can still leak off-machine via support bundles, CI artifacts, or casual copy. We should add a best-effort redaction pass before events are emitted.

## Proposed approach

1. Introduce a `redact_secrets(&str) -> Cow<'_, str>` helper in `harness::logging` (or a sibling `redaction` module).
2. Match common patterns via `regex` (already a harness dep):
   - `sk-[A-Za-z0-9]{20,}`
   - `ghp_[A-Za-z0-9]{36}`, `github_pat_[A-Za-z0-9_]{80,}`
   - `AKIA[0-9A-Z]{16}` and contextual AWS secret key pairs
   - Bearer tokens in `Authorization:` headers
   - Generic high-entropy strings gated by prefix hints
3. Replace matches with `"<redacted:<kind>>"` so operators can see that redaction happened without seeing the secret.
4. Apply to `user_prompt` first; extend to future agent-loop / tool-dispatch event payloads in PRs 2-5.
5. Add unit tests for each pattern class plus a property test ensuring redaction is idempotent.

## Non-goals (explicitly)

- Perfect redaction. This is defense-in-depth, not a security boundary. If secrets must be kept out of logs, they should not be in prompts in the first place.
- Encryption of log files at rest (out of scope per #219).

## Acceptance criteria

- [ ] `redact_secrets` helper with unit coverage for each supported pattern class.
- [ ] `session_start` (and any later event that accepts free-form strings) funnels payloads through the helper.
- [ ] Module docstring updated to describe the redaction pass and its limits.
- [ ] CHANGELOG / release notes entry when landed.

See PR #229 review thread: https://github.com/DominicBurkart/nanna-coder/pull/229#discussion_r3103962517

DominicBurkart:
job-id: `3x7Ew`
oldest-issue-id: `232`

Closes #232

## Summary

Adds `harness::logging::redaction::redact_secrets` plus a
`RedactingMakeWriter` `tracing_subscriber::fmt::MakeWriter` adapter so
every log event emitted through the default subscriber passes through a
best-effort secret-redaction pass before hitting stderr (or any other
sink installed by a future caller). Wires the adapter into the default
subscriber installed from `harness/src/main.rs`.

This PR is decoupled from PR #229: the module lands as a self-contained
`harness::logging` submodule so that when #229 merges, its richer
JSON-structured `logging::init(...)` can adopt the same
`RedactingMakeWriter` without reshaping this patch.

## Pipeline

Two orthogonal passes, both idempotent:

**1. Field-name allowlist** (matched case-insensitively, for both JSON
   `"<name>":"<value>"` and logfmt `<name>=<value>` / `<name>: <value>`
   styles):

- `password`
- `api_key`, `apikey`
- `token`
- `secret`
- `authorization`
- `bearer`
- `gh_token`, `github_token`
- `access_key`, `access_token`, `refresh_token`
- `private_key`
- `client_secret`

**2. Regex defense in depth** (matched anywhere in the input):

- `sk-[A-Za-z0-9_\-]{20,}` — OpenAI / Anthropic / generic `sk-` keys
- `ghp_[A-Za-z0-9]{20,}` — GitHub classic PAT
- `github_pat_[A-Za-z0-9_]{20,}` — GitHub fine-grained PAT
- `AKIA[0-9A-Z]{16}` — AWS access-key-id
- `Bearer\s+[A-Za-z0-9._\-]{8,}` — HTTP Authorization bearer tokens
- `ey[A-Za-z0-9_\-]+\.ey[A-Za-z0-9_\-]+\.[A-Za-z0-9_\-]+` — JWT-like
  three-segment base64url
- `\b[0-9a-fA-F]{32,}\b` — long hex strings (hashes, digests,
  raw-key-material shape)

All matches are replaced with `***REDACTED***`.

## Threat model — what this does NOT catch

This is **defense in depth**, not a security boundary. In particular,
the pass does not attempt to redact:

- **Structured secrets that don't match any pattern.** A password of
  `"hunter2"` with no surrounding `password=` key won't be recognized.
- **Partial / fingerprint logs.** If code deliberately logs the first
  six characters of a PAT, those six characters remain.
- **Tokens split across log events.** If a secret is split across two
  `info!()` calls, neither half matches a full-shape regex.
- **Sinks that bypass the configured `tracing_subscriber`.** `println!`,
  `eprintln!`, direct stderr writes from `std::process::Command`, and
  panic messages all go around the `MakeWriter`.
- **Unknown secret shapes.** Non-`sk-`, non-GitHub, non-AWS provider
  formats (Azure AD, GCP, Slack, Twilio, Stripe, etc.) are not currently
  in the regex set.
- **Non-UTF-8 bytes.** The writer forwards non-UTF-8 payloads verbatim
  rather than mangling them; this is a correctness choice, not a
  security one.

See the module docstring on `harness::logging` for the full enumeration.

## Tests (21 new)

- `harness::logging::redaction::tests` (14): per-pattern coverage for
  every regex branch and every field-name style, case-insensitivity,
  short-hex negative case, idempotency.
- `harness::logging::tests` (7): end-to-end via a capturing
  `MakeWriter` wired into a real `tracing_subscriber::fmt()` subscriber
  and driven by `info!()`. Asserts the secret never reaches the sink
  for `sk-*`, `Bearer *`, JWT-like, long hex, `password` field,
  `api_key` field, and a negative-case plain string.

Feedback loop (local, all green):

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test -p harness --lib` (352 tests)
- `cargo test -p harness --tests` (all integration targets)

## Follow-up / incomplete work

- **Provider-specific patterns.** Azure AD, GCP service-account JSON,
  Slack `xox[abpr]-`, Twilio `SK`/`AC`, Stripe `sk_live_`/`pk_live_`,
  Cloudflare API tokens. Safe to extend `SHAPE_PATTERNS`.
- **Structured-field redaction.** Replace the post-serialization byte
  pass with a `tracing::field::Visit` pass that sees field values
  before they are stringified, eliminating the
  whitespace-boundary ambiguity in the `kv_field_regex` path.
- **Depends on PR #229.** Once the structured JSON logging pipeline
  lands, `logging::init(...)` should adopt `RedactingMakeWriter` on the
  non-blocking file writer as well as stderr.
- **Property-test.** Generate random strings containing known secret
  shapes interleaved with plain text and assert no substring of any
  injected secret survives; not yet written.
- **CHANGELOG entry.** Skipped; the repo currently lacks a CHANGELOG
  file.

Not marking `ready-for-review`.


harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,306 @@
...
+///
+/// Each entry is an anchored-enough pattern to be unambiguous. Order is only
+/// significant for cosmetics (longer matches first reduce churn on
+/// overlapping candidates).
+const SHAPE_PATTERNS: &[&str] = &[
+    // OpenAI / Anthropic / generic `sk-` prefixed keys (20+ chars of body)
+    r"sk-[A-Za-z0-9_\-]{20,}",
+    // GitHub personal access tokens
+    r"ghp_[A-Za-z0-9]{20,}",
+    r"github_pat_[A-Za-z0-9_]{20,}",
```
DominicBurkart, line 59: Coverage gaps vs issue #232 and common providers worth adding to `SHAPE_PATTERNS`:

- **GitHub PAT length mismatch**: issue #232 specifies `ghp_[A-Za-z0-9]{36}` and `github_pat_[A-Za-z0-9_]{80,}`; current `{20,}` is looser than requested (arguably fine, but diverges from acceptance criteria).
- **AWS secret access key** (40 char base64-ish) — paired with `AKIA` per the issue.
- **Slack** (`xox[abpr]-...`), **Stripe** (`sk_live_`, `pk_live_`, `rk_live_`), **Twilio** (`AC[0-9a-f]{32}`, `SK[0-9a-f]{32}`), **Google API keys** (`AIza[0-9A-Za-z_\-]{35}`), **GCP service-account JSON** (`-----BEGIN PRIVATE KEY-----`).
- **URLs with embedded credentials**: `https?://[^:\s]+:[^@\s]+@` — neither the field-name nor shape pass catches `https://user:pw@host`.

These are listed in the PR description as follow-ups, which is acceptable, but the `url-with-creds` case feels important enough for v1 given that `reqwest`/curl-style debug logs regularly contain them.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,306 @@
...
+    // AWS access key id
+    r"AKIA[0-9A-Z]{16}",
+    // HTTP Bearer tokens
+    r"Bearer\s+[A-Za-z0-9._\-]{8,}",
+    // JWT-like three-segment base64url string
+    r"ey[A-Za-z0-9_\-]+\.ey[A-Za-z0-9_\-]+\.[A-Za-z0-9_\-]+",
```
DominicBurkart, line 65: The `\b[0-9a-fA-F]{32,}\b` long-hex rule is likely to produce false positives on benign values that show up routinely in log output: git SHAs (40-char hex), container image digests (`sha256:<64-hex>`), trace / span IDs, request IDs, and UUIDs without hyphens. That will redact useful debugging context and could hide causal chains operators rely on.

Consider either:
1. Gating long-hex detection behind a prefix hint (`key=`, `secret=`, `password=`), as issue #232's "Generic high-entropy strings gated by prefix hints" suggests, or
2. Raising the threshold (e.g. 48+ chars) and explicitly excluding `sha256:` / `sha1:` prefixes, or
3. Dropping the pure-shape hex rule and relying on the field-name pass for hex secrets.

At minimum, add a negative test that asserts a git SHA / trace ID is preserved.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,306 @@
...
+/// from the constants in this module.
+pub fn redact_secrets(input: &str) -> Cow<'_, str> {
+    // Short-circuit: if nothing matches any of the shape patterns *and* no
+    // sensitive field name appears (case-insensitive substring check), the
+    // input is clean.
+    let shapes_hit = shape_regex_set()
+        .matches(input)
+        .into_iter()
+        .next()
+        .is_some();
```
DominicBurkart, line 136: Performance concern on the hot logging path: `redact_secrets` unconditionally allocates `lower = input.to_ascii_lowercase()` for the short-circuit check, which defeats the `Cow::Borrowed` fast-path's primary value (zero alloc on clean events). Every single non-secret log line pays an allocation proportional to its length, plus an N-iteration `contains` over `SENSITIVE_FIELD_NAMES`.

Two cheaper alternatives:
1. Build a `RegexSet` combining the field-name alternation (case-insensitive via `(?i)`) and probe with `is_match()` — no allocation, single DFA scan.
2. Skip the lowercase entirely and rely on `shape_regex_set().is_match()` alone for the fast-path, accepting that events containing the bare word "password" (no `=`/`:`) will take the slow path.

Since `RedactingWriter::write` runs on every log event, including high-volume `DEBUG`/`TRACE`, the per-line alloc matters.

---

harness/src/logging/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+/// the UTF-8 prefix / interpretation of the bytes.
+#[derive(Clone, Debug)]
+pub struct RedactingMakeWriter<M> {
+    inner: M,
+}
+
+impl<M> RedactingMakeWriter<M> {
+    /// Wrap an existing [`MakeWriter`] so that every emitted event is run
+    /// through [`redaction::redact_secrets`] before it reaches the inner
+    /// writer.
```
DominicBurkart, line 62: `RedactingWriter::write` assumes each `write` call receives a whole log event. `tracing_subscriber::fmt` happens to call `make_writer()` once per event and `write_all` the serialized bytes, but nothing in the `MakeWriter` contract guarantees a single `write` covers an entire event. If `fmt::Layer` ever splits a line across two `write` calls (or a caller wraps this in something that does), a secret that straddles the boundary will not match any regex — a silent leak.

Options:
1. Buffer until a newline is seen and redact per-line.
2. Implement `WriteAdaptor` over `io::Write` that accumulates into an internal `Vec<u8>` and flushes on `\n` or `flush()`.

Also worth a doc-test / test asserting behavior when `write` is called with a partial event (e.g. two calls, each carrying half of `sk-...`).

---

harness/src/logging/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn new(inner: M) -> Self {
+        Self { inner }
+    }
+}
+
+/// Per-write handle returned by [`RedactingMakeWriter::make_writer`].
+pub struct RedactingWriter<W> {
```
DominicBurkart, line 75: The `Err(_) => self.inner.write(buf)` branch forwards non-UTF-8 bytes verbatim. The docstring calls this a "correctness choice, not a security one" — but it's exploitable as a redaction bypass: a caller emitting a secret with an intentionally (or accidentally) invalid UTF-8 byte prefix would skip redaction entirely. Given `tracing_subscriber::fmt` produces UTF-8 in practice, consider instead using `String::from_utf8_lossy` so the redaction pass still runs on the lossy interpretation, and then writing the lossy-redacted bytes. That's a strictly safer default.

---

harness/src/logging/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn leaves_non_secret_strings_untouched() {
+        let (sink, buf) = VecSink::new();
+        with_subscriber(sink, || {
+            info!(message = "hello world, short string, no secrets");
+        });
+        let out = captured(&buf);
+        assert!(out.contains("hello world"));
+        assert!(!out.contains("***REDACTED***"));
+    }
+}
```
DominicBurkart, line 327: Test coverage vs issue #232 acceptance criteria:

- **Missing positive cases the issue explicitly lists**: AWS secret access key (paired, not just `AKIA` id), Azure/GCP tokens. At least a TODO-skip test would document the gap.
- **Missing negative / false-positive guardrails**: git SHA, UUID, `sha256:` image digest, trace-id, request-id — all likely to collide with the `\b[0-9a-fA-F]{32,}\b` pattern.
- **Missing `session_start` / `user_prompt` integration test**: issue #232 explicitly names `session_start` as the first event to funnel through the helper (acceptance criterion #2). The PR wires the `MakeWriter` globally, which arguably supersedes that requirement, but there is no test asserting a `user_prompt` carrying `sk-...` gets redacted end-to-end.
- **Missing property test**: issue lists "property test ensuring redaction is idempotent." Current `redaction_is_idempotent` is a single example, not a generated one (follow-up section acknowledges this — fine, but flag it against the acceptance checkbox).
- **Missing `env-var-style` keys like `FOO_TOKEN=...` / `BAR_KEY=...` / `BAZ_SECRET=...`**: `kv_field_regex` uses `\b(?:names)` which only matches the literal names in `SENSITIVE_FIELD_NAMES`; `MY_API_TOKEN=xyz` will not match `token` because the preceding `_` is not a word boundary that starts the token. Worth either adding a suffix match (`.*_token`, `.*_key`, `.*_secret`) or extending the list.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,306 @@
...
+            // Split on first `=` or `:` (the only separator per the regex).
+            let sep_idx = whole
+                .find(['=', ':'])
+                .unwrap_or(whole.len());
+            let (key, rest) = whole.split_at(sep_idx);
+            let sep = rest.chars().next().unwrap_or('=');
+            format!("{key}{sep}{REDACTED}")
+        })
+        .into_owned();
+
```
DominicBurkart, line 181: PII coverage is out of scope per the threat model, which is fine, but worth an explicit note in the module docstring that email addresses, phone numbers, IP addresses, SSNs, and credit-card PANs are not redacted. Issue #232 is silent on PII, so leaving it out matches the acceptance criteria, but agents extending this module later will want the scope boundary made explicit so they don't assume coverage.

---

harness/src/main.rs
```diff
@@ -74,8 +74,13 @@ enum Commands {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Install the default tracing subscriber with a redacting writer so that
+    // any logs emitted by the process get a best-effort secret-redaction pass
+    // before hitting stderr. See `harness::logging::redaction` for the scope
+    // and threat model of this pass.
     tracing_subscriber::fmt()
         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_writer(harness::logging::RedactingMakeWriter::new(std::io::stderr))
```
DominicBurkart, line 83: Repo-wide scope note: the default subscriber is only installed from `harness/src/main.rs`. Any other binary target in the workspace that calls `tracing_subscriber::fmt().init()` directly (or any test that installs its own subscriber) bypasses redaction. Worth either:

1. Extracting an `init_default()` helper in `harness::logging` so downstream binaries get redaction for free, or
2. Adding a workspace-level clippy/lint rule forbidding direct `tracing_subscriber::fmt().init()` outside the helper.

Per the PR description, PR #229 is expected to adopt `RedactingMakeWriter` — a shared helper now would avoid having two integration seams.

---

DominicBurkart (commented):
Solid draft PR with thorough docs and 21 tests. Flagging eight concerns for agent follow-up before marking ready-for-review:

1. **Write-boundary bypass**: `RedactingWriter::write` assumes one-call-per-event; no newline buffering, so a split event silently leaks.
2. **Non-UTF-8 bypass**: invalid-UTF-8 path skips redaction entirely; use `from_utf8_lossy`.
3. **Hot-path allocation**: `to_ascii_lowercase()` allocates on every clean log line, defeating the `Cow::Borrowed` fast path.
4. **Long-hex false positives**: `[0-9a-fA-F]{32,}` will redact git SHAs, image digests, trace IDs, UUIDs.
5. **Pattern gaps**: URLs with embedded credentials, AWS secret keys, Slack/Stripe/Twilio/GCP/Google — issue #232 names several.
6. **Env-var field shape**: `MY_API_TOKEN=...` does not hit `\b(?:token)\b`; consider suffix match.
7. **Integration gap**: no end-to-end test covering `session_start` / `user_prompt`, which is acceptance criterion #2 on issue #232.
8. **Workspace scope**: only `harness/src/main.rs` installs the subscriber; other binaries/tests bypass redaction. Extract `logging::init_default()`.

Also missing from #232 acceptance criteria: property-based idempotency test (acknowledged as follow-up), and CHANGELOG (acknowledged as N/A).

Threat model doc is excellent — consider also calling out PII non-coverage explicitly so future agents don't assume it.

---

DominicBurkart:
**Status (automated orchestration cron, 3-agent fanout)**

Review posted with 8 inline comments: write-boundary splitting risk, non-UTF-8 redaction bypass, hot-path `to_ascii_lowercase` alloc defeating `Cow::Borrowed`, long-hex false-positives on git SHAs/UUIDs/trace-ids, missing URL-with-creds + Slack/Stripe/Twilio/GCP provider patterns, `MY_API_TOKEN=` env shape not matching `\b(?:token)\b`, missing `session_start`/`user_prompt` integration test, and workspace-scope gap (only `harness/src/main.rs` installs the subscriber). Rustfmt fix pushed as `e76d7e7` + branch rebased on main; `cargo fmt --check`, `cargo clippy -D warnings`, and `cargo doc` all clean locally; CI re-running on all 7 jobs. Not yet ready — review comments should be addressed and CI confirmed green before promoting out of draft.

---

DominicBurkart (commented):
Reviewing against head `1cab51e` (rebase + rustfmt fix on top of the original implementation). The rustfmt pass and rebase are clean; no new substantive changes were introduced in the update. The eight concerns from the prior review remain open — they are re-stated below as inline comments on the current diff so they are not lost against the outdated positions.

Summary of outstanding issues (all security-relevant given this is a secret-redaction feature):

**Must-fix before ready-for-review:**
1. `RedactingWriter::write` processes each `write` call in isolation — a secret split across two calls leaks silently.
2. Non-UTF-8 bytes bypass redaction entirely; `from_utf8_lossy` is safer.
3. `to_ascii_lowercase()` allocates on every log event, defeating the `Cow::Borrowed` fast-path.
4. Long-hex `\b[0-9a-fA-F]{32,}\b` will redact git SHAs, trace IDs, and container digests — active false-positive risk on debugging context.

**Should-fix / gaps vs. acceptance criteria:**
5. `kv_field_regex` uses `\b(?:names)` — `MY_API_TOKEN=xyz` doesn't match because `_TOKEN` has no word boundary before `token`.
6. No `session_start`/`user_prompt` end-to-end test (issue #232 AC #2).
7. URL-with-embedded-credentials (`https://user:pw@host`) not covered; common in `reqwest` debug logs.
8. Workspace scope: other binaries/tests that call `tracing_subscriber::fmt().init()` directly bypass redaction.

**Acknowledged/follow-up (not blocking):**
- Provider gaps (Slack, Stripe, Twilio, GCP, Azure) — listed in PR description.
- Property-test idempotency — acknowledged follow-up.
- PII not covered — worth a doc note so future agents don't assume coverage.

Patch coverage is 88.52% (Codecov) — below the 90% target; the 7 uncovered lines are likely the non-UTF-8 branch and related error paths.

---

harness/src/logging/mod.rs
DominicBurkart, line 62: **[Unresolved from prior review — still present in 1cab51e]** Write-boundary split: `RedactingWriter::write` redacts each `write` call in isolation. The `MakeWriter` contract does not guarantee a single `write` per event; if `fmt::Layer` or any wrapping layer ever splits a log event across two calls, a secret straddling the boundary passes through unredacted — a silent leak.

Recommended fix: buffer until `\n` is seen (or `flush()` is called) and redact per-line:

```rust
pub struct RedactingWriter<W> {
    inner: W,
    buf: Vec<u8>,
}

impl<W: Write> Write for RedactingWriter<W> {
    fn write(&mut self, data: &[u8]) -> io::Result<usize> {
        self.buf.extend_from_slice(data);
        if self.buf.contains(&b'\n') {
            self.flush()?;
        }
        Ok(data.len())
    }

    fn flush(&mut self) -> io::Result<()> {
        if !self.buf.is_empty() {
            let chunk = std::mem::take(&mut self.buf);
            match std::str::from_utf8(&chunk) {
                Ok(s) => self.inner.write_all(redaction::redact_secrets(s).as_bytes())?,
                Err(_) => self.inner.write_all(&chunk)?,
            }
        }
        self.inner.flush()
    }
}
```

Also worth adding a test that calls `write` twice with each half of `sk-ABCDE...` and asserts the full token is still redacted.

---

harness/src/logging/mod.rs
DominicBurkart, line 75: **[Unresolved from prior review — still present in 1cab51e]** Non-UTF-8 bypass: the `Err(_) => self.inner.write(buf)` branch forwards bytes verbatim, skipping redaction entirely. While the docstring frames this as a "correctness choice," it is also an exploitable bypass: any secret prefixed with even one invalid UTF-8 byte avoids the redaction pipeline.

`tracing_subscriber::fmt` always emits UTF-8 in practice, but other writers wrapping this adapter may not. A safer default:

```rust
Err(_) => {
    let lossy = String::from_utf8_lossy(buf);
    let redacted = redaction::redact_secrets(&lossy);
    self.inner.write_all(redacted.as_bytes())?;
    Ok(buf.len())
}
```

This accepts the minor lossy-conversion cost in exchange for ensuring the redaction pass always runs.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,304 @@
...
+    r"sk-[A-Za-z0-9_\-]{20,}",
+    // GitHub personal access tokens
+    r"ghp_[A-Za-z0-9]{20,}",
+    r"github_pat_[A-Za-z0-9_]{20,}",
+    // AWS access key id
+    r"AKIA[0-9A-Z]{16}",
+    // HTTP Bearer tokens
+    r"Bearer\s+[A-Za-z0-9._\-]{8,}",
+    // JWT-like three-segment base64url string
+    r"ey[A-Za-z0-9_\-]+\.ey[A-Za-z0-9_\-]+\.[A-Za-z0-9_\-]+",
```
DominicBurkart, line 65: **[Unresolved from prior review — still present in 1cab51e]** Long-hex false positives: `\b[0-9a-fA-F]{32,}\b` will redact benign values that appear constantly in harness log output:

- Git commit SHAs (40-char hex) — e.g. from `git log`, `subprocess` output, repo references
- Container image digests (`sha256:<64-hex>`)
- Trace/span IDs and request IDs (often 32-char hex)
- UUIDs without hyphens

This will silently strip debugging context operators rely on to correlate events.

Issue #232 specifically says "Generic high-entropy strings **gated by prefix hints**" — the current pattern has no such gate. Consider one of:
1. Gating on a sensitive-field prefix (e.g. only match long hex that follows `key=`, `secret=`, `password=`).
2. Raising the threshold to 48+ chars (excludes git SHAs and most trace IDs) and excluding `sha256:`/`sha1:` prefixed digests.
3. Dropping the pure-shape hex rule entirely and relying on the field-name pass for hex secrets.

At minimum, add a negative test asserting a 40-char git SHA and a 32-char trace ID are preserved.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,304 @@
...
+/// from the constants in this module.
+pub fn redact_secrets(input: &str) -> Cow<'_, str> {
+    // Short-circuit: if nothing matches any of the shape patterns *and* no
+    // sensitive field name appears (case-insensitive substring check), the
+    // input is clean.
+    let shapes_hit = shape_regex_set()
+        .matches(input)
+        .into_iter()
+        .next()
+        .is_some();
```
DominicBurkart, line 136: **[Unresolved from prior review — still present in 1cab51e]** Hot-path allocation: `input.to_ascii_lowercase()` allocates a fresh `String` on **every** log event, including high-volume `DEBUG`/`TRACE` lines that contain no secrets. This defeats the `Cow::Borrowed` fast-path's primary benefit (zero allocation on clean events).

The `any_name` check immediately after doubles the cost by also calling `n.to_ascii_lowercase()` inside the closure, though `SENSITIVE_FIELD_NAMES` are already lowercase so that inner call is a no-op allocation.

A cheaper short-circuit: combine the field-name alternation with `shape_regex_set` into a single `RegexSet` with `(?i)` flag, so the entire guard is one DFA scan with no allocation:

```rust
fn quick_check_regex_set() -> &'static RegexSet {
    static SET: OnceLock<RegexSet> = OnceLock::new();
    SET.get_or_init(|| {
        let field_pat = format!(r"(?i)\b(?:{})\b", field_name_alternation());
        let mut pats: Vec<&str> = SHAPE_PATTERNS.to_vec();
        // field_pat is a String; collect all into owned vec
        let mut all: Vec<String> = SHAPE_PATTERNS.iter().map(|s| s.to_string()).collect();
        all.push(field_pat);
        RegexSet::new(&all).expect("valid quick-check set")
    })
}
```

Then the short-circuit becomes:
```rust
if !quick_check_regex_set().is_match(input) {
    return Cow::Borrowed(input);
}
```

No allocation, single pass.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,304 @@
...
+    })
+}
+
+fn field_name_alternation() -> String {
+    SENSITIVE_FIELD_NAMES
+        .iter()
+        .map(|n| regex::escape(n))
+        .collect::<Vec<_>>()
+        .join("|")
+}
```
DominicBurkart, line 93: **[Unresolved from prior review — still present in 1cab51e]** Env-var field shape gap: `kv_field_regex` uses `\b(?:token|secret|api_key|...)` which requires a word boundary immediately before the field name. This means `MY_API_TOKEN=xyz`, `FOO_SECRET=abc`, `DATABASE_PASSWORD=...` are not matched — the underscore preceding the name is a non-boundary character in Unicode word rules.

In practice, harness environment variables (and many CI secret names) follow the `PREFIX_KEYWORD` pattern. Consider either:
1. Adding a suffix-match alternative: `(?:[A-Z0-9_]*_)?(?:token|secret|key|password)` (case-insensitive).
2. Expanding `SENSITIVE_FIELD_NAMES` with common prefixed variants.
3. Changing the word-boundary to a look-behind for `=` or start-of-string, which is what most secrets managers produce.

A test case like `MY_API_TOKEN=verysecretvalue` not being redacted would make this gap visible.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,304 @@
...
+///
+/// Each entry is an anchored-enough pattern to be unambiguous. Order is only
+/// significant for cosmetics (longer matches first reduce churn on
+/// overlapping candidates).
+const SHAPE_PATTERNS: &[&str] = &[
+    // OpenAI / Anthropic / generic `sk-` prefixed keys (20+ chars of body)
+    r"sk-[A-Za-z0-9_\-]{20,}",
+    // GitHub personal access tokens
+    r"ghp_[A-Za-z0-9]{20,}",
+    r"github_pat_[A-Za-z0-9_]{20,}",
```
DominicBurkart, line 59: **[Unresolved from prior review — still present in 1cab51e]** URL-with-credentials gap: `https://user:password@host` URLs are not caught by either the field-name pass (the field is `user`, not a sensitive name) or the shape pass (no pattern for `://...@`). These appear routinely in `reqwest` connection debug logs, `sqlx` connection string traces, and git remote URLs.

Issue #232 lists "Generic high-entropy strings gated by prefix hints" — a URL-with-creds pattern fits squarely in the v1 scope. Suggested addition to `SHAPE_PATTERNS`:

```rust
r"https?://[^:\s]+:[^@\s]+@\S+",
```

Also missing from the current pattern list (issue #232 named these explicitly):
- AWS secret access key (40-char base64-ish, contextually paired with `AKIA`) — the PR covers access key ID but not the secret.
- `-----BEGIN PRIVATE KEY-----` / `-----BEGIN RSA PRIVATE KEY-----` PEM blocks.

These are listed as follow-ups in the PR description which is acceptable, but the URL-with-creds case is high-likelihood enough to be worth v1 given reqwest debug output.

---

harness/src/logging/mod.rs
DominicBurkart, line 327: **[Unresolved from prior review — still present in 1cab51e]** Missing `session_start`/`user_prompt` integration test — issue #232 acceptance criterion #2: "`session_start` (and any later event that accepts free-form strings) funnels payloads through the helper."

The PR wires `RedactingMakeWriter` globally at `main.rs` startup, which is a broader approach than threading it per-event, but there is no test confirming that a `user_prompt` carrying an `sk-...` key actually arrives at the sink redacted. The existing end-to-end tests use `info!(message = ...)` which covers the subscriber path, but adding one named test that mirrors the real `session_start` call site (e.g. `info!(user_prompt = %prompt, ...)`) would directly satisfy the acceptance criterion and guard against regressions if `session_start` is refactored to bypass the subscriber.

Also: `redacts_field_named_api_key` (line ~206) does not assert `out.contains("***REDACTED***")` — it only asserts the key is absent. Add the positive assertion to guard against the redaction silently dropping the entire field rather than replacing the value.

---

harness/src/main.rs
DominicBurkart, line 83: **[Unresolved from prior review — still present in 1cab51e]** Workspace scope: `RedactingMakeWriter` is only installed here in `harness/src/main.rs`. Any other binary, integration test, or example in the workspace that calls `tracing_subscriber::fmt().init()` directly bypasses redaction entirely — including any future binary added by PR #229.

Recommended: extract a public `harness::logging::init_default()` that installs the redacting subscriber, and have `main.rs` call it. That way:
- PR #229's logging init can call the same helper.
- Contributors get a single auditable call site.
- A workspace-level lint/comment can point to it.

```rust
// In harness/src/logging/mod.rs
pub fn init_default() {
    tracing_subscriber::fmt()
        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
        .with_writer(RedactingMakeWriter::new(std::io::stderr))
        .init();
}
```

This is especially important given the PR description says "once #229 merges, its richer JSON-structured `logging::init(...)` can adopt the same `RedactingMakeWriter`" — a shared `init_default()` makes that adoption mechanical.

---

DominicBurkart (commented):
Inline comments against head `1cab51e` (rebase + rustfmt on top of original implementation). The formatting and rebase are clean. All eight concerns from the prior review remain unaddressed — they are re-attached as inline threads on the current diff positions.

**Must-fix before ready-for-review (security correctness):**
1. `RedactingWriter::write` is not write-boundary-safe — a secret split across two `write` calls leaks silently. Buffer to `\n`/`flush()`. (→ `mod.rs` line 58)
2. Non-UTF-8 branch skips redaction entirely — use `from_utf8_lossy` instead. (→ `mod.rs` line 65)
3. `to_ascii_lowercase()` allocates on every log event, defeating the `Cow::Borrowed` zero-alloc fast-path. Replace with a combined `RegexSet` + `(?i)`. (→ `redaction.rs` line 136)
4. `\b[0-9a-fA-F]{32,}\b` will redact git SHAs, trace IDs, and container digests — add prefix-gating or raise the threshold; add negative tests. (→ `redaction.rs` line 65)

**Should-fix (gaps vs. issue #232 acceptance criteria):**
5. `kv_field_regex` `\b(?:token)` word-boundary misses `MY_API_TOKEN=`, `FOO_SECRET=`, etc. (→ `redaction.rs` line 93)
6. No `session_start`/`user_prompt` end-to-end test (AC #2 of issue #232); also `redacts_field_named_api_key` is missing the positive `contains("***REDACTED***")` assertion. (→ `mod.rs` line 235)
7. URL-with-embedded-credentials (`https://user:pw@host`) not covered; common in reqwest/sqlx debug logs. AWS secret access key also absent despite issue #232 naming it. (→ `redaction.rs` line 59)
8. Only `harness/src/main.rs` installs the subscriber — extract `logging::init_default()` so PR #229 and other binaries inherit redaction automatically. (→ `main.rs` line 83)

**Acknowledged as follow-up (not blocking):**
- Provider patterns (Slack `xox*`, Stripe `sk_live_`, Twilio, GCP, Azure) — listed in PR description.
- Property-based idempotency test — acknowledged follow-up.
- PII (email, phone, IP, PAN) not covered — worth a one-line doc note so future agents don't assume coverage.

Codecov reports 88.52% patch coverage (7 lines uncovered), below the 90% target — the non-UTF-8 bypass branch and related error paths are the likely gap.

---

DominicBurkart:
CI is currently failing on `lint (ubuntu-latest)` and `integration-container (ubuntu-latest)`, causing the "All Checks Passed" gate to fail. These failures block promotion to ready for review. Remaining in draft until CI is green. (Session: https://claude.ai/code/session_017vrZ4D421Ty4uKSgmm7Qsb)

---

harness/src/logging/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+/// on the lossy string before being forwarded to the inner writer.
+#[derive(Clone, Debug)]
+pub struct RedactingMakeWriter<M> {
+    inner: M,
+}
+
+impl<M> RedactingMakeWriter<M> {
+    /// Wrap an existing [`MakeWriter`] so that every emitted event is run
+    /// through [`redaction::redact_secrets`] before it reaches the inner
+    /// writer.
```
DominicBurkart, line 62: **Split-secret leak risk (security).** The flush trigger is "buffer contains `\n`". If a single `write()` call delivers a payload that *ends* with the first half of a secret and a newline arrives later, the current flush semantics are fine; but a write that contains `...secret_prefix\n...secret_suffix` will flush the entire accumulated buffer at the first `\n` — including secret_prefix which is then redacted as-is, splitting the real secret across two flushes. Concretely: if `fmt::Layer` ever emits a single `write` where a secret spans a line boundary (multi-line Display impls, pretty JSON, backtraces, `Debug` on error chains), the regex pass sees only half of the secret on each flush and both halves pass through. Consider flushing only up to the last newline seen, retaining bytes after the final `\n` in the buffer until the next write or explicit flush. The `split_write_redacts_secret` test exercises buffering but not the line-boundary edge case.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,399 @@
...
+fn kv_field_regex() -> &'static Regex {
+    static RE: OnceLock<Regex> = OnceLock::new();
+    RE.get_or_init(|| {
+        // Matches `password=...`, `password="..."`, `password: ...` up to the
+        // next whitespace / comma / closing bracket.
+        let pat = format!(
+            r#"(?i)\b(?:{names})\s*[=:]\s*(?:"(?:[^"\\]|\\.)*"|[^\s,;)\]]+)"#,
+            names = field_name_alternation()
+        );
+        Regex::new(&pat).expect("valid kv field regex")
```
DominicBurkart, line 116: **Hot-path allocation defeats `Cow::Borrowed` promise.** `input.to_ascii_lowercase()` allocates a full copy of every log line just to do a substring check, even when nothing is sensitive. This fires on 100% of log events, i.e. the common case the `Cow::Borrowed` fast path is supposed to optimize. Replace with `SENSITIVE_FIELD_NAMES.iter().any(|n| input.as_bytes().windows(n.len()).any(|w| w.eq_ignore_ascii_case(n.as_bytes())))` or a precompiled case-insensitive `RegexSet` over the field names. As-is, every `tracing::info!` pays an alloc + a lowercase pass over the entire serialized event.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,399 @@
...
+    "refresh_token",
+    "private_key",
+    "client_secret",
+];
+
+/// Regex patterns used by the shape-based redaction pass.
+///
+/// Each entry is an anchored-enough pattern to be unambiguous. Order is only
+/// significant for cosmetics (longer matches first reduce churn on
+/// overlapping candidates).
```
DominicBurkart, line 53: **`\b[0-9a-fA-F]{32,}\b` is high false-positive / false-negative.**

False positives: git SHA-256s, git commit hashes (40 hex), UUIDs stripped of dashes, OpenTelemetry trace IDs (32 hex), Docker image digests, file checksums. All of these are routinely logged in observability contexts and are not secrets. Redacting them makes logs harder to correlate with traces/commits and may break downstream log tooling. Consider at minimum excluding common non-secret hex shapes (exact 40-char for git SHA-1, exact 64-char commonly used for sha256 image digests) or, better, gate long-hex redaction behind a field-name hint (`sha256=`, `digest=`, `hash=` remain clean; bare long-hex gets redacted only inside sensitive-field values).

False negatives: actual high-entropy secrets that mix base64url characters (not hex) won't match this, but that's expected for the hex rule.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,399 @@
...
+    "authorization",
+    "bearer",
+    "gh_token",
+    "github_token",
+    "access_key",
+    "access_token",
+    "refresh_token",
+    "private_key",
+    "client_secret",
+];
```
DominicBurkart, line 47: **Missing high-value provider patterns.** Issue #232 explicitly calls out "Cloud provider tokens (Azure, GCP)" and `.env` pasted contents. Not currently covered:

- Slack: `xox[abprs]-[0-9]+-[0-9]+-[0-9a-zA-Z]+`
- Stripe: `sk_live_[0-9a-zA-Z]{24,}`, `rk_live_`, `pk_live_`
- Twilio: `AC[a-f0-9]{32}`, `SK[a-f0-9]{32}`
- Google API keys: `AIza[0-9A-Za-z_\-]{35}`
- GCP service-account PEM / JSON `"private_key": "-----BEGIN PRIVATE KEY-----"` (PEM block wraps `\n` chars so needs multi-line handling)
- URLs with embedded credentials: `https?://[^:/\s]+:[^@/\s]+@` (common in `DATABASE_URL`, git remotes)
- AWS secret access key: there is no reliable shape for the 40-char secret (issue #232 mentions "contextual AWS secret key pairs"); at minimum redact when it appears next to an `AKIA` access-key-id match.
- GitHub newer prefixes: `gho_`, `ghu_`, `ghs_`, `ghr_` (OAuth, user, server, refresh)

Safe to extend `SHAPE_PATTERNS`; PR body acknowledges this as follow-up but issue #232 lists several of these as in-scope for PR1.

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,399 @@
...
+            .collect::<Vec<_>>()
+            .join("|");
+        Regex::new(&joined).expect("valid combined shape pattern")
+    })
+}
+
+fn field_name_alternation() -> String {
+    SENSITIVE_FIELD_NAMES
+        .iter()
+        .map(|n| regex::escape(n))
```
DominicBurkart, line 90: **`kv_field_regex` has subtle under-redaction cases.**

1. `\b(?:{names})\b` with `names` including `token` will NOT match `MY_API_TOKEN=secretvalue` (because `_` is a word char, so the word boundary between `I` and `T` is absent; the `\b` before `token` fails). Many env-var-derived log lines use uppercase `X_Y_TOKEN` shape. Consider `(?:^|[^A-Za-z0-9])(?:{names})` or adding explicit uppercase variants.

2. The value terminator `[^\s,;)\]]+` does not stop at `}` or `>`, so `{api_key=abc}` captures `abc}` and the redaction overwrites the trailing `}`, corrupting adjacent JSON/HTML.

3. Values containing a literal `=` (e.g. base64 padding: `api_key=Zm9vYmFy==`) are fine (they match `[^\s,;)\]]+`), but a subsequent logfmt field on the same line without whitespace boundary would be swallowed. logfmt convention requires whitespace, but free-form `tracing` `Display` output does not.

---

harness/src/main.rs
DominicBurkart, line 83: **Workspace-scope gap.** Redaction is only installed in the `harness` binary's `main`. Other workspace binaries (any `cargo run -p ...` target) and integration test binaries will continue to emit via the default subscriber (no redaction) or via `eprintln!`/`panic!`. Issue #232's acceptance criterion mentions `session_start.user_prompt` specifically — verify that crate still routes through this subscriber and add an end-to-end integration test that emits the `session_start` event and asserts redaction. As-is, there is no test that proves the `#232`-named code path is covered.

---

harness/src/logging/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+}
+
+/// Per-write handle returned by [`RedactingMakeWriter::make_writer`].
+///
+/// Bytes are accumulated in an internal buffer. Redaction runs and the buffer
+/// is flushed to the inner writer whenever a newline (`\n`) is seen in the
+/// incoming data, or when [`flush`](Write::flush) is called explicitly. This
+/// ensures that a secret split across two `write` calls — which the
+/// `MakeWriter` contract permits — is still fully redacted.
```
DominicBurkart, line 74: **`from_utf8_lossy` on invalid UTF-8 can still split a secret.** If a stray non-UTF-8 byte lands *inside* a secret, `from_utf8_lossy` replaces it with U+FFFD (3 UTF-8 bytes), which breaks the contiguous `sk-[A-Za-z0-9_\-]{20,}` match and the secret partially survives. The `non_utf8_bytes_redacted_via_lossy` test only exercises an invalid byte *before* the secret. Add a negative-case test with the bad byte embedded in the middle of an `sk-...` key and assert redaction still succeeds (it will not, with the current regex set — either normalize by stripping FFFD before matching or relax the character class to include FFFD's UTF-8 representation).

---

harness/src/logging/redaction.rs
```diff
@@ -0,0 +1,399 @@
...
+
+    if !shapes_hit && !any_name {
+        return Cow::Borrowed(input);
+    }
+
+    let mut out: String = input.to_string();
+
+    // Pass 1: shape-based defense in depth. Done first so that values like
+    // `Bearer <token>` are redacted as a unit before the field-name passes
+    // have a chance to slice them at whitespace boundaries.
```
DominicBurkart, line 150: **Missing property/fuzz test for idempotency across inputs.** The PR body lists this as follow-up, but for a security-critical helper it is very cheap to add a `proptest!` that interleaves random alphanumeric strings with injected secret shapes and asserts (a) no substring of any injected secret survives and (b) `redact_secrets(redact_secrets(x)) == redact_secrets(x)`. The current `redaction_is_idempotent` test covers a single hand-picked input.

---

DominicBurkart (commented):
## Security-sensitive review: secret redaction

_(Posting as COMMENT since the GitHub API disallows REQUEST_CHANGES on one's own PR; treat the "Must-fix before merge" items below as blocking.)_

Thanks for the thorough PR description and explicit threat model — it made this review much easier. The post-serialization `MakeWriter` approach is sound as defense-in-depth, the `OnceLock` regex caching is appropriate, tests are extensive for a first pass, and no real secrets appear in the diff (all test secrets are obviously synthetic: `hunter2NotReallyASecret`, `AKIAIOSFODNN7EXAMPLE` is the AWS public docs placeholder, etc.).

### Must-fix before merge
1. **Buffering flush boundary** (mod.rs) — flushing on "any `\n` in buffer" can split a secret that straddles a line boundary across two redaction passes. Flush only up to the last `\n`, retain the tail. The existing `split_write_redacts_secret` test would pass either way; add a test that places a `\n` *inside* the buffered secret.
2. **Hot-path allocation** (redaction.rs) — `input.to_ascii_lowercase()` allocates for every log event, defeating the `Cow::Borrowed` optimization the rest of the function is designed around. Use a byte-level case-insensitive substring scan.
3. **`\b...token\b` misses `MY_API_TOKEN=...`** (redaction.rs) — `_` is a word character so `\b` fails. This is the most common env-var-derived shape and should match.
4. **`from_utf8_lossy` mid-secret** (mod.rs) — test only covers prefix bytes; mid-secret FFFD breaks the regex and leaks the remainder.

### Should-address
5. Long-hex rule (`{32,}` hex) will redact git SHAs, UUIDs, OTel trace IDs, image digests — significant observability regression. Gate behind sensitive-field context or exclude common non-secret lengths.
6. Missing provider patterns explicitly called out in issue #232: Azure/GCP tokens, Slack, Stripe, Twilio, Google API keys, URLs with embedded creds, newer GitHub prefixes (`gho_`/`ghu_`/`ghs_`/`ghr_`), GCP service-account PEM blocks, `.env`-pasted shapes.
7. Workspace scope: only `harness/src/main.rs` installs the subscriber; integration-test binaries and any sibling workspace crates bypass redaction. Issue #232 acceptance criterion names `session_start.user_prompt` — add an end-to-end test for that path.
8. `kv_field_regex` value terminator doesn't stop at `}`/`>`, so it corrupts adjacent JSON/HTML.
9. Property test for idempotency + no-secret-substring-survives is one-page of `proptest!` and high-value for a security primitive.

### Nits
- Module docstring mismatch with PR body: PR says non-UTF-8 is forwarded verbatim; code uses `from_utf8_lossy`. Pick one and update the other.
- `combined_shape_regex` and `shape_regex_set` duplicate the same pattern set into two compiled regexes; a single combined regex is enough.

No real secrets detected in the diff.

Session: https://claude.ai/code/session_01PATZg9FPFycvHzZoed9oiv

---

DominicBurkart (commented):
Security-sensitive review. No real secrets committed in diff (the AWS key `AKIAIOSFODNN7EXAMPLE` is AWS's documented public example; the `sk-*` / `ghp_*` / JWT strings in tests are synthetic). Approach (post-serialization `MakeWriter` byte pass) is sound and actually stronger than issue #232's `session_start`-funnel proposal because every event is covered. Tests are organized per-pattern and the UTF-8-lossy + split-write cases are nice defense-in-depth catches.

Marking COMMENT (cannot request changes on own PR) but treating the "Blocking" items below as must-fix before landing.

### Must-fix before merge

1. **Unbounded internal buffer in `RedactingWriter::write` (memory / DoS).** `self.buf.extend_from_slice(data)` only flushes on `\n`. A record without a newline — or a hostile / malformed payload — grows memory without bound. Please cap the buffer (e.g. 1 MiB) and on overflow force a flush-through-redact so the pathological case degrades gracefully instead of OOMing the process.

2. **Unflushed buffer lost on panic / drop.** `RedactingWriter` has no `Drop` impl. If the process panics (or if tracing emits a partial line at shutdown), buffered bytes held pending-newline are silently dropped — meaning the last log line before a crash never reaches stderr. Add `impl<W: Write> Drop for RedactingWriter<W>` that calls `self.flush()` on a best-effort basis. This is especially important because the last log line is typically the most useful one operators have after a crash.

3. **Over-redaction of common non-secrets by the long-hex pattern.** `\b[0-9a-fA-F]{32,}\b` will redact:
   - git commit SHAs (40 chars) — logged constantly in this repo's domain,
   - MD5 (32) / SHA-1 (40) / SHA-256 (64) content digests,
   - ULIDs / base16-encoded UUIDs,
   - container image digests (`sha256:<hex>` — the `\b` anchor still allows the hex body to match).

   Issue #232 explicitly says *"Generic high-entropy strings gated by prefix hints"* — this pattern has no prefix gating. Recommend either dropping it or gating behind prefix hints (e.g. `(?:key|secret|token|password|auth)[^A-Za-z0-9]*[0-9a-fA-F]{32,}`). Please add a regression test asserting a bare git SHA is **not** redacted so the decision is explicit and lasting.

### Non-blocking

4. **Short-circuit does extra work.** `input.to_ascii_lowercase()` allocates on every event even when shapes would have short-circuited. Swap order: check `shape_regex_set().matches()` first and only lowercase-scan for field names if shapes missed. Hot path — every log line.

5. **`kv_field_regex` value char class is missing `}`.** `[^\s,;)\]]+` does not stop at `}`, so `{secret=value}` will redact the trailing brace too. Add `}` to the negated set.

6. **`Bearer` overlap.** `bearer` is in `SENSITIVE_FIELD_NAMES` and `Bearer\s+...` is a shape pattern. No correctness bug — the shape pass runs first, which is the right choice (redacts `Bearer <tok>` as a unit before kv slices at whitespace) — but worth a code comment.

7. **Cheap patterns the threat model acknowledges but which are one-line additions.** Particularly Google API keys `AIza[0-9A-Za-z\-_]{35}`, Slack `xox[abprs]-[0-9a-zA-Z\-]+`, Stripe `sk_live_[0-9a-zA-Z]{24,}`. Standard in secret-scanner rulesets.

8. **Test gaps.** Missing: (a) over-redaction / false-positive tests (git SHA, UUID, content digest), (b) buffer-overflow behaviour once (1) is fixed, (c) panic-flush once (2) is fixed, (d) CRLF line boundaries on Windows-origin input. Property test correctly called out as follow-up.

9. **`redact_secrets` always returns `Cow::Owned` once any pass fires, even when the inner `replace_all` produced no change.** Not wrong, minor perf. Preserve `Cow::Borrowed` through passes by checking `Cow::is_borrowed()` before promoting to `String`.

10. **Alignment note for issue #232.** The issue asks for `session_start` to funnel `user_prompt` through the helper explicitly. This PR achieves the stronger invariant (all events via `MakeWriter`) but doesn't update `session_start` call-sites. Recommend a one-line note on the issue explaining why direct call-site redaction is unnecessary once the writer is installed — otherwise a future reviewer will reasonably ask.

### Confirmed clean

- No real credentials in test fixtures.
- `main.rs` wiring correctly installs the redactor on stderr.
- Threat-model documentation is unusually thorough and accurate.
- Idempotency test is present and meaningful.

Overall: good direction, but please address (1)–(3) before marking ready-for-review.

Review session: https://claude.ai/code/session_01PATZg9FPFycvHzZoed9oiv

---

PR #258, opened 2026-04-23
feat(entities/ast): RustAstEntity first slice (#23 partial)

### LINKED ISSUE #23: Entity Type: AST & Filesystem Entities (ast-entities)
## Overview

Implement AST (Abstract Syntax Tree) and filesystem entities for complete, queryable, and modifiable code structure representation. This is part of the entity management system (#20).

## Entity Type: AST & Filesystem

**Purpose**: Provide complete, queryable AST representation of code with fallback strategies for all file types.

**Labels**: `entity-management`, `entity-type:ast`

### Programming Languages
- **Rust** - Using `syn` or `tree-sitter-rust`
- **Python** - Using `rustpython-parser` or `tree-sitter-python`
- **JavaScript/TypeScript** - Using `swc` or `tree-sitter-javascript`
- **Java** - Using `tree-sitter-java`

### Configuration & Data Formats
- **YAML** - Using `serde_yaml` with structure preservation
- **TOML** - Using `toml` crate with location tracking
- **JSON** - Using `serde_json` with position info
- **CSV** - Using `csv` crate with schema inference

### Build & Infrastructure
- **Dockerfile** - Custom parser or `tree-sitter-dockerfile`
- **Nix** - Using `rnix-parser`
- **Makefile** - Custom parser for GNU Make syntax
- **CMake** - Custom parser for CMakeLists.txt

### Shell Scripts
- **POSIX sh** - Using `tree-sitter-bash`
- **Bash** - Using `tree-sitter-bash`
- **Zsh** - Using `tree-sitter-bash` with zsh extensions

### Text Files (no AST available)
```rust
pub struct TextFileEntity {
    pub path: PathBuf,
    pub line_count: usize,
    pub char_count: usize,
    pub lines: Vec<LineInfo>,
    pub encoding: String,
}

pub struct LineInfo {
    pub number: usize,
    pub content: String,
    pub char_count: usize,
}
```

### Binary Files
```rust
pub struct BinaryFileEntity {
    pub path: PathBuf,
    pub size: u64,
    pub mime_type: String,
    pub base64_content: String,
    pub metadata: FileMetadata,
}
```

### Free Text Search
- Full-text search across all file contents
- AST-aware search (e.g., "find all function definitions")
- Regex pattern matching
- Case-sensitive/insensitive options

### Graph Relationships
- Files ↔ Line segments
- Functions ↔ Call sites
- Imports ↔ Definitions
- Types ↔ Usage locations

### Structural Navigation
- "Find all functions calling X"
- "Show implementations of trait Y"
- "List all imports from module Z"
- "Find dead code (unused functions)"

### Storage
- Efficient AST representation in memory
- Incremental parsing for large files
- Cached parse results with invalidation
- Position-preserving modifications

### Querying
- Tree-sitter query language support
- Custom query DSL for common patterns
- Graph traversal for relationships
- Fuzzy matching for symbols

### Modification
- Precise source location editing
- AST-aware refactoring operations
- Automatic formatting after changes
- Validation before applying changes

### 1. Core AST Trait (`harness/src/entities/ast/mod.rs`)
```rust
#[async_trait]
pub trait AstEntity: Send + Sync {
    /// Parse file into AST
    async fn parse(path: &Path) -> Result<Self, AstError>
    where
        Self: Sized;

    /// Query AST nodes matching pattern
    async fn query(&self, pattern: &str) -> Vec<AstNode>;

    /// Find relationships to other entities
    async fn find_relationships(&self) -> Vec<EntityRelationship>;

    /// Modify AST node at location
    async fn modify(&mut self, location: Location, modification: Modification) -> Result<(), AstError>;

    /// Render AST back to source code
    async fn to_source(&self) -> String;
}

pub struct AstNode {
    pub kind: NodeKind,
    pub location: Location,
    pub content: String,
    pub children: Vec<AstNode>,
}

pub struct Location {
    pub file: PathBuf,
    pub start_line: usize,
    pub start_col: usize,
    pub end_line: usize,
    pub end_col: usize,
}
```

### 2. Language-Specific Implementations

Each language gets its own module:
- `harness/src/entities/ast/rust.rs`
- `harness/src/entities/ast/python.rs`
- `harness/src/entities/ast/javascript.rs`
- etc.

### 3. Query System (`harness/src/entities/ast/query.rs`)
```rust
pub struct AstQuery {
    pattern: String,
    language: Language,
    scope: QueryScope,
}

pub enum QueryScope {
    AllFiles,
    Directory(PathBuf),
    SingleFile(PathBuf),
}

impl AstQuery {
    pub async fn execute(&self) -> Vec<QueryResult> {
        // Execute query across scope
    }
}
```

### 4. Relationship Graph (`harness/src/entities/ast/graph.rs`)
```rust
pub struct RelationshipGraph {
    nodes: HashMap<EntityId, AstEntity>,
    edges: Vec<Edge>,
}

pub struct Edge {
    from: EntityId,
    to: EntityId,
    relationship: RelationshipType,
}

pub enum RelationshipType {
    Calls,
    Imports,
    Implements,
    Extends,
    References,
}
```

### 5. Integration with RAG
- Index AST nodes for semantic search
- Graph-based context retrieval
- Relationship-aware querying
- Code structure understanding

### Test Contracts (write BEFORE implementation)
```rust
#[cfg(test)]
mod tests {
    #[test]
    fn test_rust_ast_parsing() {
        // Parse valid Rust code
    }

    #[test]
    fn test_python_ast_parsing() {
        // Parse valid Python code
    }

    #[test]
    fn test_query_function_definitions() {
        // Find all function defs
    }

    #[test]
    fn test_relationship_graph_building() {
        // Build call graph
    }

    #[test]
    fn test_ast_modification() {
        // Modify and re-render
    }

    #[test]
    fn test_text_file_fallback() {
        // Handle non-AST files
    }

    #[test]
    fn test_binary_file_handling() {
        // Base64 encode binaries
    }

    proptest! {
        #[test]
        fn test_ast_roundtrip(source: String) {
            // Property: parse → modify → render = valid
        }
    }
}
```

### Integration Tests
- Multi-language codebase parsing
- Cross-file relationship tracking
- Large file handling
- Incremental update scenarios

### Coverage Target
- 95%+ code coverage
- All supported languages tested
- Edge cases for each file type

## Dependencies

```toml
[dependencies]
# Rust parsing
syn = { version = "2.0", features = ["full", "extra-traits"] }
quote = "1.0"

# Tree-sitter for multi-language
tree-sitter = "0.20"
tree-sitter-rust = "0.20"
tree-sitter-python = "0.20"
tree-sitter-javascript = "0.20"
tree-sitter-bash = "0.20"

# Alternative parsers
swc_ecma_parser = "0.140"  # JS/TS
rnix = "0.11"              # Nix

# Data formats
serde_yaml = "0.9"
toml = "0.8"
csv = "1.3"

# File handling
walkdir = "2.4"
mime_guess = "2.0"
base64 = "0.21"
```

## Success Criteria

- [ ] All AST types implemented and queryable
- [ ] Free text search across all entities
- [ ] Graph relationships working
- [ ] Modification operations preserve formatting
- [ ] Fallback strategies for non-AST files
- [ ] Integration with RAG system
- [ ] 95%+ test coverage
- [ ] Documentation complete

## Related Issues

- #20 - Parent issue: Entity Management
- #22 - Git entities (for tracking AST changes)

## Implementation Notes

- Use tree-sitter for consistent multi-language support
- Implement lazy parsing for performance
- Cache AST with file modification tracking
- Support incremental updates
- Handle syntax errors gracefully
- Preserve comments and whitespace where possible

## Architecture Reference

See [ARCHITECTURE.md](../ARCHITECTURE.md#ast--filesystem-entities) for architectural context.

DominicBurkart:
Part of #23 "Entity Type: AST & Filesystem Entities (ast-entities)".

## Summary

First slice of the AST entity work. Adds a `syn 2.0`-based Rust AST entity that coexists with the existing `FileEntity` / `WorkspaceScanner`. This PR is intentionally scoped to a single language and a narrow query surface so that it can land, get reviewed, and establish the pattern for the full multi-language rollout.

### Added
- `harness/src/entities/ast/rust.rs` — `RustAstEntity`, `AstSummary`, `FunctionSig`, `StructSig`, `ImplSig`, `AstError`, `parse_rust_file`, `parse_rust_source`.
- `harness/src/entities/ast/query.rs` — `AstQuery` with `find_functions`, `find_structs`, `find_impls` (substring matchers).
- `harness/benches/ast_parse.rs` — criterion bench.
- `harness/Cargo.toml` — appended `syn = "2"` and `proc-macro2` (the latter for `span-locations` to emit line numbers).

### Preserved
- `WorkspaceScanner` and `FileEntity` in `entities/ast/mod.rs` and `types.rs` are untouched.
- No changes to `entities/mod.rs`.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p harness --all-targets --all-features -- -D warnings`
- [x] `cargo test -p harness --lib entities::ast` (24/24 pass: 19 new + 5 existing)
- [x] `cargo build -p harness --bench ast_parse`
- [x] `cargo nextest` — swarm env had no nextest; CI will run it.
- [x] `cargo deny` — currently skipped in CI; no risky licenses added (`syn` / `proc-macro2` are MIT/Apache-2).

## Follow-ups (not in this slice)

- Python / JS / TS / Java / Bash / Nix / Dockerfile / YAML / TOML / JSON AST via tree-sitter and specialized parsers.
- `BinaryFileEntity` with base64 + mime detection.
- `TextFileEntity` per-line indexing.
- Tree-sitter query DSL + full `AstQuery::execute()` workspace-wide.
- `modify(location, modification)` API + re-rendering.
- Relationship graph (Calls/Imports/Implements) populated from parsed ASTs.
- Incremental parsing / cached parse invalidation.
- Wire `WorkspaceScanner` to produce `RustAstEntity` for `.rs` files (currently still produces `FileEntity`).
- Proptest: parse → render → parse idempotency.

## Notes for reviewers

- Branch has 3 commits due to the environment's API-based push path (code-signing server was returning HTTP 400 throughout the session; local `git push` had no credentials). Recommend squash-merge.
- Commits are unsigned — if branch protection requires signed commits, either merge with admin override or re-push once the signing backend recovers.

---

- job-id: `claude-cron-20260423T192050Z-03678285`
- oldest-issue-id: `one_track#2`

Closes #23 (partial)

---
_Generated by [Claude Code](https://claude.ai/code/session_01RueT9k7KwCLy9Z1R5NE2Kh)_


harness/src/entities/ast/mod.rs
```diff
@@ -5,6 +5,12 @@
 //!
 //! See issue #23 and ARCHITECTURE.md for details.
 
+pub mod query;
+pub mod rust;
+
+pub use query::*;
```
DominicBurkart, line 11: Glob re-exports (`pub use query::*; pub use rust::*;`) at the entity root mix parser types (`AstError`, `FunctionSig`, etc.) with filesystem types from `types::*` below. As the module grows (python.rs, javascript.rs per issue #23), name collisions are likely (e.g. every language will want its own `FunctionSig`/`AstError`). Recommend explicit re-exports, or keep language parsers addressable via their language module (e.g. `ast::rust::FunctionSig`). This will save churn later.

---

harness/src/entities/ast/rust.rs
```diff
@@ -0,0 +1,439 @@
...
+    #[error("Syntax error parsing Rust source: {0}")]
+    Syntax(String),
+
+    /// The file extension / language is not supported by this parser.
+    #[error("Unsupported language (expected Rust, got: {0})")]
+    UnsupportedLanguage(String),
+}
+
+/// Signature of a top-level function declaration.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
DominicBurkart, line 34: Field name `is_pub` is misleading: `is_visible()` (line ~220) returns `true` for any non-`Inherited` visibility, including `pub(crate)`, `pub(super)`, and `pub(in path)`. The doc comment says "any visibility beyond private", which is the real semantic. Rename to `is_exported` / `is_visible` (matching the helper), or track the full `Visibility` as an enum (`Private | Crate | Super | Public`). As-is, consumers using this for symbol-index filtering will over-count "public" APIs.

---

harness/src/entities/ast/rust.rs
```diff
@@ -0,0 +1,439 @@
...
+            for item in &g.items {
+                collect_use_paths(item, prefix.clone(), out);
+            }
+        }
+    }
+}
+
+fn join_path(prefix: &str, segment: &str) -> String {
+    if prefix.is_empty() {
+        segment.to_string()
```
DominicBurkart, line 236: `type_path_string` drops generic arguments: `impl Foo for Bar<T>` yields `type_name == "Bar"`. Similarly `path_to_string` strips `PathArguments`, so `impl Service<Http> for Handler` becomes `("Handler", Some("Service"))` — indistinguishable from the non-generic case. For an AST summary that will feed RAG/graph lookups (#23 "Types ↔ Usage locations"), this loses signal. Either preserve generics in the rendered string, or add a dedicated `generics: Vec<String>` field. Minimally, add a test documenting the current behavior so it's intentional.

---

harness/src/entities/ast/rust.rs
```diff
@@ -0,0 +1,439 @@
...
+                let trait_name = i.trait_.as_ref().map(|(_, path, _)| path_to_string(path));
+                summary.impls.push(ImplSig {
+                    type_name,
+                    trait_name,
+                    line: line_of(&i.impl_token),
+                });
+            }
+            syn::Item::Use(u) => {
+                collect_use_paths(&u.tree, String::new(), &mut summary.uses);
+            }
```
DominicBurkart, line 178: The `_ => {}` arm silently drops `Item::Mod`, `Item::Trait`, `Item::Enum`, `Item::Const`, `Item::Static`, `Item::Macro`, and `Item::Type`. This is a big gap for a "queryable AST" per issue #23 ("List all imports from module Z", "Show implementations of trait Y"): trait defs and enums in particular are first-class in Rust symbol indexes. If this is intentional for the first slice, please (a) log which item kinds are skipped via a `skipped_items: usize` counter on `AstSummary`, or at minimum (b) add a `#[test]` asserting the current behavior (trait/enum/mod files parse but yield empty summaries) so the limitation is visible. Also: top-level items inside `mod foo { ... }` are never walked — that should be a follow-up ticket.

---

harness/src/entities/ast/rust.rs
```diff
@@ -0,0 +1,439 @@
...
+    use tempfile::NamedTempFile;
+
+    fn write_rs(contents: &str) -> NamedTempFile {
+        let mut f = tempfile::Builder::new()
+            .suffix(".rs")
+            .tempfile()
+            .expect("create temp .rs file");
+        f.write_all(contents.as_bytes()).unwrap();
+        f.flush().unwrap();
+        f
```
DominicBurkart, line 268: Hand-rolled FNV-1a is fine for stability, but the workspace already uses `sha2`/`blake3`-adjacent hashers and `DefaultHasher` elsewhere (check `harness/src/` for convention). If determinism is the goal, `blake3::hash(source.as_bytes()).to_hex()` gives a stable, collision-resistant fingerprint and is already a common pattern. If keeping FNV, add a brief comment noting this is deliberately not for cache keys across processes with potentially-adversarial input. Minor, but worth a single-line decision doc on why not `DefaultHasher`.

---

harness/src/entities/ast/query.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl<'a> AstQuery<'a> {
+    /// Build a new query over `entity`.
+    pub fn new(entity: &'a RustAstEntity) -> Self {
+        Self { entity }
+    }
+
+    /// Find all functions whose name contains `needle` as a substring.
+    ///
+    /// Returns an empty slice-equivalent `Vec` when no match is found.
```
DominicBurkart, line 28: This `AstQuery` diverges from the `AstQuery { pattern, language, scope }` design described in issue #23. That's defensible as a first slice, but the type name collision will be painful when the workspace-wide `AstQuery::execute()` lands — callers will have to disambiguate or rename. Consider renaming this to `RustAstQuery` (symmetrical with `RustAstEntity`) and reserving `AstQuery` for the multi-language executor described in the issue. Codecov also flags `find_structs`/`find_impls` as partially uncovered (66% patch coverage on this file) — the uncovered paths appear to be the no-match branches; add a `find_functions("")` edge-case test to bump coverage above the 90% target and unblock the codecov gate.

---

harness/Cargo.toml
```diff
@@ -30,6 +30,8 @@ glob = "0.3"
 regex = "1.0"
 toml = "0.8"
 reqwest = { version = "0.11", features = ["json", "blocking"] }
+syn = { version = "2", features = ["full", "extra-traits"] }
+proc-macro2 = { version = "1", features = ["span-locations"] }
```
DominicBurkart, line 41: Depending on `proc-macro2` as a runtime dep (not just through `syn`) solely to get `span-locations` is fine, but this feature is a known source of breakage when `syn` is used in proc-macro contexts (mismatched features across the crate graph). Please (a) add a comment here explaining *why* `span-locations` is required (line numbers in `AstSummary`), and (b) verify via `cargo tree -e features -p harness` that no transitive proc-macro crate is now forced into span-locations mode at compile time. See also https://github.com/dtolnay/proc-macro2#unstable-features for caveats.

---

harness/benches/ast_parse.rs
```diff
@@ -0,0 +1,49 @@
...
+    s.push_str("use std::collections::HashMap;\nuse std::fmt::Display;\n\n");
+    for i in 0..32 {
+        s.push_str(&format!(
+            "pub struct Item{i} {{ pub id: u64, pub name: String }}\n\
+             impl Item{i} {{\n\
+             \x20   pub fn new(id: u64) -> Self {{ Self {{ id, name: String::new() }} }}\n\
+             \x20   pub async fn load(id: u64) -> Self {{ Self::new(id) }}\n\
+             }}\n\n",
+            i = i
+        ));
```
DominicBurkart, line 23: The "realistic" fixture is 32 synthetic items of identical shape — it exercises the item-loop but not the shapes that dominate real-world parse cost (long function bodies, deeply nested generics, `macro_rules!`, doc comments, attribute macros). Consider either (a) vendoring a real file from this repo (`harness/src/entities/mod.rs` is ~500 lines and representative) via `include_str!`, or (b) adding a third `bench_parse_large` variant that scales to ~500 items so regressions become visible at realistic sizes. Without this, a 10x slowdown on real code could hide in the noise.

---

DominicBurkart (commented):
Automated review for PR #258 (draft, #23 partial, `syn`-based Rust AST first slice).

## Overall

Well-scoped first slice. Clean separation of `parse_rust_source` vs `parse_rust_file`, good error taxonomy, decent unit coverage for the happy paths, criterion bench included. The PR body is honest about what's deferred (proptest, tree-sitter, modify API, scanner wiring).

## Blockers (suggest addressing before un-drafting)

1. **`is_pub` mis-names `pub(crate)`/`pub(super)` as public** (`rust.rs`). Consumers that filter by this field for "exported API" analyses will over-count. Either rename to `is_visible`/`is_exported` or track full `Visibility`.
2. **Codecov gate failing** at 87.5% (target 90%). Uncovered branches are mostly the no-match paths in `query.rs` and error display paths in `rust.rs` — cheap to close.
3. **Silent `_ => {}` arm** drops `Item::Trait`, `Item::Enum`, `Item::Const`, `Item::Mod`, `Item::Type`, `Item::Macro`. Issue #23 explicitly lists "Show implementations of trait Y" as a query capability; without trait/enum capture the next slice has to rework the schema. At minimum, log-as-count or add a test asserting current limits.

## Non-blocking but worth addressing here

4. Generics are dropped by `type_path_string`/`path_to_string` — future RAG/graph work will want them.
5. Glob re-exports at `ast/mod.rs` will collide when Python/JS parsers land with their own `FunctionSig`/`AstError`.
6. `AstQuery` name will collide with the workspace-wide executor from issue #23 — consider `RustAstQuery`.
7. Bench fixture is too uniform; add a real-file or ~500-item variant so regressions surface.
8. Document why `proc-macro2` `span-locations` is a direct dep, and why hand-rolled FNV-1a vs `DefaultHasher`/`blake3`.

## Alignment with issue #23

Partial, as claimed. Missing from this slice (all explicitly deferred): `AstEntity` trait with async `parse`/`query`/`find_relationships`/`modify`/`to_source`, relationship graph, tree-sitter integration, proptest roundtrip, 95% coverage target, fallback strategies (`TextFileEntity`/`BinaryFileEntity`). The issue is not close-recommended; this PR is one of several slices needed.

## Secrets/PII

None observed. `syn`/`proc-macro2` are MIT/Apache-2 (OK).

## Readiness

**Not ready to merge.** Draft status is appropriate. Close the codecov gate and the `is_pub` semantics issue, then this can land as a squash-merge (per PR notes re: 3 API-pushed commits) with the other items tracked as #23 follow-ups.

---
_Automated review via claude-cron. Not a substitute for human review._

---

DominicBurkart:
**Automated CRON triage (iteration 1)**

Code review posted (COMMENT, 7 findings — `is_pub` mislabels `pub(crate)`/`pub(super)` as public (semantic bug for symbol indexing), item-walk `_ => {}` silently drops traits/enums/consts/mods (misaligned with issue #23's "Show implementations of trait Y" capability), generics dropped by `type_path_string`). No in-scope comments to address. CI was failing codecov/patch (87.5% vs 90% target) — CI-fixes subagent added unit tests for uncovered branches in `query.rs` and `rust.rs`, pushed as `922f4f75`.

Awaiting codecov re-score. Recommend addressing `is_pub` naming and missed-item arm before un-drafting.

_Stateless CRON orchestrator session._

---

DominicBurkart (commented):
Re-review on head commit 922f4f7 (previous review was on b2e7980, which has since been superseded).

---

## Summary

This PR adds the first `syn`-based Rust AST slice for issue #23: `RustAstEntity`, `AstSummary`, `FunctionSig`, `StructSig`, `ImplSig`, `parse_rust_source`, `parse_rust_file`, and `AstQuery` with `find_functions`/`find_structs`/`find_impls`. A Criterion bench is included. All 17 CI checks pass. The PR is honest about what is deferred.

---

## Issues requiring attention before merge

### 1. `is_visible` overstates visibility — field doc is inconsistent with behaviour

The `is_pub` doc comment on `FunctionSig` says _"any visibility beyond private counts"_, and `is_visible` correctly returns `true` for `pub(crate)`, `pub(super)`, and `pub(in path)`.  However, the `FunctionSig` field name `is_pub` (and the identical `is_pub` on `StructSig`) strongly implies `pub` only. Downstream callers that want "exported as part of the public API" will over-count `pub(crate)` items; callers that expect "`is_pub` means exactly `pub`" will be confused by `pub(crate)` returning `true`.

Options:
- Rename the field to `is_visible` / `is_accessible` and update docs accordingly, or
- Store the full visibility as a serialisable enum (`Pub`, `PubCrate`, `PubSuper`, `PubIn`, `Private`) so callers can filter precisely.

This is a public schema field that will appear in serialised JSON; changing it later is a breaking change.

### 2. `path_to_string` / `type_path_string` silently drops generic arguments

```rust
fn path_to_string(path: &syn::Path) -> String {
    path.segments
        .iter()
        .map(|seg| seg.ident.to_string())   // <-- PathArguments ignored
        .collect::<Vec<_>>()
        .join("::")
}
```

`impl<T> Foo<T>`, `impl Iterator<Item = u8>`, and `impl From<Vec<u8>>` all render as `Foo`, `Iterator`, and `From` respectively. The `trait_name` field for a generic trait impl loses all type parameters. This makes the `uses` collection equally lossy for `use std::collections::HashMap` (fine, no args) but will cause silent truncation for any future trait or type with parameters.

This is a data-fidelity issue: the stored summary cannot reconstruct which concrete instantiation was implemented.  Consider appending `<…>` when `PathArguments` is non-empty, or storing the raw token string via `quote::quote!(#path).to_string()` — which `syn` + `quote` makes trivial.

### 3. `AstQuery` is substring matching, not AST querying — name will collide with issue #23's executor design

The module doc is admirably honest: _"Matching is case-sensitive substring matching on names."_  But the struct is named `AstQuery`, which is the exact name issue #23 uses for its planned tree-sitter-backed query executor:

```rust
// From issue #23 implementation plan:
pub struct AstQuery {
    pattern: String,
    language: Language,
    scope: QueryScope,
}
impl AstQuery {
    pub async fn execute(&self) -> Vec<AstNode> { … }
}
```

When the tree-sitter executor lands it will need to be either a different name or it will shadow this struct.  Suggest renaming this facade to `RustAstQuery` (mirrors `RustAstEntity`) or `AstSymbolSearch` to avoid the future collision.

### 4. Glob re-exports in `ast/mod.rs` will cause name collisions when other language modules arrive

```rust
pub use query::*;
pub use rust::*;
```

`FunctionSig`, `AstError`, and `AstSummary` are re-exported without a namespace.  When `python.rs` or `javascript.rs` land with their own `FunctionSig`/`AstError` the wildcard re-exports will conflict.  Prefer explicit re-exports (`pub use rust::{RustAstEntity, AstSummary, …}`) or module-namespaced access (`ast::rust::FunctionSig`).

---

## Non-blocking observations

### 5. `line_of` requires the `span-locations` feature — this should be documented

`spanned.span().start().line` returns a meaningful value only when `proc-macro2` is compiled with the `span-locations` feature, which is present in `Cargo.toml`. Without it, `.start()` always returns line 0. This is correct as written, but the `line_of` function and the `line` field on all three `*Sig` structs deserve a doc comment explaining the compile-time dependency, so anyone who removes or condenses the feature flag understands the breakage.

### 6. `collect_use_paths` uses `prefix.clone()` inside the `Group` arm — minor allocation

```rust
syn::UseTree::Group(g) => {
    for item in &g.items {
        collect_use_paths(item, prefix.clone(), out);
    }
}
```

The last iteration clones unnecessarily.  This is a micro-optimisation but for workspace-scale parsing (`use std::{…}` with many children) it adds up. Consider passing `&str` + building `String` lazily, or consuming `prefix` only on the last iteration.

### 7. The `no-extension` test writes to a derived path that may conflict with the temp file

```rust
let f = tempfile::Builder::new().prefix("noext").suffix("").tempfile()?;
let path = f.path().with_extension("");
std::fs::write(&path, b"fn main() {}\n").unwrap();
```

`NamedTempFile::path()` already has no extension (the suffix is `""`), so `with_extension("")` returns the same path, `std::fs::write` overwrites the temp file contents, and `remove_file` at the end races with the `NamedTempFile` drop which also deletes it. In practice both deletions succeed (one silently returns `NotFound`), but the test is fragile. Using a fixed temp-dir path (e.g. `env::temp_dir().join("noext_test_file")`) or simply calling `write_rs` with the existing `.rs` suffix and asserting the name rather than the extension guard would be cleaner.

### 8. Bench fixture is too homogeneous for regression detection

`realistic_rust_source` generates 32 identical struct+impl+fn triplets.  A single pathological input (deep nesting, large generics, many `use` groups) would surface `syn` performance cliffs that this fixture misses.  Consider adding a third bench variant that feeds back one of the project's own `.rs` source files via `include_str!`.

### 9. Items silently dropped — `_ => {}` deserves a metric or comment

`Item::Enum`, `Item::Trait`, `Item::Const`, `Item::Type`, `Item::Mod`, `Item::Macro`, and `Item::ForeignMod` are all silently discarded.  Issue #23 lists _"Show implementations of trait Y"_ as a first-class query; without capturing `Item::Trait` definitions the relationship graph (a required deliverable in #23) cannot be built in a later slice without a schema change.

At minimum, add a `// TODO(#23): …` comment on the wildcard arm listing what is intentionally deferred, so the next slice author knows what was dropped and why.

### 10. `RustAstEntity` does not implement `AstEntity` from issue #23

Issue #23 defines:

```rust
pub trait AstEntity: Send + Sync {
    async fn parse(path: &Path) -> Result<Self, AstError>;
    async fn query(&self, pattern: &str) -> Vec<AstNode>;
    async fn find_relationships(&self) -> Vec<…>;
    async fn modify(&mut self, location: Location, modification: Modification) -> Result<(), AstError>;
    async fn to_source(&self) -> String;
}
```

The PR implements `Entity` (the generic entity trait) but not `AstEntity`. This is fine for a partial slice, but should be called out as a deferred item in the PR description or a `// TODO(#23)` comment on the `impl Entity for RustAstEntity` block.

---

## Alignment with issue #23

| Requirement | Status |
|---|---|
| `syn`-based Rust parser | Done |
| `parse` / `query` / `find_relationships` / `modify` / `to_source` on `AstEntity` trait | Deferred |
| `AstQuery` with `QueryScope` / tree-sitter DSL | Partially delivered (substring facade only) |
| `RelationshipGraph` | Deferred |
| Tree-sitter integration | Deferred |
| `TextFileEntity` / `BinaryFileEntity` fallbacks | Deferred |
| proptest roundtrip | Deferred |
| 95% coverage target | Below target (87.5%) |
| Multi-language support | Deferred |

The partial scope is appropriate and the PR description is transparent. The items above labelled "blocking" should be resolved before this merges; the rest can be tracked as follow-up issues against #23.

---

## CI / Security

All 17 checks pass. No secrets or PII observed. `syn` and `proc-macro2` are MIT/Apache-2 licensed. `fnv1a_64` is correct — FNV-1a with the standard 64-bit offset and prime.

---
_Automated review via claude-cron. Not a substitute for human review._

---

PR #262, opened 2026-04-24
test(tools): harden path-security invariants for filesystem tools

DominicBurkart:
## Chosen gap

The agent exposes three workspace-scoped filesystem tools to model
output (`ReadFileTool`, `WriteFileTool`, `ListDirTool` in
`harness/src/tools.rs`). These are hardened by two internal helpers:

- `validate_path_within_workspace` – canonicalize, reject if outside
  the canonical workspace root.
- `validate_path_for_write` – same check for the existing path prefix,
  **plus** a component-level `..` rejection so a non-existent target
  (which `canonicalize` cannot resolve) still cannot escape.

These helpers are the sandbox boundary that prevents a compromised /
jailbroken model from reading host secrets (e.g. `/etc/passwd`,
`~/.ssh/*`) or clobbering files outside the agent workspace.

## Prior test disposition

The in-module test suite (`harness/src/tools.rs`) covers only **one**
sandbox-escape case: `ReadFileTool` against `../../../etc/passwd`
(`test_read_file_path_security`). There was no coverage for:

- Absolute paths that land outside the workspace.
- Symlinks inside the workspace whose target escapes it (this is the
  main reason the validator canonicalizes *after* joining the workspace
  root, rather than doing a textual prefix check).
- `WriteFileTool`'s `..`-component rejection, which fires on a
  *non-existent* path where `canonicalize()` cannot help.
- Any `ListDirTool` traversal / symlink escape at all.
- Positive controls that prove legitimate deep-nested access still
  works (otherwise "reject everything" would silently pass).

## Invariants now validated

New integration test at
`harness/tests/filesystem_tools_path_security.rs` (12 tests):

**ReadFileTool**
- Relative `..` traversal rejected.
- Absolute path outside workspace rejected (Unix).
- Symlink inside workspace pointing outside rejected (Unix).
- Legitimate deeply-nested read succeeds (positive control).

**WriteFileTool**
- `..` prefix rejected even when the target does not yet exist.
- Nested `subdir/../../…` traversal rejected.
- Absolute-path write outside workspace rejected (Unix), with the
  post-condition that no file was written externally.
- Legitimate write to a new deep subdirectory succeeds (positive
  control – confirms the non-existent-path branch still works).

**ListDirTool**
- Relative `..` traversal rejected.
- Absolute path outside workspace rejected (Unix).
- Default (no `path` arg) lists the workspace root (positive control).
- Symlink inside workspace pointing outside rejected (Unix).

All 12 tests pass locally (`cargo test -p harness --test
filesystem_tools_path_security`); `cargo fmt --all -- --check` and
`cargo clippy -p harness --tests --all-features -- -D warnings` are
clean.

https://claude.ai/code/session_01LEodkCtiog4uAeXmJwHnWB

---
_Generated by [Claude Code](https://claude.ai/code/session_01LEodkCtiog4uAeXmJwHnWB)_


DominicBurkart (commented):
## Review: `harness/tests/filesystem_tools_path_security.rs`

Good additions overall. The file provides real integration coverage for three tools, includes positive controls for every tool, and the `#[track_caller]` on the assertion helper is a nice touch. CI is green. A few issues worth addressing before merge — one is a genuine test-fragility risk, the rest are coverage gaps.

---

### 1. `read_file_rejects_relative_parent_traversal` — fragile on hosts where `/etc/passwd` does not exist (potential false-negative)

```rust
let result = tool.execute(json!({ "path": "../../../etc/passwd" })).await;
assert_path_security_violation(result, "read_file '..' traversal");
```

The comment acknowledges the file "may or may not exist on the test host." If it does **not** exist and the validator calls `std::fs::canonicalize` on the resolved path, `canonicalize` returns `Err(NotFound)` rather than a canonical path. The tool may then surface an `IoError` (or equivalent) instead of `PathSecurityViolation`. The `assert_path_security_violation` helper panics on any non-`PathSecurityViolation` error, so the test fails with a confusing message on any CI host that does not have that file.

**Fix**: target a path that will always exist on any Unix host (e.g. `"../../../tmp"` or `"../../../var"`), or — better — use a path that provably lands outside but stays within `../` of the temp dir root without relying on a real host path at all. Alternatively, only assert on the `/etc/passwd` path under `#[cfg(unix)]` and replace the non-`cfg`-guarded call with a known-absent synthetic path whose traversal your validator should reject on component inspection alone (before reaching `canonicalize`).

---

### 2. `read_file_rejects_absolute_path_outside_workspace` — `#[cfg(unix)]` is inside the function body, not on the function

```rust
#[tokio::test]
async fn read_file_rejects_absolute_path_outside_workspace() {
    let workspace = tempdir().unwrap();
    #[cfg(unix)]
    {
        let tool = ReadFileTool::new(workspace.path().to_path_buf());
        let result = tool.execute(json!({ "path": "/etc" })).await;
        assert_path_security_violation(result, "read_file absolute outside workspace");
    }
    let _ = workspace; // swallow unused warning on non-Unix
}
```

This is inconsistent with the other platform-specific tests in the file (e.g. `read_file_rejects_symlink_escaping_workspace`, `list_dir_rejects_absolute_path_outside_workspace`) which all use `#[cfg(unix)]` on the function itself. On non-Unix the test becomes an empty async function that passes unconditionally — a silent non-test. The `let _ = workspace` workaround is a code smell that signals the inconsistency.

**Fix**: move `#[cfg(unix)]` to the function attribute, matching the style of the other tests in this file.

---

### 3. Missing escape vector: `WriteFileTool` + symlink inside workspace pointing outside

Every other (tool × attack vector) combination is tested except:

| Tool | `..` traversal | absolute path | symlink escape |
|---|---|---|---|
| `ReadFileTool` | ✅ | ✅ | ✅ |
| `WriteFileTool` | ✅ | ✅ | ❌ |
| `ListDirTool` | ✅ | ✅ | ✅ |

A write through a symlink is arguably the highest-severity escape (agent overwrites host files), so the gap stands out. The `#[cfg(unix)]` pattern from the existing symlink tests can be reused verbatim — create a symlink inside the workspace pointing to a file outside it and assert `PathSecurityViolation`.

---

### 4. Minor: symlink-to-file vs symlink-to-directory for `ListDirTool`

`list_dir_rejects_symlink_target_outside_workspace` creates a symlink to a **directory** outside the workspace. There is no test for a symlink to a **file** outside the workspace. This matters if `ListDirTool`'s validator takes a different code path for files vs directories (e.g., it calls `read_dir` on the resolved path, which fails on a file, potentially surfacing a non-security error before the workspace check runs). Worth adding a second `#[cfg(unix)]` case or folding both into one parameterised check.

---

### 5. No test for `workspace_root` itself being a symlink

If `workspace_root` is initialized from a symlink path (e.g. `/tmp` on macOS resolves to `/private/tmp`), `starts_with` comparisons may fail even for legitimate in-workspace paths. This isn't a new problem introduced by this PR, but since this test file explicitly documents and validates the canonicalization behaviour it would be the right place to add a regression test: create a workspace, symlink to it, construct the tool with the symlink path, and verify a legitimate read still succeeds.

---

### Positive notes

- Positive controls for all three tools are present and meaningful — without them a broken validator that rejects everything would still pass.
- `write_file_accepts_new_deep_subdirectory` correctly verifies both the tool return value (`bytes_written == 8`) and the actual filesystem state, which is the right way to write a write-tool positive test.
- `list_dir_default_lists_workspace_root` with `json!({})` is a good edge case.
- `#[track_caller]` on `assert_path_security_violation` gives useful line numbers on failure.
- Tests are fully parallel-safe (each uses its own `tempdir()`).

---

**Summary**: Issues 1–3 are worth fixing before merge (1 is a correctness/fragility risk, 2 is a silent non-test on non-Unix, 3 is a missing high-severity test case). Issues 4–5 are lower priority but this is a security-critical test surface so thoroughness pays off.

---

DominicBurkart:
**CRON status** — Review posted with 3 actionable issues: (1) `../../../etc/passwd` traversal test is fragile if the file doesn't exist on CI (canonicalize returns NotFound before the security check fires), (2) `#[cfg(unix)]` placed inside the function body instead of on the `#[tokio::test]` attribute causes the test to run as an empty no-op on non-Unix rather than being skipped, (3) WriteFileTool+symlink-escape test is missing despite being the highest-severity sandbox escape vector. Branch updated to main; Codecov at 50% (target 90%), likely explained by issue #2.

https://claude.ai/code/session_01JFqu5ihNxu1KY4mQgBXvwg

---

PR #268, opened 2026-04-25
arch: remove dead code and consolidate AgentLoop constructors

DominicBurkart:
## Summary

Three small, low-risk reductions aligned with the precepts in [ARCHITECTURE.md](./ARCHITECTURE.md), [AGENTS.md](./AGENTS.md), and [TESTING.md](./TESTING.md) ("Untested code is brittle. Harden it, don't bend it."):

- `harness/src/entities/ast/types.rs`: drop the deprecated `AstEntity` type alias for `FileEntity`. No remaining call sites in the workspace.
- `harness/src/monitoring.rs`: drop the dead `system_snapshots` field on `InternalMetrics`. Never written, never read; the supporting `SystemResourceMetrics` type stays in active use elsewhere.
- `harness/src/agent/mod.rs`: have `with_llm` and `with_tools` chain through `with_entity_store` via struct update syntax, so the `EnrichingEntities` initial state and zeroed counters live in one place. Pure refactor; the four constructor entry points remain identical.

Net: 3 insertions, 39 deletions across 3 files.

## Test plan

- [x] `cargo check --workspace --all-features` clean
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
- [x] `cargo test --workspace --lib --all-features` — same 2 pre-existing failures as `main` (`eval::swebench::tests::materialize_from_url_*`, unrelated to this change)
- [ ] CI green

---
_Generated by [Claude Code](https://claude.ai/code/session_01FzKpNrStguLtayd2ia2SKq)_

PR #269, opened 2026-04-25
test: cover tool-call diagnostics helpers and constructor invariants

DominicBurkart:
## Summary

Adds six focused unit tests in `harness/src/agent/mod.rs` covering surface that the existing test suite exercised only transitively:

- `extract_tool_calls_from_history`: pairs assistant tool calls with their matching tool responses, keeps unanswered calls with an empty result, and ignores orphan tool responses. This helper drives every `AgentError::*` diagnostic, so invariants here are load-bearing for debuggability.
- `extract_result_summary`: skips assistant turns that emit tool calls, returns empty when no plain assistant turn exists.
- All four `AgentLoop` constructors initialize to a clean `EnrichingEntities` state with zeroed counters and empty histories — a regression guard against future struct-update refactors of `with_llm` / `with_tools`.

Aligns with [TESTING.md](./TESTING.md):
- Unit tests for core deterministic helpers (the explicit lowest-priority test type, but appropriate here because the helpers are pure and complex).
- "Untested code is brittle. Harden it, don't bend it."

## Test plan

- [x] `cargo test --workspace --lib --all-features agent::tests::extract` — 5 passed
- [x] `cargo test --workspace --lib --all-features agent::tests::all` — 1 passed
- [ ] CI green

---
_Generated by [Claude Code](https://claude.ai/code/session_01FzKpNrStguLtayd2ia2SKq)_


DominicBurkart (commented):
Overall this is a solid set of focused unit tests — the naming is clear, the doc-comments explain the invariant under test, and placement in `harness/src/agent/mod.rs` follows TESTING.md conventions. A few issues worth addressing before merge, roughly in priority order.

---

**1. Constructor test uses `Box<dyn Any>` + `downcast_ref` unnecessarily (lines 2124–2145)**

The indirection through type erasure adds complexity and an `unwrap()` that will panic with a cryptic message if the wrong concrete type is ever boxed. All four constructors already return `AgentLoop<InMemoryEntityStore>`, so a plain array works:

```rust
let agents = [
    AgentLoop::new(AgentConfig::default()),
    AgentLoop::with_entity_store(AgentConfig::default(), InMemoryEntityStore::new()),
    AgentLoop::with_llm(AgentConfig::default(), InMemoryEntityStore::new(), providers[0].clone()),
    AgentLoop::with_tools(AgentConfig::default(), InMemoryEntityStore::new(), providers[0].clone(), ToolRegistry::new()),
];
for agent in &agents {
    assert_eq!(agent.state(), &AgentState::EnrichingEntities);
    // ...
}
```

---

**2. Constructor test does not assert the LLM-provider and tool-registry fields (lines 2143–2152)**

The docstring says the test guards "which LLM/tool dependencies are wired in", but the loop body never checks `llm_provider` or `tool_registry`. A struct-update refactor of `with_llm` that accidentally cleared `llm_provider` would silently pass. Either assert per-constructor (four separate `#[test]` fns) or at minimum add:

```rust
// after constructing with_tools agent specifically:
assert!(agent.llm_provider.is_some());
assert!(agent.tool_registry.is_some());
// and for new / with_entity_store:
assert!(agent.llm_provider.is_none());
assert!(agent.tool_registry.is_none());
```

The `progress_counter` field is also unasserted; `assert!(agent.progress_counter.is_none())` should be added alongside the others.

---

**3. Constructor test docstring describes a refactor that hasn't landed yet (line 2114)**

> "chain through `with_entity_store` via struct update syntax"

PR #268 (still open/draft as of this review) is the PR that would make that true. Right now each constructor spells out every field independently. The comment is a forward reference to #268 rather than a description of the current code — it will be accurate once #268 merges, but is misleading in the meantime. Suggest: reference #268 explicitly ("see #268") so the relationship is clear to future readers, or update the comment after #268 lands.

---

**4. Missing edge case for `extract_result_summary` — "most recent" semantics (line 2092)**

`extract_result_summary` uses `.rev().find(...)`, returning the **most recent** plain assistant turn. The two existing tests cover (a) last turn is plain assistant, and (b) no plain assistant turn. Neither covers the case where a plain assistant turn appears *before* a subsequent tool-calling turn:

```rust
// history: [user, assistant("interim"), assistant_tool_calls(...)]
// extract_result_summary should skip the tool-calling turn and return "interim"
```

This is the exact boundary the `.rev()` is there to handle; it deserves a test.

---

**5. Minor: `configs` array is inconsistently used (lines 2121–2140)**

`configs[0]` feeds `AgentLoop::new`, `configs[1]` feeds `with_entity_store`, but `with_llm` and `with_tools` call `AgentConfig::default()` directly. Either use `configs` uniformly or drop the array and call `AgentConfig::default()` at each site.

---

**Nit: `extract_tool_calls_pairs_calls_with_their_responses` — sort is correct but worth a comment (line 2051)**

The sort is necessary because `extract_tool_calls_from_history` iterates a `HashMap` internally. A one-line comment (`// sort because HashMap iteration order is non-deterministic`) would make this self-documenting for future readers.

---

DominicBurkart:
Review posted with 5 issues: unnecessary `Box<dyn Any>` type erasure in constructor tests, missing assertions on `llm_provider`/`tool_registry`/`progress_counter` fields (a struct-update mistake would pass silently), a docstring claiming constructor chaining that is PR #268's goal (not yet landed), a missing `.rev()` boundary edge case for `extract_result_summary`, and inconsistent use of the `configs` array across constructor tests. CI is fully green across all platforms. A fix agent is addressing these before the PR is opened. [Session](https://claude.ai/code/session_01415kJkxATT8RQs5MXke3TX)

---

Issue #92, opened 2026-03-22
eval: resource utilization tracking


DominicBurkart:
Part of #84.

Extend the eval runner to capture computational resource metrics.

## Scope

Metrics to capture:
- CPU time (milliseconds)
- Peak memory (bytes)
- Model inference duration
- API call counts and token counts

Implementation:
- Linux: capture via `/proc/` filesystem or `procfs` crate
- Extend model provider interface to return token and duration data
- Enhance reports with: resource summary tables, cost estimation, comparative Mermaid visualizations
- Missing metrics display "N/A" gracefully

## Acceptance criteria

- [ ] CPU time and peak memory captured on Linux
- [ ] Token counts and inference duration surfaced in reports
- [ ] Report renders resource summary table and Mermaid visualization
- [ ] Missing metrics handled gracefully with "N/A"
- [ ] `cargo clippy` and `cargo fmt` pass

## Dependencies

Blocked by #89 (eval runner).

Issue #84, opened 2026-03-22
agentic eval suite


DominicBurkart:
Post-MVP, nanna now needs a means to evaluate changes to the harness in terms of outcomes. The ideal is to formalize this in terms of CI/CD, though key evaluation metrics are too costly to run per commit (example: the full SWE-bench suite). 

An MVP eval suite needs the following minimal capabilities:
1. Run nanna against SWE-bench and report results (including ancillary statistical information such as result histograms and quantiles) in a concise, structured markdown report with associated Mermaid graphs. MVP UX: the job is triggered manually on a PR and the report is posted as a comment on the PR. 
2. The capacity to create and check in happy-path evals similar to SWE-bench for targeted workflows of interest in dummy repos checked into this one. These will, for now, have an equivalent report as the SWE-bench one.

Extended eval capacity would involve:
1. The ability to compare outcomes between nanna and another harness, e.g. the one used in swebench, using the same model, with similar report-comments as described above. 
2. The ability to estimate resource utilization in terms which describe direct computational resources (mCPU-seconds + memory budget) and can easily be used to estimate cost (now or based on resource costs at a different time)., 
3. The ability for evals involving nanna being used in a remote environment by an agent run by a foundational model (focus for now on Claude Code, either their background agent offering (preferred) or if necessary their CLI. Evaluating variation in cost (quantified as: tokens in / out per model and USD), the difference of outcomes with and without nanna usage (see SWE-BENCH MVP reqs), duration (interesting by itself, critical for subsequent climate impact estimations), and estimated climate impact (foundational model resources used [amortized hardware, electricity, water] vs local model [amortized hardware cost, electricity (drawing from a residential address in Paris, FR), no water as it's a 2025 framework 13" with Ryzen mainboard]).  

- [literature-review.md](https://github.com/user-attachments/files/26166809/compass_artifact_wf-2dd3370f-f3f7-4650-b01b-0487de34ae9b_text_markdown.md)
- [swebench.com](https://www.swebench.com/). Note: I'm interested in the multilingual and verified sets the most, the former because as of writing nanna only works in Rust and the latter because the recent attention to human curation is interesting.

PR #272, opened 2026-04-25
ci(eval): run happy-path cases against gemma4:e4b, fix #235

### LINKED ISSUE #235: ci: ignored model-integration tests never run — Ollama probe on :11434 always fails

DominicBurkart:
## Summary

Three CI changes that, together, take the eval suite from "scaffolding only" to "exercises the agent end-to-end on every workflow_dispatch":

1. **`.github/workflows/ci.yml`** — drop the broken Ollama probe in the `integration-container` matrix leg. The probe checked `localhost:11434`, but nothing in the job ever started Ollama there, so `--run-ignored ignored-only` was always silently skipped — every model-backed integration test passed CI without ever running. Fixes **#235**.

2. **`.github/workflows/eval.yml`** — add a "Run happy-path eval cases" step that invokes the new `run_eval` runner (from #271) via `cargo nextest run --features eval-runner --run-ignored ignored-only -E 'test(eval_runner)'`. Default model bumped to `gemma4:e4b`; both `gemma4:e4b` and `qwen3:0.6b` remain in the workflow_dispatch choice list. Workflow now exports `NANNA_EVAL_MODEL` alongside `MODEL`. Ollama startup hardened (60s wait, retry/timeout around `ollama pull`). Added `happy-path-output.txt` to the uploaded `eval-results` artefact.

3. **`harness/tests/{integration_tests,dev_container_integration,nanna_self_dev_integration}.rs`** — replace `const E2E_MODEL: &str = "gemma4:e4b"` with an `e2e_model()` helper reading `NANNA_EVAL_MODEL` (default `gemma4:e4b`), so the workflow's model choice flows into the E2E tests. Mirrors the salvageable parts of #236 without taking on that PR's competing runner design.

## Why this is a separate PR from #271

#271 adds the `eval-runner` feature and the `run_eval` function. This PR's happy-path workflow step depends on that feature existing. Splitting them keeps both diffs small and reviewable. **Do not merge this before #271.**

## Stacked on #271

This PR is opened against `claude/eval-runner-issue-89` (the branch behind #271). Once #271 lands, GitHub will auto-rebase against `main`. If you'd prefer it against `main` directly, that's also fine — the conflict surface is small (this branch carries one extra commit on top of #271's tree).

## Test plan

- [x] `cargo check --workspace --all-features --tests` clean
- [x] `cargo nextest run --workspace --all-features` — 610 passed, 17 skipped (the 17 are `#[ignore]` tests requiring Ollama)
- [ ] CI green
- [ ] Manual `Eval Suite` workflow_dispatch run against `gemma4:e4b` produces a green `eval-results` artefact with happy-path-001/002/003 pass marks
- [ ] Push a deliberately-failing ignored test on a throwaway branch and confirm CI fails (validates that #235's silent-skip is gone)

## Notes

- **`workflows` permission caveat**: the GitHub App used by this session may not have `workflows` permission. If the push of `.github/workflows/*` was blocked, the maintainer needs to apply the workflow diff manually. (My push succeeded today, so this caveat may not apply.)
- The pre-existing `cargo deny` advisory on `rand 0.8.5` is unrelated.
- Defers #236's container-orchestrated runner design and resource utilization (#92).

Closes #235.
Part of #84, #129.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,11 +147,7 @@ jobs:
       shell: bash
       run: |
         nix develop --command cargo nextest run --workspace --test '*' --all-features --test-threads=1
-        if curl -s --max-time 5 http://localhost:11434/api/tags > /dev/null 2>&1; then
-          nix develop --command cargo nextest run --workspace --test '*' --all-features --run-ignored ignored-only --test-threads=1
-        else
-          echo "Ollama not reachable on port 11434, skipping model integration tests"
-        fi
+        nix develop --command cargo nextest run --workspace --test '*' --all-features --run-ignored ignored-only --test-threads=1
 
     - name: Run lint checks
       if: matrix.test-type == 'lint'
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -52,10 +52,18 @@ jobs:
       run: |
         curl -fsSL https://ollama.com/install.sh | sh
         ollama serve &
+        READY=false
         for i in $(seq 1 60); do
-          ollama list > /dev/null 2>&1 && break
+          if ollama list > /dev/null 2>&1; then
+            READY=true
+            break
+          fi
           sleep 1
         done
+        if [ "$READY" != "true" ]; then
+          echo "::error::ollama daemon never became ready within 60s" >&2
+          exit 1
+        fi
         for attempt in 1 2 3; do
           if timeout 600 ollama pull "$MODEL"; then
             break
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -79,6 +79,19 @@ jobs:
         fi
         ollama list
 
+    # Test-image / model coverage notes:
+    #   * This workflow installs Ollama directly on the runner and pulls $MODEL,
+    #     so NANNA_EVAL_MODEL flows end-to-end into the happy-path runner.
+    #   * The harness integration tests in ci.yml's "integration-container" leg
+    #     reference a pre-baked qwen3 image (nanna-coder-test-ollama-qwen3:latest,
+    #     defined in nix/containers.nix). When that image is present locally the
+    #     ContainerConfig short-circuits model_to_pull (see container.rs ~line 321),
+    #     so those tests effectively run against qwen3 regardless of NANNA_EVAL_MODEL.
+    #     That mismatch is pre-existing and orthogonal to this workflow; aligning
+    #     test images with NANNA_EVAL_MODEL is tracked separately.
+    #   * --workspace --features eval-runner: cargo applies the feature only to
+    #     packages that declare it (the eval crate); other workspace members
+    #     compile without it. The flag is package-scoped, not workspace-wide.
     - name: Run eval suite
       id: eval
       env:
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -64,13 +64,19 @@ jobs:
           echo "::error::ollama daemon never became ready within 60s" >&2
           exit 1
         fi
+        PULLED=false
         for attempt in 1 2 3; do
           if timeout 600 ollama pull "$MODEL"; then
+            PULLED=true
             break
           fi
           echo "ollama pull attempt ${attempt} failed; retrying..."
           sleep 5
         done
+        if [ "$PULLED" != "true" ]; then
+          echo "::error::ollama pull failed for $MODEL after 3 attempts" >&2
+          exit 1
+        fi
         ollama list
 
     - name: Run eval suite
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -10,9 +10,10 @@ on:
       model:
         description: 'Model to use for evaluation'
         required: false
-        default: 'qwen3:0.6b'
+        default: 'gemma4:e4b'
         type: choice
         options:
+          - gemma4:e4b
           - qwen3:0.6b
           - llama3.1:8b
       case_filter:
@@ -51,17 +52,24 @@ jobs:
       run: |
         curl -fsSL https://ollama.com/install.sh | sh
         ollama serve &
-        # Wait for Ollama to be ready
-        for i in $(seq 1 30); do
+        for i in $(seq 1 60); do
           ollama list > /dev/null 2>&1 && break
           sleep 1
         done
-        ollama pull "$MODEL"
+        for attempt in 1 2 3; do
+          if timeout 600 ollama pull "$MODEL"; then
+            break
+          fi
+          echo "ollama pull attempt ${attempt} failed; retrying..."
+          sleep 5
+        done
+        ollama list
 
     - name: Run eval suite
       id: eval
       env:
         MODEL: ${{ inputs.model }}
+        NANNA_EVAL_MODEL: ${{ inputs.model }}
         CASE_FILTER: ${{ inputs.case_filter }}
       run: |
         echo "## Eval Results" > eval-results.md
@@ -84,20 +92,37 @@ jobs:
           --workspace --all-features \
           -E "$FILTER" \
           --no-fail-fast 2>&1 | tee eval-output.txt
-        EXIT_CODE=$?
+        UNIT_EXIT=$?
         set -e
 
         cat eval-output.txt >> eval-results.md
         echo '```' >> eval-results.md
 
-        if [ "$EXIT_CODE" -eq 0 ]; then
+        echo "" >> eval-results.md
+        echo "### Happy-path eval cases" >> eval-results.md
+        echo '```' >> eval-results.md
+
+        set +e
+        nix develop --command cargo nextest run \
+          --workspace --features eval-runner \
+          --run-ignored ignored-only \
+          -E 'test(eval_runner)' \
+          --test-threads=1 \
+          --no-fail-fast 2>&1 | tee happy-path-output.txt
+        HAPPY_EXIT=$?
+        set -e
+
+        cat happy-path-output.txt >> eval-results.md
+        echo '```' >> eval-results.md
+
+        if [ "$UNIT_EXIT" -eq 0 ] && [ "$HAPPY_EXIT" -eq 0 ]; then
           echo "outcome=success" >> "$GITHUB_OUTPUT"
           echo "" >> eval-results.md
           echo "**Status: PASSED**" >> eval-results.md
         else
           echo "outcome=failure" >> "$GITHUB_OUTPUT"
           echo "" >> eval-results.md
-          echo "**Status: FAILED** (see logs above)" >> eval-results.md
+          echo "**Status: FAILED** (unit_exit=${UNIT_EXIT} happy_exit=${HAPPY_EXIT})" >> eval-results.md
         fi
 
     - name: Write eval summary
@@ -120,4 +145,5 @@ jobs:
         path: |
           eval-results.md
           eval-output.txt
+          happy-path-output.txt
         if-no-files-found: warn
--- a/harness/tests/dev_container_integration.rs
+++ b/harness/tests/dev_container_integration.rs
@@ -20,7 +20,9 @@ use tokio::time::timeout;
 const MAX_ATTEMPTS: usize = 5;
 const MAX_TURNS: usize = 32;
 const TEST_TIMEOUT: Duration = Duration::from_secs(600);
-const E2E_MODEL: &str = "gemma4:e4b";
+fn e2e_model() -> String {
+    std::env::var("NANNA_EVAL_MODEL").unwrap_or_else(|_| "gemma4:e4b".to_string())
+}
 
 fn example_repo_path() -> PathBuf {
     PathBuf::from(env!("CARGO_MANIFEST_DIR"))
@@ -131,7 +133,7 @@ async fn run_single_attempt(image_ref: &str) -> Result<(), String> {
         max_iterations: MAX_TURNS,
         verbose: true,
         system_prompt: "You are a coding assistant. Modify this Rust project to compute the first 10 prime numbers instead of Fibonacci numbers. Update src/lib.rs to implement a `primes(n: usize) -> Vec<u64>` function that returns the first n prime numbers. Update src/main.rs to call `primes(10)` and print the result. Update tests/fib_test.rs to test that primes(10) returns [2, 3, 5, 7, 11, 13, 17, 19, 23, 29]. Use run_command to run `cargo test` and `cargo run` to verify your changes work.".to_string(),
-        model_name: E2E_MODEL.to_string(),
+        model_name: e2e_model(),
     };
 
     let context = AgentContext {
@@ -222,7 +224,7 @@ async fn test_task_manager_submit_with_dev_container() {
                 .to_string(),
             repo_path,
             "HEAD".to_string(),
-            E2E_MODEL.to_string(),
+            e2e_model(),
             MAX_TURNS,
             provider,
         )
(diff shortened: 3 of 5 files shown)


DominicBurkart (commented):
Automated review — CRON workflow, session https://claude.ai/code/session_01415kJkxATT8RQs5MXke3TX

## Summary

The three-part change correctly diagnoses and addresses the silent-skip bug from #235, propagates the model env var cleanly into the test harness, and hardens Ollama startup. The core logic is sound. Four issues are worth addressing before marking this ready and before merging on top of #271.

---

### [BLOCKER] `ollama pull` retry loop silently succeeds even when all attempts fail

**File:** `.github/workflows/eval.yml`, Ollama startup step (around line 63–70 in the post-patch file)

```yaml
for attempt in 1 2 3; do
  if timeout 600 ollama pull "$MODEL"; then
    break
  fi
  echo "ollama pull attempt ${attempt} failed; retrying..."
  sleep 5
done
ollama list
```

If all three `timeout 600 ollama pull` invocations fail the loop exits normally and the job continues. The subsequent `ollama list` will show an empty registry, but the "Run eval suite" step will still launch, hit a missing-model error, and produce a `HAPPY_EXIT` failure rather than a clear setup-failure message. More critically, depending on how Ollama handles a missing model at inference time, this could also produce a cryptic failure that's hard to diagnose.

Suggested fix:

```yaml
PULL_OK=false
for attempt in 1 2 3; do
  if timeout 600 ollama pull "$MODEL"; then
    PULL_OK=true
    break
  fi
  echo "ollama pull attempt ${attempt} failed; retrying..."
  sleep 5
done
if [ "$PULL_OK" != "true" ]; then
  echo "::error::Failed to pull $MODEL after 3 attempts"
  exit 1
fi
ollama list
```

---

### [BLOCKER] Ollama readiness loop has no exit-nonzero path

**File:** `.github/workflows/eval.yml`, same startup step (around line 56–61 post-patch)

```yaml
for i in $(seq 1 60); do
  ollama list > /dev/null 2>&1 && break
  sleep 1
done
```

If `ollama serve` never becomes responsive (daemon crash, port conflict, etc.) this loop exits after 60 seconds with exit code 0 and the job proceeds into `ollama pull`. Add a sentinel:

```yaml
READY=false
for i in $(seq 1 60); do
  if ollama list > /dev/null 2>&1; then
    READY=true
    break
  fi
  sleep 1
done
if [ "$READY" != "true" ]; then
  echo "::error::Ollama did not become ready within 60 s"
  exit 1
fi
```

Both sentinel patterns are consistent with the TESTING.md principle: "Surface failures as early as possible" and "Never merge … silent failures."

---

### [CONCERN] `test_image` name vs. pulled model mismatch in integration tests

**File:** `harness/tests/integration_tests.rs`, tests `test_e2e_container_to_validated_inference`, `test_e2e_performance_and_reliability`, `test_e2e_agent_with_containerized_ollama`

After this PR the `model_to_pull` field correctly reads from `NANNA_EVAL_MODEL` (defaulting to `gemma4:e4b`), but `test_image` still references `"nanna-coder-test-ollama-qwen3:latest"`. If the CI runner uses a container image baked with qwen3 weights and the test then tries to pull and run `gemma4:e4b` against it, the behaviour depends on whether the container image is model-agnostic (i.e., just an Ollama daemon) or pre-baked with a specific model. If the latter, these tests will fail at inference time even when `NANNA_EVAL_MODEL` is set correctly.

This may be a pre-existing issue unrelated to this PR's diff, but the `e2e_model()` change makes the inconsistency more visible. Please confirm whether `nanna-coder-test-ollama-qwen3:latest` is a model-agnostic Ollama daemon image or a pre-baked-weights image, and update the image name (or add a comment) so the intention is clear.

---

### [NITS] Minor workflow hygiene

**File:** `.github/workflows/eval.yml`

1. The "Run eval suite" step opens a markdown code-fence with `echo '```'` but never explicitly closes the happy-path fence before the status line — it relies on the second `echo '```'` after `cat happy-path-output.txt`. This is correct structurally but fragile if `happy-path-output.txt` itself contains a triple-backtick. Consider a different delimiter (e.g., `---` or an HTML comment) for the report fences.

2. `--workspace --features eval-runner` is used in the happy-path step. `--workspace` combined with `--features` applies the feature only to packages that declare it; packages that don't will still compile without it. This is fine here but worth documenting in a comment so future editors don't assume `eval-runner` is activated workspace-wide.

---

### Stacking risk (informational, not blocking)

This PR is correctly based on `claude/eval-runner-issue-89` (#271's branch) and the `--features eval-runner` invocation requires the feature to exist. The draft status and the "Do not merge before #271" note in the PR body are the right controls. One additional safeguard: consider adding a comment in `eval.yml` near the happy-path step noting the `eval-runner` feature dependency, so a future rebase or cherry-pick doesn't silently lose it.

---

### What's correct and well-done

- **`ci.yml` fix**: Removing the broken Ollama probe and running `--run-ignored ignored-only` unconditionally is the right call per issue #235's option (1). The tests manage their own container lifecycle, so no pre-started Ollama is needed.
- **`e2e_model()` helper**: Replacing the `const E2E_MODEL` with an env-reading function is a clean, minimal change. The default preserves existing behaviour; CI override flows naturally from `NANNA_EVAL_MODEL`.
- **60-second wait + 3-attempt pull**: Both are meaningful improvements over the prior 30-second wait with no retry. The retry sleep (5 s) is appropriate given typical Ollama model-pull failure modes.
- **`HAPPY_EXIT` distinction in failure message**: Splitting `UNIT_EXIT` and `HAPPY_EXIT` in the status line gives operators immediately actionable signal when debugging.
- **Artefact inclusion of `happy-path-output.txt`**: Correct and complete.


---

DominicBurkart:
Review posted with 2 blockers in `eval.yml`: the `ollama pull` retry loop exits 0 even when all retries fail (silently proceeding against a missing model), and the 60s readiness poll loop has no failure path if Ollama never starts. PR CI itself is clean (3/3 eval checks pass), but merge is blocked because the stacked PR #271 has a Codecov patch coverage failure (71.42% vs 90% threshold) — #272 cannot land until #271 does; once #271 merges, the two retry/readiness fixes in this PR should be addressed before opening. [Session](https://claude.ai/code/session_01415kJkxATT8RQs5MXke3TX)

---

DominicBurkart:
Both `eval.yml` blockers from the prior review are already addressed on the current head (`9da2044`):

- Ollama readiness poll: commit `74b9bc2` ("ci(eval): hard-fail when ollama daemon never reaches ready") — the 60s loop now sets `READY=true` on success and the step `exit 1`s with a `::error::` annotation if the daemon never responds (eval.yml lines 55-66).
- `ollama pull` retry loop: commit `66b6701` ("ci(eval): hard-fail on ollama pull retry exhaustion") — `PULLED=true` is only set on a successful `timeout 600 ollama pull`, and the step `exit 1`s with a `::error::` annotation after 3 failed attempts (eval.yml lines 67-79).

No further code changes needed for those two items. Codecov/`#271` blockers are out of scope for this PR.

---

DominicBurkart (commented):
Automated review by coding agent — re-reviewing head 9da2044 (after the previous owner review at 11:26Z). Focus: alignment with #235, correctness of the new happy-path step, and silent-skip regressions.

The owner-review BLOCKERs (sentinel-checked `ollama serve` readiness and `ollama pull` retry) are correctly addressed in 9da2044 — both now `exit 1` on exhaustion. `gemma4:e4b` is a real Ollama model (https://ollama.com/library/gemma4:e4b, ~9.6 GB), so the "Gemma 4" name is fine. Three new concerns introduced by the happy-path step itself:

---

### [BLOCKER] Nextest filter `test(eval_runner)` matches zero tests

`.github/workflows/eval.yml`:

```yaml
nix develop --command cargo nextest run \
  --workspace --features eval-runner \
  --run-ignored ignored-only \
  -E 'test(eval_runner)' \
  --test-threads=1 \
  --no-fail-fast
```

`test(<substr>)` filters by **test name**, not binary name. The four `#[ignore]` happy-path tests in `harness/tests/eval_runner_tests.rs` are named `test_run_eval_returns_result`, `test_run_eval_timeout`, `test_run_eval_isolation`, `test_discover_and_run_all_cases` — none contain the substring `eval_runner`. So this step will discover zero tests, exit 0, and silently report PASSED. This recreates exactly the silent-skip pathology of #235 that the PR is meant to fix.

Use the binary selector instead: `-E 'binary(eval_runner_tests)'` (or `test(/run_eval/)` to match by name pattern). Recommend adding the deliberate-fail validation step from the PR's own test plan to the merge-blocking checklist — that step would have caught this.

---

### [BLOCKER] Happy-path tests silently skip on `ModelProvider` error

`harness/tests/eval_runner_tests.rs` (all four `#[ignore]` tests):

```rust
Err(EvalRunnerError::ModelProvider(msg)) => {
    eprintln!("test_run_eval_returns_result: model provider unavailable ({msg}); skipping");
    return;
}
```

Returning early treats provider unavailability as a pass. In the eval workflow Ollama is a hard prerequisite (we just exited 1 if it didn't start), so a `ModelProvider` error here is a real CI failure, not a skip. As written, even if the filter bug above is fixed, a transient Ollama crash mid-suite produces a green PR comment. Either `panic!` on `ModelProvider` in this context, or gate the skip behind an explicit env var (e.g. `NANNA_ALLOW_PROVIDER_SKIP=1`) that CI does **not** set.

---

### [BLOCKER] `NANNA_EVAL_MODEL` does not reach the happy-path runner

`harness/src/eval/runner.rs:67` hardcodes:

```rust
fn default() -> Self {
    Self { model_name: "qwen3:0.6b".to_string(), ... }
}
```

and the `#[ignore]` tests call `EvalRunnerConfig::default().with_max_iterations(...)` without overriding the model. So even after the workflow exports `NANNA_EVAL_MODEL=gemma4:e4b` and `ollama pull gemma4:e4b` succeeds, the happy-path tests will request `qwen3:0.6b` from Ollama and fail (or skip silently per the issue above). The PR description's claim that "the workflow's model choice flows into the E2E tests" only holds for the harness/tests/ E2E binaries — not for the new happy-path runner.

Add `.with_model(&std::env::var("NANNA_EVAL_MODEL").unwrap_or_else(|_| "gemma4:e4b".to_string()))` to each `EvalRunnerConfig::default()` call in `eval_runner_tests.rs` (or change the runner default to read the env var, mirroring `e2e_model()` in the harness tests).

---

### [CONCERN] ci.yml integration-container leg now unconditionally runs the heavy E2E suite

Removing the probe is the right fix per #235 option (1), but the three `test_e2e_*` tests in `harness/tests/integration_tests.rs` (`...containerized_ollama`, `...performance_and_reliability`, `...container_to_validated_inference`) each carry `E2E_TIMEOUT = 300s` and pull/start container Ollama. On every PR's `integration-container` leg they will now execute serially (`--test-threads=1`). Worth confirming the runner has the disk + time budget, and consider whether these belong in the gated `Eval Suite` workflow rather than per-PR CI. Not a blocker for #235's correctness, but a regression in CI minutes you should size before merging.

---

### [NIT] Markdown fence collision

`echo '\`\`\`' >> eval-results.md` around the happy-path block: if `happy-path-output.txt` ever contains a triple-backtick (likely once you start logging agent tool output verbatim) the rendered comment breaks. Same comment from the prior owner review still applies.

---

### What's correct

- ci.yml probe removal matches #235 option (1) cleanly.
- 60s readiness loop and 3-attempt pull both now fail loudly (good fix of prior owner BLOCKERs).
- `e2e_model()` env-reading helper in the three harness/tests/ files is clean and minimal.
- Documentation comment about the qwen3 pre-baked image vs. `NANNA_EVAL_MODEL` mismatch is honest and points at the right follow-up.
- `UNIT_EXIT` / `HAPPY_EXIT` split in the status line gives operators actionable signal.

Sources:
- [ollama.com/library/gemma4:e4b](https://ollama.com/library/gemma4:e4b)

---
_Generated by an automated coding-agent review pass. Read-only audit — no commits made._

---

DominicBurkart:
Closing as superseded by the stack #289 → #291 (workflow + runner) and #290 (#235 fix). Owner review on this PR flagged three blockers in the happy-path step (nextest filter `test(eval_runner)` matched zero tests, NANNA_EVAL_MODEL didn't reach EvalRunnerConfig, ModelProvider errors silently skipped). #291 takes the dedicated-bin approach which avoids all three — filter is explicit, errors propagate, and the bin reads NANNA_EVAL_MODEL directly. The ci.yml probe-removal portion is in #290; the eval.yml gemma4:e4b portion is in #289; the e2e test helper changes are deferred (the workflow doesn't currently set NANNA_EVAL_MODEL on ci.yml, so they're a separate PR if/when wanted).

DominicBurkart:
Superseded by stack #289 → #291 + #290 (avoids the owner-flagged blockers).

PR #273, opened 2026-04-25
feat(eval): SWE-bench data model and report generator (#129)

DominicBurkart:
## Summary

Adds SWE-bench domain types and a Markdown+Mermaid report renderer, plus a `swe-bench-report` CLI subcommand for off-line rendering. Standalone — no dependency on #271 or #272.

This is a rebased version of #134 onto current `main`, with one bug fix surfaced during rebase: `swebench_report.rs` was importing from a `swebench` module that didn't exist in the PR's tree (the data model lived in `swebench_results.rs`). Fixed in this PR.

## What lands

- **`harness::eval::swebench_results`** — `SweBenchRunConfig`, `SweBenchInstanceResult`, `TokenUsage`, `SweBenchRunResult`. Helpers: `resolved_count`, `total_count`, `resolve_rate`, `total_claude_tokens`, `avg_wall_time`, `tokens_per_resolved`.
- **`harness::eval::swebench_report`** — `SweBenchReport` Markdown+Mermaid renderer. `render_markdown()` produces header, summary, pie chart (resolution status), xychart bar charts (wall time + Claude token usage per instance), and per-instance detail table. `write_to_directory(base)` writes to `<base>/<sha>/<bench>/<scenario>/report.md`; `write_comparison_to_directory(other, base)` writes a side-by-side comparison.
- **`swe-bench-report` clap subcommand** on `harness::main` — `--input <run.json> [--output-dir <base>] [--compare <other.json>]`.
- **`scripts/swebench-instances.txt`** — curated instance list (no execution yet — consumed by the future SWE-bench runner).

## Differences from #134

- Rebased on current `origin/main` (no merge conflicts; PR's `main.rs` had drifted vs current main, so the subcommand was re-applied cleanly).
- Fixed import path in `swebench_report.rs` (was `super::swebench::SweBenchRunResult`, now `super::swebench_results::SweBenchRunResult`).
- Doctests in `swebench_report.rs` updated to use the correct module path.
- Tests' `use crate::eval::swebench::*` updated to `use crate::eval::swebench_results::*`.
- **Did NOT pull in `model/src/anthropic_config.rs`** from #134 — it's unused by SWE-bench reporting and looked like future-Anthropic-provider scaffolding. Splitting it out keeps the diff focused; can land separately when consumers exist.

## End-to-end verification

```bash
nix develop --command cargo run -p harness --all-features -- \
  swe-bench-report --input /tmp/swebench-fixture.json --output-dir /tmp/swe-out
# -> Report written to: /tmp/swe-out/<sha>/<bench>/<scenario>/report.md
```

I rendered a 2-instance fixture (1 resolved, 1 with build failure) and confirmed the Markdown output contains the summary table, pie chart, both bar charts, and the per-instance details table. The hyphenated subcommand name `swe-bench-report` is what clap produces from the `SweBenchReport` enum variant; document if the user prefers `swebench-report`.

## Test plan

- [x] `cargo check --workspace --all-features` clean
- [x] `cargo nextest run --workspace --all-features` — 575 passed, 13 skipped
- [x] `cargo test --workspace --doc --all-features` — 5 doctests passed (judge module — no SWE-bench doctests run, but module imports compile cleanly)
- [x] End-to-end `swe-bench-report --input fixture.json --output-dir /tmp/swe-out` produces a valid report tree with all charts.
- [ ] CI green

## Notes

- Pre-existing `cargo deny` advisory on `rand 0.8.5` is unrelated.
- Defers #134's open inline review items (markdown-cell escapes for `error` and `instance_id`, char-based `truncate_id`/`truncate_label`, doc on `tokens_per_resolved` integer-truncation semantics, three tempfile integration tests). They can be addressed in a follow-up — flagged as TODOs once this lands so reviewers don't re-flag them.
- Defers #134's `model/src/anthropic_config.rs` (no consumers in this scope).

Closes #134.
Part of #84, #129.


DominicBurkart (commented):
Automated review — https://claude.ai/code/session_01415kJkxATT8RQs5MXke3TX

## Summary

The rebased import-path fix is correct and the overall structure (data model / renderer / CLI subcommand) is clean. CI gate requires 100 % patch coverage (TESTING.md); a handful of untested paths below will block merge. Five correctness issues are also flagged.

---

## Blocking — correctness bugs

**1. `truncate_id` / `truncate_label` slice on bytes, not chars — panics on multi-byte instance IDs**

`&id[..13]` and `&label[..21]` index into a `&str` by *byte* offset. Any instance ID containing a non-ASCII character (e.g. a future project name with an accented letter, or a test case) causes a panic at the slice boundary. Real SWE-bench IDs are ASCII today, but this is a latent landmine and the project principle is "surface failures early".

Fix: use `id.char_indices().nth(13).map_or(id, |(i, _)| &id[..i])` (or collect into chars). The existing `test_truncate_id` test only exercises ASCII, so it won't catch this.

**2. `tokens_per_resolved` silently truncates — misleading metric, undocumented**

```rust
self.total_claude_tokens() / resolved as u64   // integer division
```

For a run with 3 resolved instances and 10 001 total tokens this returns 3 333 instead of 3 333.67. The PR description flags this as a deferred TODO from #134, but the doc-comment says "Average Claude tokens per resolved instance" with no truncation caveat. Either switch to `f64` or update the doc comment to "integer-truncated average …" before landing. Silently misleading metrics are worse than no metrics.

**3. `generate_swebench_report` default output dir is relative to CWD, not documented**

```rust
let base_dir = output_dir.unwrap_or_else(|| std::path::Path::new("results"));
```

`Path::new("results")` is resolved relative to the process working directory at the time of the call. When the binary is invoked from a CI script that `cd`s into a temp dir, reports land in an unexpected location. The clap help string says `(default: results/<sha>/<bench>/<scenario>/)` but does not say "relative to CWD". At minimum, add that to the help text. Better: resolve to an absolute path using `std::env::current_dir()?.join("results")` and propagate the error instead of a silent `unwrap_or_else`.

**4. `render_comparison_bar_chart` only charts resolve-rate — title claims "Resolve Rate and Avg Time"**

The method writes one `bar` series (resolve rate %) but the chart title says `"Scenario Comparison: Resolve Rate and Avg Time"`. Mermaid `xychart-beta` requires a second `bar` or `line` directive for the second series. As written, the chart renders only resolve rate while the title promises wall-time data, which misleads readers. Either drop "and Avg Time" from the title, or add a second series.

**5. `render_differing_instances` silently ignores instances present in B but not in A**

```rust
for (id, &a_resolved) in &a_map {
    if let Some(&b_resolved) = b_map.get(id) { ... }
}
```

Instances that appear only in run B (new instances added in the second scenario) are never emitted in the diff table. The "No instances differ" path can therefore fire even when the two runs cover different instance sets. Add a complementary pass over `b_map` keys not present in `a_map`, or document that the comparison is intentionally restricted to the intersection.

---

## Blocking — 100 % patch coverage gate

TESTING.md: *"Codecov patch coverage must be 100% on every PR."*

Untested lines in this diff:

| Location | Untested path |
|---|---|
| `swebench_results.rs` — `total_claude_tokens` | No test that verifies `nanna_token_usage` is *not* summed (the field exists but is always `None` in all fixtures; a test with a populated `nanna_token_usage` would confirm the metric is Claude-only) |
| `swebench_report.rs` — `render_comparison_bar_chart` | No test asserts the chart content or that `y_max` clamps at 100.0 when both resolve rates exceed ~83 % |
| `swebench_report.rs` — `render_differing_instances` table header + rows | `test_comparison_report` asserts instance IDs appear in the output but does not assert the column headers `\| Instance \| scenario_a \| scenario_b \|` or the specific YES/NO values in each row |
| `main.rs` — `generate_swebench_report` with `--compare` | The `compare` branch (second `serde_json::from_str` + `write_comparison_to_directory`) has no integration test; `write_comparison_to_directory` is tested in `swebench_report.rs` but the CLI wiring is not |
| `main.rs` — `generate_swebench_report` error paths | Bad JSON / missing file paths are untested |

The two `write_to_directory` / `write_comparison_to_directory` unit tests in `swebench_report.rs` use `tempfile::tempdir()`. Verify `tempfile` is listed in `[dev-dependencies]` in `harness/Cargo.toml`; it is not visible in this diff.

---

## Non-blocking observations

**`TokenUsage` re-exported with an alias that conflicts with the module's own name**

`eval/mod.rs` re-exports `TokenUsage as SweBenchTokenUsage`. Consumers who import via `harness::eval::*` get `SweBenchTokenUsage`; consumers who import `harness::eval::swebench_results::TokenUsage` get the original name. This two-name situation will cause confusion when the type appears in error messages. Consider either using the alias everywhere (rename the struct) or removing the alias and keeping the original name at the public boundary.

**`SweBenchReport` is not `#[non_exhaustive]` and all fields are `pub`**

All three fields (`title`, `generated_at`, `run_result`) are `pub`. Adding a field later (e.g. a report metadata struct) is a breaking change for any code that constructs `SweBenchReport { title: …, generated_at: …, run_result: … }` outside the crate. Since the only stable constructor is `SweBenchReport::new`, consider making the fields private or marking the struct `#[non_exhaustive]`.

**`write_to_directory` uses `commit_sha` / `bench_name` / `scenario` as path components without sanitization**

If any of these strings contain path separators (`/`, `..`), the output path escapes the intended base directory. For a read-only CLI tool this is low-severity, but since `commit_sha` comes from deserialized JSON it is attacker-controlled if the input file is untrusted. Add a validation step (e.g. reject strings containing `/` or `\`) or at minimum document the assumption.

**`ARCHITECTURE.md` lists six CLI subcommands; this PR adds a seventh**

The architecture doc enumerates `chat`, `agent`, `mcp-serve`, `models`, `tools`, `health`. `swe-bench-report` is not listed. Update the doc as part of this PR so the architecture stays accurate (it is the project's primary navigation aid for agents).

**`scripts/swebench-instances.txt` — no consumer yet, no schema doc**

The file is described as "consumed by the future SWE-bench runner" but there is no `#[test]` that parses it (to catch malformed lines introduced by future edits) and no doc comment in the file itself explaining the format beyond the inline `#` comments. A trivial parse test — iterate lines, skip `#`, assert each non-empty line matches `r"^[a-z0-9_-]+__[a-z0-9_-]+-\d+$"` — would lock in the format and is cheap to add.

**`generated_at` captured in `SweBenchReport::new`, not at render time**

`SweBenchReport::new` calls `chrono::Utc::now()` but `render_markdown` is typically called later. If a report is created, stored, then rendered hours later, the "Generated" timestamp will be the creation time. This is probably fine for offline rendering, but worth a doc comment to set expectations.


---

DominicBurkart:
Review posted with 5 blocking correctness bugs: `truncate_id`/`truncate_label` slice by byte offset and panic on multi-byte characters in instance IDs or scenario names; `render_comparison_bar_chart` title claims "Resolve Rate and Avg Time" but only renders one series; `render_differing_instances` silently uses intersection-only comparison without documentation; `tokens_per_resolved` does integer truncation with no doc caveat; and `generate_swebench_report` defaults to a CWD-relative path with no documentation. CI is fully green (all 20 checks pass, Codecov 99.1%). Blocking bugs must be fixed before this PR can be opened for human review. [Session](https://claude.ai/code/session_01415kJkxATT8RQs5MXke3TX)

---

Issue #276, opened 2026-04-25
Guard against patch-coverage bypass via #[cfg(not(tarpaulin))] / #[ignore] annotation growth


DominicBurkart:
## Context

PR #275 lands an agent-proof regression guard for `codecov.yml` (closes #211). However, the 100% patch-coverage floor can still be bypassed by an agent without ever touching `codecov.yml`:

- Adding `#[cfg(not(tarpaulin))]` (or `#[cfg(not(tarpaulin_include))]`) to source files removes lines from the coverage build entirely.
- Adding `#[cfg_attr(coverage_nightly, coverage(off))]` does the same on nightly.
- Marking tests `#[ignore]` so they don't exercise their target lines.
- Moving uncovered code into a file matching `tarpaulin.toml`'s `exclude-files` glob.

Originally raised by the bot reviewer on PR #224 ([comment 3103966406](https://github.com/DominicBurkart/nanna-coder/pull/224#discussion_r3103966406)). It's a real gap: with the codecov.yml side now locked down by #275, this is the obvious next bypass surface.

## Proposed approach

A separate workflow that runs on PRs and computes the **net delta** of these annotations + ignored tests in source files:

- `git diff $base..$head -- '*.rs'` and count occurrences of:
  - `#[cfg(not(tarpaulin))]`
  - `#[cfg(not(tarpaulin_include))]`
  - `#[cfg_attr(coverage_nightly, coverage(off))]`
  - `#[ignore]` on `#[test]` / `#[tokio::test]`
- Also detect growth of `tarpaulin.toml`'s `exclude-files`/`exclude` lists (similar to the codecov-guard's `ignore` check).
- If net additions > 0, fail the check unless the PR description carries an admin-acknowledgment label (e.g. `coverage-exception-approved` applied by the owner).

Same trust-anchor design as #275: the workflow lives in `.github/workflows/`, no scripts/ indirection. Add the workflow path and `tarpaulin.toml` to the Repository Ruleset path-restriction so agents cannot weaken the guard.

## Acceptance criteria

- [ ] New workflow under `.github/workflows/` (e.g. `coverage-bypass-guard.yml`) with all logic inlined.
- [ ] Detects net additions of the listed annotations across PR diff and fails the check.
- [ ] Allows reductions (cleanup) without complaint.
- [ ] Label-based escape hatch applied by repo admin (or admin force-merge).
[...]

PR #275, opened 2026-04-25
ci(codecov): agent-proof regression guard + 100% floor (closes #211, supersedes #224)

DominicBurkart:
## Summary

Lands the 100% patch-coverage floor with a guardrail that survives the threat model the owner raised on PR [#224](https://github.com/DominicBurkart/nanna-coder/pull/224#discussion_r3105030677):

> "how is this a guardrail? what stops an agent from changing this too? most but not all agents working in this repo can't change the github workflows. a big issue is that the codecov configuration doesn't live there, [it] can be manipulated."

The fix is structural — policy logic now lives only in the trust anchor, not in agent-writable territory.

## Changes

| File | Change |
|---|---|
| `codecov.yml` | Patch target 90% → **100%**. Adds `codecov.strict_yaml_branch: main` so PR-time `codecov/patch` checks always use `main`'s YAML. |
| `.github/workflows/codecov-guard.yml` | **New.** All policy logic inlined — no `scripts/*.sh` indirection. Uses `yq` for proper YAML parsing (handles block + flow style, decimal targets). Fails closed on every edge case. |
| `.github/CODEOWNERS` | **New.** Surfaces `@DominicBurkart` as touched-paths reviewer for `codecov.yml`, `.github/workflows/*`, `.github/CODEOWNERS`, `tarpaulin.toml`, `.github/rulesets/*`. Solo-maintainer-friendly — does not block self-merges. |
| `AGENTS.md` | Documents the threat model, trust anchor, admin-bypass escape hatch, and owner-side one-time UI setup. |

## How this addresses the open review on #224

- **Comment 3105030677 ("how is this a guardrail?")** — The guard logic is fully inlined in `.github/workflows/codecov-guard.yml`. There is no `scripts/check-codecov-guard.sh`. Trust anchor and policy are the same file.
- **Comment 3103963364 / 3103964277 (no override)** — No override path: no commit trailer, env var, or script flag. Admin force-merge is the only bypass.
- **Comment 3103964506 (regex doesn't catch decimals or `auto`)** — Replaced with `yq` parse + bash `[[ =~ ^([0-9]+(\.[0-9]+)?)%?$ ]]` numeric extraction; non-numeric replacements (`auto`, missing) are explicit regressions.
- **Comment 3103964879 (awk parser fails on flow-style)** — Replaced with `yq '.ignore // [] | length'`, which handles both block and flow-style lists uniformly.
[...]

Issue #274, opened 2026-04-25
lighter verification framework


DominicBurkart:
post https://github.com/DominicBurkart/nanna-coder/issues/196: can systems with fewer guardrails at the dev container level yield better results? how are they structured? how do they perform on the evals? how can they be introduced into the strategy listing?

DominicBurkart:
TL;DR of the research note in PR #277 (`docs/research/lighter-verification-framework.md`):

- Current verification stack is heavy by design: container isolation, `nix develop`, full CI gates, 100% patch-coverage floor, plus the #275 + #276 agent-resistance guards.
- Hypothesis: lighter wins on exploratory/RAG/orchestration domains from #196; lighter loses on SWE-bench-style code-edit-with-tests. The point of A/B testing is to make the crossover explicit.
- Concrete A/B design: three variants (`heavy`/`medium`/`light`) over the runner from #271 with the cases under `evals/cases/`, with a pre-registration template and metrics that include security violations + coverage faithfulness, not just pass-rate.
- Registry integration: add a `VerificationLevel` field to the `Workflow` trait from #203 and have the selector from #204 refuse `Light` for security-sensitive tasks. Routing-level enforcement is orthogonal to the per-PR coverage guard, which stays on for all variants.
- Tension with #276 is mitigated by audit-only sandboxing + two-stage promotion (a `light` workflow's output is re-validated by a `heavy` workflow before any PR opens).

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #204, opened 2026-04-12
[adas] Workflow selector node: interpret task and route (or reject)


DominicBurkart:
Parent: #196
Depends on: #203 (Workflow trait & registry)

## Context
Extension (1) of #196. Nanna needs a routing node that inspects an incoming task, chooses among registered workflows using their `description` / `capabilities` / `suitability_hints`, or rejects the task with a semi-structured explanation if no workflow is a good fit. Today there is only one workflow, so the selector must degenerate gracefully to "use the default" when that is the only good match — but the selection logic itself must be built against the general (multi-workflow) case.

## Requirements
- Add a selector node upstream of workflow execution that takes `(task, WorkflowRegistry)` and returns either `Selected { workflow, rationale }` or `Rejected { reasons: Vec<StructuredReason> }`.
- Selection must be driven by an LLM call (local model via existing `model/` crate) conditioned on registry metadata — not hardcoded rules — so that adding a new workflow via the registry automatically makes it selectable without code changes to the selector.
- Rejections must be structured enough to be machine-readable (tag + message) so callers can branch on them.
- Emit a trace/telemetry event for each selection decision, including the rationale, so eval jobs (#84, sub-issue for complementarity evals) can analyze selector quality.
- When only one workflow is registered and it is suitable, the selector must still produce a rationale; when it is not suitable, it must reject rather than fall back silently.

## Acceptance criteria
- [ ] Unit tests with a stub registry containing 0, 1, and N workflows, covering: forced selection, forced rejection, and tie-breaking.
- [ ] Integration test using the real default workflow on a trivially-suitable task (expects `Selected`) and on a clearly-unsuitable task (expects `Rejected` with a non-empty reason list).
- [ ] Selector output is logged via the existing `observability.rs` telemetry plumbing.
- [ ] Clippy + fmt + nextest green under `nix develop --command`.

## Non-goals
- Creating additional workflows to select among.
- Ensemble voting — this is single-selection only; ensemble logic lives in the eval sub-issue.

PR #277, opened 2026-04-25
research: lighter verification framework — eval & strategy implications (#274)

DominicBurkart:
## Summary

Research note addressing issue #274: can systems with fewer dev-container guardrails yield better results, how should they be structured, how should they perform on the eval suite, and how should they be surfaced in the workflow registry?

The note lives at `docs/research/lighter-verification-framework.md` and covers:

- Definition of the current verification stack (containers, nix, CI, codecov 100%, #275 + #276 guards).
- Hypothesis: lighter wins on exploration / RAG / orchestration; loses on edit-with-tests and SWE-bench.
- Three best-effort reference points (Aider, OpenHands sandbox-off, raw-tool-use category) — flagged where hand-waved.
- A/B plan against the runner from #271 + cases under `evals/cases/`, with a pre-registration template, three variants (`heavy`/`medium`/`light`), and the extra evals needed to capture security and coverage-faithfulness regressions.
- Registry integration sketch: a `VerificationLevel` field on the `Workflow` trait from #203 + selector behaviour from #204 that refuses `Light` for security-sensitive tasks.
- Risks vs #276's coverage-bypass guard, with an audit-only sandbox + two-stage promotion as the proposed mitigation.
- Seven concrete follow-up issues to file (not filed here — research issues stay open until human review).

## Test plan

- [ ] Human review of the report's framing and recommendations.
- [ ] Decide which of the seven follow-ups to file as separate issues.
- [ ] Decide whether to leave #274 open as a meta-tracking issue or close once follow-ups are filed.

Refs #274

job-id: I7oXc
oldest-issue-id: 274

https://claude.ai/code/session_01EaRrEfDt6tn4JYs8Qe5Ebn

---
_Generated by [Claude Code](https://claude.ai/code/session_01EaRrEfDt6tn4JYs8Qe5Ebn)_


docs/research/lighter-verification-framework.md
```diff
@@ -0,0 +1,282 @@
...
+will likely *regress* on **code-edit-with-tests** domains (SWE-bench,
+test implementation, merge conflict resolution). The point of A/B testing
+is to make those crossovers explicit rather than guessed.
+
+## 3. Reference systems (best-effort; flagged where hand-waved)
+
+Three points on the verification-density axis. *Verifiable* claims cite
+public docs; *hand-waved* claims are based on general public reputation
+and need follow-up before being treated as load-bearing.
+
```
DominicBurkart, line 104: Factual accuracy: the claim that Aider "under-performs on tasks requiring multi-file orchestration with strong test signal" is presented without citation and runs against Aider's publicly reported SWE-bench Verified results (historically top-tier on the leaderboard, including multi-file edits). Per the doc's own framing for §3 ("flagged where hand-waved"), this sentence should either be cited or explicitly marked as a prior rather than a finding. Recommend softening to "Aider's published SWE-bench numbers are competitive; per-domain breakdowns by orchestration depth are not, to our knowledge, published — treat any orchestration-vs-edit-locality split as a hypothesis."

---

docs/research/lighter-verification-framework.md
```diff
@@ -0,0 +1,282 @@
...
+— are the starting suite. Ensemble/coverage analysis lives in #207's
+runner.
+
+**Variants.** Define three concrete verification levels:
+
+- `verification = "heavy"` — current default. Container-isolated, nix
+  shell, full CI gates simulated locally, 100% patch coverage required
+  for promotion, all coverage-bypass guards active.
+- `verification = "medium"` — drop the 100%-patch-coverage promotion
+  requirement (run coverage as advisory only), keep nix shell + container.
```
DominicBurkart, line 155: Verify model identifier: `gemma4:e4b` does not match a known Ollama tag (Gemma 3 ships as `gemma3:e4b`/`gemma3:4b`-style tags; no public "gemma4" release exists as of the PR date). If this is meant to mirror what `eval.yml` actually sets via `NANNA_EVAL_MODEL`, please double-check against #272's workflow file before pre-registration commits, otherwise the pre-registration id will reference an unrunnable model.

---

docs/research/lighter-verification-framework.md
```diff
@@ -0,0 +1,282 @@
...
+- Strong CI gates mean the agent can only learn from the *gate* signal,
+  not from intermediate failure modes. This collapses the reward surface
+  and reduces the diagnostic value of a failed run.
+- Coverage and lint gates create false-negatives on novel approaches:
+  a refactor that *would* pass a human reviewer's read but happens to
+  drop one untested edge-case path is killed at the gate, even when the
+  diff strictly improves architecture.
+
+**Where lighter bites.** The same gradient that frees exploration also
+frees regression:
```
DominicBurkart, line 74: Terminology nit: in standard test-theory framing, a gate that rejects code which would have passed human review is a false *positive* (positive = "rejected as bad"), not a false negative. Consider rephrasing as "false rejections" to avoid the ambiguity, since the doc audience is coding agents that may key off the term.

---

docs/research/lighter-verification-framework.md
```diff
@@ -0,0 +1,282 @@
...
+- **Coverage-bypass guard (in flight).** Issue #276 tracks closing the
+  obvious next bypass surface (`#[cfg(not(tarpaulin))]`,
+  `#[cfg_attr(coverage_nightly, coverage(off))]`, `#[ignore]` on tests, and
+  `tarpaulin.toml exclude-files` growth).
+- **Behavioural rules in AGENTS.md.** Hard prohibitions on lowering the
+  coverage target, editing the guard workflow, or replacing the numeric
+  target with `auto`. Escalation contract: file an issue rather than weaken.
+- **Eval harness.** `harness/src/eval/` plus `evals/cases/**/task.toml`
+  (e.g. `happy-path-001`, the SWE-bench-verified samples) plus the runner
+  that PR #271 lands and PR #272 wires into `eval.yml`. PR #273 lands the
```
DominicBurkart, line 35: #207 ("complementarity runner") is referenced here and again in §3/§4/§7 but never linked or one-line-defined the way #203/#204/#271/#275/#276 are in §1. For an audience of coding agents reading this cold, add #207 to the §1 inventory (or a one-clause gloss on first mention) so the §4 dependency on it is grounded.

---

docs/research/lighter-verification-framework.md
```diff
@@ -0,0 +1,282 @@
...
+  coverage". They tend to dominate on greenfield code generation and
+  underperform on regression-heavy maintenance work. This matches the
+  hypothesis in §2 and is the regime nanna-coder's lighter variant
+  should occupy.
+
+What we'd need to verify properly: per-system numbers on the same eval
+suite (SWE-bench-verified is the obvious common denominator). That's
+exactly what #207's complementarity runner is meant to produce. Until
+those numbers exist, the case studies above are *priors*, not evidence.
+
```
DominicBurkart, line 135: Internal consistency: the `light` variant is defined as "no container, run inside the harness shell directly" but §6's first mitigation says "audit-only sandbox even in `light`" (i.e. a container is still present, just non-blocking). These are reconcilable but the §4 definition should call out the audit-only container explicitly so a reader doesn't conclude that `light` literally has zero container surface.

---

DominicBurkart (commented):
Research note is well-structured and addresses all four prompts in #274 (definition, hypothesis, eval design, registry integration). Inline comments cover: one factual claim about Aider that needs softening or citing, a likely model-tag typo (`gemma4:e4b`), one terminology nit, a missing #207 gloss in §1, and a small inconsistency between §4's `light` definition and §6's audit-only-sandbox mitigation. No secrets/PII. Length and structure feel appropriate; the seven-item §7 follow-up list is a good unit of escalation. Non-blocking on factual/consistency cleanup.

---

DominicBurkart:
CRON sub-agent (job-id: No3dF) gap-closure pass on the research note. Pushed `20dbac1` addressing all five unresolved review threads:

| Thread | Fix |
| --- | --- |
| Aider §3 claim (factual accuracy) | Softened — calls out top-tier SWE-bench Verified (incl. multi-file), explicitly marks the orchestration-vs-edit-locality split as a hypothesis. |
| `gemma4:e4b` model id | Replaced with `qwen3:0.6b` (the live default in `.github/workflows/eval.yml` `workflow_dispatch.inputs.model.default`); §4 setup paragraph + pre-registration TOML both updated, with an inline NOTE requiring re-read at commit time. |
| "false-negatives" terminology | Rephrased to "false rejections" with an inline test-theory gloss (rejection = positive) so coding-agent readers don't key off the wrong term. |
| #207 not in §1 inventory | Added a "Complementarity runner (planned)" bullet to §1 and noted that §3/§4/§7 depend on it. |
| §4 `light` vs §6 audit-only inconsistency | §4 `light` definition now explicitly carves out the audit-only (non-blocking) container wrapper and points to §6, so a reader doesn't infer zero container surface. |

No structural changes to the doc's hypothesis, variant matrix, registry sketch, or §7 follow-ups — those remain as-is.

**Verification feedback loop:** the repo has no `.markdownlint*` config and no docs-lint CI workflow, and there is no `docs/strategies/index.md` (or similar strategy listing file) to wire the doc into. The note is reachable only via the `docs/research/` directory and the §5 registry sketch references planned files (#203/#204) rather than existing ones. Recommend filing a small follow-up for a strategies index once #203/#204 land — otherwise this doc, and any future `docs/research/*.md`, will remain orphan-discoverable.

PR remains **DRAFT**. CI was green on `30f4892`; the new commit only touches one markdown file, so no behaviour change expected.

Followups (not in this PR):
1. Create `docs/strategies/index.md` (or equivalent registry index) once #203/#204 land, and link this note from it.
2. Add a lightweight markdownlint + link-checker CI job for `docs/**` so future research notes get the same feedback loop as code.
3. Resolve the five review threads on this PR (now addressed in `20dbac1`) before promoting out of draft.

---

DominicBurkart (commented):
Lighter validation approaches make sense to me in several domains:
1. Where validation is deferred to CI
2. Where the domain is not meaningfully decomposable.

Lets unpack the lighter verification concept and point (2). My objective for "lighter verification" is to provide Nanna more ways of working. Globally, Nanna must have the ability to solve different classes of problems, some of which do not benefit from the default verification approaches Nanna employs at each turn. They are informative, but that information (and the resources to collect it) is wasteful when applied to irrelevant domains. Nanna shouldn't see "Software Development" as a coherent, single domain with a single validation toolset.

Nanna is specialized for rust+nix codebases, for example, and so its harness provides its models little knowledge to codebases outside of its intended domain (like all of SWE-BENCH). A more accurate replacement for "lighter verification" would be "domain-specific verification." 

As an example where verification requires architectural decisions: generating 500 lines of SQL to answer a complex question correctly requires semantic and syntactic correctness. COT etc can facilitate this through differentiating the semantic work from the syntax generation, but it's hard for small models do be 100% correct given the two problems that it needs to solve (how to translate natural language queries into a relational data model and then how to express that in correct SQL).

The conventional solution is decomposition (first transformation: language to language, from structure-agnostic to structure-specific; second transformation: structure-specific language to SQL), then per-step feedback/validation, then per-step model specialization/distillation. 

Applied to this problem, an ideal future nanna implementation would decompose this task and provide appropriate validation for each step: likely input validation (a SQL query engine is not the right place to ask about the meaning of life, and adversarial input should be flagged), LLM-as-judge for semantic analysis (example strategies: domain alignment. Does the output correspond to relationships that are possible to model directly in a SQL database? inversion. Does the output yield something semantically equivalent to the input when fed through an "agnosticator" single-turn LLM? Internal consistency. Can an agent analyze the relationships in the language and formalize them? Judge: are those invariants meaninful? Code: Could a LEAN proof then identify semantic failures and provide iteration?), then syntactic validation for the prompt (purely determinstic: is it valid SQL? Does it correspond to the actual database structure?).

Is this system design or prompt engineering? In ADAS, there isn't a meaningful difference in my opinion. Nanna is essentially formalizing and validating its chain of thought through a variety of specialized agents. 

This future Nanna has a template or menu of different ways to complete tasks which it builds/maintains itself, and can use different "golden" specialized nodes to resolve them.

Code verification is irrelevant while validating a language to language conversion, unless some aspect of that conversion is decomposed into a domain which can be expressed in code.

Sometimes, practical code tasks are mechanical: "change AAA to BBB" from the orchestrator may break down to the single step "update string AAA to BBB via an AST transform." That kind of mechanical transformation requires little to no verification: "I called the fully deterministic and tested AST replace tool with the requested parameters." Ideally Nanna would understand that and so verification there would be essentially nonexistent– if cosmic rays or bit flips mess everything up, then the orchestrator will have to sort it out with a second Nanna call. 

Nanna's job, essentially, is decomposition and domain-specific verification. That means removing domain-irrelevant verification strategies, which is more actionable than "lighter" verification.

PR #280, opened 2026-04-25
ci: skip codecov upload when CODECOV_TOKEN is absent

DominicBurkart:
## Summary

- Make `codecov/codecov-action@v5` step in the `security` job conditional on `secrets.CODECOV_TOKEN` being present.
- Today every Dependabot PR's `security (ubuntu-latest)` check fails because Dependabot-actor runs cannot read repo secrets, and the step is configured with `fail_ci_if_error: true`.
- Coverage upload is observability — not a gate. Skipping when the token is unavailable is safe.

## Context

Discovered while shipping #256 (openssl 0.10.78). The Dependabot PR was fully green except for this one step, which always fails on Dependabot-actor runs with `Upload failed: {"message":"Token required because branch is protected"}`. The workaround was a manual owner-rebase + force-push so CI re-ran under user identity with secrets available — not something the routine PR agents can or should do by default.

## Test plan

- [ ] Verify CI runs to completion on this PR (token present → upload runs as before).
- [ ] After merge, verify the next Dependabot-authored CI run shows the upload step skipped (not failed) and `security` is green.
- [ ] Verify `cargo audit/deny` skipped behavior is unchanged.


.github/workflows/ci.yml
```diff
@@ -176,7 +176,9 @@ jobs:
         fi
 
     - name: Upload coverage reports
-      if: matrix.test-type == 'security'
+      if: matrix.test-type == 'security' && env.HAS_CODECOV_TOKEN == 'true'
+      env:
+        HAS_CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN != '' }}
```
DominicBurkart, line 181: The env-indirection here is necessary because `secrets.*` is not available in step-level `if:` expressions, while step-level `env:` evaluated from `secrets` is. Worth a one-line comment in the workflow noting why the indirection exists, since a future maintainer is likely to try collapsing this into `if: ... && secrets.CODECOV_TOKEN != ''` and discover it silently evaluates to false.

Also: this job declares `environment: codecov`. Confirm via a Dependabot-authored test run (per the PR's own test plan) that the step is actually skipped (not failed). Environment-scoped secrets follow the same actor restrictions, so the fix should hold, but the test plan checkbox is currently unchecked.

---

.github/workflows/ci.yml
```diff
@@ -176,7 +176,9 @@ jobs:
...
       uses: codecov/codecov-action@v5
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
```
DominicBurkart, line 184: Minor: with the upload now conditional, `fail_ci_if_error: true` is still useful for token-present runs (catches genuine codecov outages on main/owner PRs). Leaving it as-is is the right call — flagging only so a future cleanup doesn't reflexively remove it thinking it's now dead.

Consider also emitting a one-line `echo` in a preceding step (or via `if: ... == 'false'` companion step) so Dependabot PR logs explicitly state "coverage upload skipped: no token" rather than silently omitting the step. Optional; keeps the signal that observability is intentionally degraded for that run.

---

DominicBurkart (commented):
Automated review.

Scope is correct and minimal: makes one CI step conditional on token presence to unbreak Dependabot-actor runs. The env-indirection (`HAS_CODECOV_TOKEN` from `secrets.CODECOV_TOKEN != ''`) is the standard pattern since `secrets.*` cannot be referenced in step-level `if:` directly.

Checked:
- Diff alignment with stated goal — matches.
- No other codecov/coverage usage in the workflow that would also need conditioning (grepped ci.yml; this is the only `codecov-action` invocation).
- `cargo audit`/`cargo deny` skip behavior in the security job is unchanged (separate `if`-gated branch).
- The `all-checks` gate still depends on `test-matrix`, so a skipped (not failed) upload step keeps `security` green and the gate green — the intended outcome.
- No secrets/PII exposure introduced; `token:` arg still references the secret only when present.

Concerns (non-blocking, see inline):
- Worth a brief comment in the workflow explaining the env-indirection so future maintainers don't collapse it.
- Test plan checkboxes are still unchecked; the second item (post-merge Dependabot run shows skipped, not failed) is the load-bearing verification.

LGTM from automated review pending the Dependabot-run verification noted in the PR's own test plan.

---

PR #285, opened 2026-04-26
[test] proptest invariants for ToolSpec blocklist and project_detect dedup

DominicBurkart:
## Target

Two pure, security-relevant pieces of the harness with strong invariants but only example-based coverage:

1. **`harness::onboarding::profile::ToolSpec::new`** (`harness/src/onboarding/profile.rs`) — the constructor is the sole guard that prevents an onboarding profile from naming destructive shell commands (`publish` / `deploy` / `push` / `rm -rf` / `drop` / `delete` / `destroy`). Three `#[test]` examples currently cover a small handful of literal strings.
2. **`harness::agent::project_detect::detect`** (`harness/src/agent/project_detect.rs`) — composes per-manifest detector outputs into a `DetectedProfile`, with a documented post-condition that languages, frameworks, and CI expectations are deduplicated while preserving first-seen order. The existing `composite_rust_nix_ci_profile` test verifies dedup in one fixed scenario.

## Current disposition

- `profile::tests` exercises four literal cases of the blocklist + a happy path, plus a single substring acceptance check.
- `project_detect::tests` covers each detector individually plus one composite scenario; the dedup post-condition is asserted only for `Language::Rust`.
- Neither surface had property-based coverage, so the size of the implicit input space (any user-authored `ToolSpec` command, any combination of recognized manifests in a workspace) was not exercised.

## Invariants tested

`ToolSpec::new`:
1. **Anywhere-rejection** — any command containing a single-token blocked verb (`publish` / `deploy` / `push` / `drop` / `delete` / `destroy`) as a standalone whitespace-separated token is rejected, regardless of position.
2. **Substring-acceptance** — a blocked verb that occurs only as a substring of a longer identifier (e.g. `publisher`, `redeploy`) is **not** rejected. The blocklist is whitespace-token-aware, not substring-based.
3. **Two-word window** — the multi-token phrase `rm -rf` is rejected iff its two tokens appear adjacent. A bare `rm` (without an immediately following `-rf`) is allowed.
4. **Field preservation** — when construction succeeds, `name`, `command`, `description`, and `category` round-trip verbatim into the resulting `ToolSpec`.

`project_detect::detect`:
5. **Dedup across the manifest power-set** — for any subset of `{Cargo.toml, package.json, pyproject.toml, requirements.txt, go.mod, flake.nix, .github/workflows/*.yml}`, the resulting profile contains no duplicate languages, frameworks, or CI-expectation strings — even when the same signal would be emitted by multiple detectors (e.g. `tokio` in both `[dependencies]` and `[dev-dependencies]`, or `pytest` in both `pyproject.toml` and `requirements.txt`).
6. **Determinism** — calling `detect` twice on the same workspace yields equal `DetectedProfile`s (including ordering).

## Strategy rationale

Proptest is the right tool here because both surfaces have a small number of load-bearing invariants over an awkwardly-large input space:

- The blocklist is a security boundary — its true contract is "no token-equal occurrence of a destructive verb survives", which is naturally a universal claim over commands. Listing every variation by hand drifts toward coverage padding.
- `detect`'s dedup post-condition spans `2^7 = 128` manifest subsets and the full cross-product of overlapping signals; sweeping that with proptest is far tighter than enumerating cases.

Both surfaces are pure (apart from a per-test `tempdir` for `detect`) so shrinking yields legible, minimal failure cases. The proptests live in `harness/tests/blocklist_and_detect_proptests.rs` so they run under `cargo nextest run --workspace --test '*' --all-features`, alongside the other integration tests.

## Local validation

- `cargo test -p harness --test blocklist_and_detect_proptests` — 6/6 pass.
- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p harness --tests --all-features -- -D warnings` — clean.

## CI status

The repo's CI workflows (`ci.yml`, `codecov-guard.yml`, `ci-integration.yml`) trigger only on `pull_request` events targeting `main` (or specific path filters). PRs targeting sub-branches such as `claude/great-mayer-5ajrK` therefore do not run CI checks. This PR will be exercised by CI once `claude/great-mayer-5ajrK` itself is opened against `main`. The proptests were validated locally.

## Notes

- Adds a single dev-dependency: `proptest = "1"`. `Cargo.lock` was updated locally; CI's `cargo` invocations do not use `--locked`, so the lock will be refreshed on first build. If a CI step requires a frozen lock, I will push the updated `Cargo.lock` in a fix-up commit.
- Pre-existing `eval::swebench::tests::materialize_from_url_*` failures observed locally on `claude/great-mayer-5ajrK` are unrelated to this change (network-dependent).
- Commit was authored via the GitHub Contents API (push_files) because local SSH commit-signing was non-functional in this environment (empty signing key, signing server returning `400 missing source`). The resulting commit is GitHub-server-signed.

https://claude.ai/code/session_01Ufspun8cJLsqJmeuYvP1Fk

### DIFF
--- a/harness/Cargo.toml
+++ b/harness/Cargo.toml
@@ -46,6 +46,7 @@ futures = "0.3"
 serial_test = "3.4"
 tempfile = "3.0"
 criterion = { version = "0.5", default-features = false, features = ["cargo_bench_support"] }
+proptest = "1"
 
 [[bench]]
 name = "agents_md_load"
--- a/harness/tests/blocklist_and_detect_proptests.rs
+++ b/harness/tests/blocklist_and_detect_proptests.rs
@@ -1,0 +1,393 @@
+//! Property-based invariant tests for two pure, security-relevant modules:
+//!
+//! 1. [`harness::onboarding::profile::ToolSpec::new`] — the constructor is the
+//!    sole guard against onboarding profiles that name destructive shell
+//!    commands (publish/deploy/push/rm -rf/drop/delete/destroy). Existing
+//!    example-based tests in `profile.rs` cover a handful of literal cases;
+//!    these proptests assert the underlying invariants hold across a much
+//!    broader input space.
+//!
+//! 2. [`harness::agent::project_detect::detect`] — the public composer is
+//!    documented to deduplicate signals while preserving first-seen order.
+//!    Existing tests check dedup for one variant (Rust). These proptests
+//!    exercise dedup over arbitrary file combinations.
+//!
+//! Strategy rationale
+//! -------------------
+//! - Proptest is appropriate here because both surfaces have a small number of
+//!   load-bearing invariants that are awkward to enumerate exhaustively as
+//!   examples. Both surfaces are pure (no I/O beyond a tempdir) so shrinking
+//!   yields tight failure cases.
+//! - Inputs are restricted to printable ASCII (and a curated tail of
+//!   destructive verbs) so failures are legible in shrunk form.
+
+use std::fs;
+use std::path::Path;
+
+use proptest::prelude::*;
+use tempfile::tempdir;
+
+use harness::agent::project_detect::{detect, DetectedProfile, Framework, Language};
+use harness::onboarding::profile::{ToolCategory, ToolSpec};
+
+// ---------------------------------------------------------------------------
+// ToolSpec blocklist invariants
+// ---------------------------------------------------------------------------
+
+/// Tokens the constructor must reject when they appear as standalone
+/// whitespace-separated tokens (or, for `rm -rf`, as the adjacent token pair).
+const SINGLE_TOKEN_BLOCKED: &[&str] = &["publish", "deploy", "push", "drop", "delete", "destroy"];
+
+/// Generate strings that contain only whitespace-safe printable ASCII so that
+/// `split_whitespace` tokenization is well-defined and shrunk failure cases
+/// are readable.
+fn safe_word() -> impl Strategy<Value = String> {
+    // Letters, digits, and common shell-safe punctuation. No whitespace.
+    "[A-Za-z0-9_./=:+@-]{1,12}".prop_filter("non-empty, no whitespace", |s| {
+        !s.is_empty() && !s.chars().any(|c| c.is_whitespace())
+    })
+}
+
+fn safe_command() -> impl Strategy<Value = String> {
+    proptest::collection::vec(safe_word(), 1..6).prop_map(|words| words.join(" "))
+}
+
+fn category_strategy() -> impl Strategy<Value = ToolCategory> {
+    prop_oneof![
+        Just(ToolCategory::Build),
+        Just(ToolCategory::Test),
+        Just(ToolCategory::Lint),
+        Just(ToolCategory::Format),
+        Just(ToolCategory::Check),
+    ]
+}
+
+proptest! {
+    /// Invariant: a command containing a blocked token as a *standalone
+    /// whitespace-separated token* is always rejected, regardless of where it
+    /// sits in the command line.
+    #[test]
+    fn blocked_token_anywhere_is_rejected(
+        prefix in proptest::collection::vec(safe_word(), 0..3),
+        suffix in proptest::collection::vec(safe_word(), 0..3),
+        blocked_idx in 0usize..SINGLE_TOKEN_BLOCKED.len(),
+        category in category_strategy(),
+    ) {
+        let blocked = SINGLE_TOKEN_BLOCKED[blocked_idx];
+        let mut tokens = prefix;
+        tokens.push(blocked.to_string());
+        tokens.extend(suffix);
+        let command = tokens.join(" ");
+
+        let result = ToolSpec::new("tool", command.clone(), "desc", category);
+        prop_assert!(
+            result.is_err(),
+            "blocked token {:?} in command {:?} must be rejected",
+            blocked,
+            command,
+        );
+    }
+
+    /// Invariant: substring occurrences of a blocked token *within* a longer
+    /// word (e.g. "publisher", "redeploy") must NOT be rejected. The blocklist
+    /// is whitespace-token-aware, not substring-based.
+    #[test]
+    fn blocked_substring_within_word_is_accepted(
+        prefix in "[a-z]{1,5}",
+        suffix in "[a-z]{1,5}",
+        blocked_idx in 0usize..SINGLE_TOKEN_BLOCKED.len(),
+        category in category_strategy(),
+    ) {
+        // Glue prefix + blocked + suffix into a single non-whitespace word so
+        // the blocked verb is a strict substring of a longer identifier.
+        let blocked = SINGLE_TOKEN_BLOCKED[blocked_idx];
+        let glued = format!("{prefix}{blocked}{suffix}");
+        // Sanity: ensure the blocked verb is not its own token in `glued`.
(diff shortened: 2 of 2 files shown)


DominicBurkart (commented):
## Automated Code Review — PR #285

Overall verdict: **the invariants are sound and the strategies are correctly targeting the documented contracts**, but there are a small number of correctness gaps and one meaningful test-coverage hole that should be addressed before this lands.

---

### 1. `blocked_token_anywhere_is_rejected` — strategy is correct

The test correctly constructs commands where one of the six single-token verbs appears as a standalone whitespace-separated token, then asserts rejection. This matches the actual `ToolSpec::new` implementation, which uses `split_whitespace` + `.windows(1)` matching. No issues here.

---

### 2. `blocked_substring_within_word_is_accepted` — logic flaw in the prefix/suffix strategies

The test glues `prefix + blocked + suffix` into a single word and asserts acceptance. That is the right shape. However:

- The prefix strategy is `"[a-z]{1,5}"` — minimum 1 character, non-empty, so the assertion `prop_assume!(!glued.split_whitespace().any(|t| t == blocked))` will always hold. The defensive `prop_assume!` is redundant but harmless.
- **Actual gap**: the suffix strategy is also `"[a-z]{1,5}"` (min 1), which is fine. But the strategy does not test the case where a blocked token is a *prefix* or *suffix* of the word (e.g. `"publish_tool"` or `"tool_publish"`). Those are equally important membership cases and are implicitly covered by the construction (`prefix` is always at least 1 char), so this is fine in practice.
- **Real gap**: the test does not cover the case of a blocked verb appearing as a *standalone* token surrounded by punctuation that is *not* whitespace, e.g. `"foo;publish"`. The implementation uses `split_whitespace`, so `"foo;publish"` is one token and would *not* be rejected. This is arguably correct behaviour, but the test's stated goal ("substring-acceptance") does not reach this case. Not a blocker, but worth a comment.

---

### 3. `rm_rf_phrase_is_window_matched` — one correctness gap

The test correctly checks two sub-cases: adjacent `"rm -rf"` is rejected, and bare `"rm"` (without following `"-rf"`) is accepted.

**Missing sub-case**: a stray `"-rf"` token without a preceding `"rm"` is not tested for acceptance. The blocklist entry is `"rm -rf"`, which is matched by a two-token window, so `"-rf"` alone is safe — but this is not asserted. More critically:

**Missing sub-case**: `"rm"` appearing *after* `"-rf"` (reversed order, `"-rf rm"`) should be accepted. The window matcher is order-sensitive, so this passes, but the property is not exercised.

Neither is a blocker since the implementation is clearly order-sensitive, but both would harden the test's specification of the two-token window semantics.

---

### 4. `accepted_spec_preserves_fields` — field coverage gap

The test checks `name`, `command`, `description`, and `category`. These are all four public fields on `ToolSpec`. No issues.

One minor style note: `prop_assume!` is used to filter blocked commands, which is correct but will silently discard cases where randomly generated `safe_word` tokens happen to collide with blocked verbs. The `safe_word` alphabet deliberately excludes all blocked verb characters except `"p"` and `"d"` (which appear in non-blocked words), so false discards will be rare. No action needed.

---

### 5. `detect_dedups_across_arbitrary_manifest_combinations` — one correctness issue

The `assert_no_duplicates` helper uses `sort()` + `dedup()` on a cloned vector to verify uniqueness. This is semantically correct for enums that derive `Ord`.

**Correctness gap — `ci_expectations` dedup check is too strict for the workflow test fixture**: the CI workflow fixture written in this test contains the line `"- run: cargo fmt -- --check"` **twice** in the same file:

```yaml
steps:
  - run: cargo fmt -- --check
  - run: cargo nextest run
  - run: cargo fmt -- --check   ← duplicate run: line
```

`detect_github_workflows` already deduplicates within its own output via the `seen` closure, so the profile will not contain a duplicate `CiExpectation`. The test therefore passes. However, the fixture is intentionally constructed to "provoke a dedup challenge" (per the inline comment), yet `detect_github_workflows` deduplicates *before* returning, not in `detect`. The test is correct — it exercises `detect`'s dedup path for `Language` and `Framework` signals, and the `CiExpectation` dedup is exercised by `detect_github_workflows` itself. But the inline comment overstates what is being challenged: the workflow duplicate `run:` line does not exercise `detect`'s dedup loop, only `detect_github_workflows`'s `seen` closure. This is a documentation bug, not a logic bug.

---

### 6. `detect_is_deterministic` — correct but narrow

The determinism test covers three manifest types (Cargo, Flake, Workflow). Missing from this test: `package.json`, `pyproject.toml`, `requirements.txt`, `go.mod`. For a property that is trivially guaranteed by the deterministic `detect` implementation (no randomness, sorted path iteration in `detect_github_workflows`), the narrow scope is acceptable. However, adding `has_package_json` and `has_pyproject` booleans to this test at minimal cost would improve confidence and is worth considering.

---

### 7. Missing blocklist verbs in `SINGLE_TOKEN_BLOCKED`

The PR description states the blocklist covers `publish / deploy / push / drop / delete / destroy`. The implementation in `profile.rs` confirms these six plus `"rm -rf"`. The test's `SINGLE_TOKEN_BLOCKED` array matches exactly. No gap here.

**However**: `"rm -rf"` is not in `SINGLE_TOKEN_BLOCKED`; it is tested separately via `rm_rf_phrase_is_window_matched`. This is correct by design since it is a two-token entry. The split is sensible.

---

### 8. Cargo.toml change

The addition of `proptest = "1"` under `[dev-dependencies]` is minimal and correct. No version pin beyond the major version, which is consistent with the other dev-dep entries in `Cargo.toml` (`serial_test = "3.4"`, `tempfile = "3.0"`, `criterion = { version = "0.5", ... }`). The existing crate already uses `proptest` as a `Framework` variant, so this is not a new transitive dependency on the workspace's production surface.

---

### Summary of actionable issues

| Severity | Location | Issue |
|---|---|---|
| Minor | `detect_is_deterministic` | Test only covers 3 of 7 manifest types; consider adding `has_package_json`, `has_pyproject` |
| Documentation | `detect_dedups_across_arbitrary_manifest_combinations` | Inline comment claims the duplicate `run:` line in the workflow fixture provokes `detect`'s dedup path, but `detect_github_workflows` deduplicates before returning; the comment is misleading |
| Nitpick | `rm_rf_phrase_is_window_matched` | Does not assert that reversed `"-rf rm"` or stray `"-rf"` alone are accepted; these would strengthen specification of window semantics |
| Nitpick | `blocked_substring_within_word_is_accepted` | Defensive `prop_assume!` is always true given non-empty prefix/suffix strategies; can be removed |

None of these are blockers. The six invariants are correctly stated, the proptest strategies are well-scoped, the implementation under test matches the invariants as written, and the Cargo.toml change is minimal. The PR is in good shape.

---

*Automated review by claude-sonnet-4-6 · https://claude.ai/code/session_01Wqp6Z41ZV3wWnCB7wxaKpZ*

---

DominicBurkart:
**Automated CRON status** (claude-sonnet-4-6 · https://claude.ai/code/session_01Wqp6Z41ZV3wWnCB7wxaKpZ)

Code review confirmed all 6 invariants are correctly implemented; non-blocking gaps noted: `rm -rf` reversed-order case untested, `detect` determinism test covers only 3/7 manifest types, and an inline comment overstates what the workflow fixture exercises (dedup happens inside `detect_github_workflows`, not in `detect`'s dedup loop). CI does not run on this sub-branch target (by design), branch is fresh and conflict-free. Marked ready for review — noted gaps are optional improvements.

---

PR #295, opened 2026-04-26
arch(harness): remove dead perform_entity_modification_mvp

DominicBurkart:
## Summary

This PR was opened to remove the `perform_entity_modification_mvp` function in `harness/src/agent/mod.rs`. Investigation shows the function **cannot be safely deleted yet** without additional refactoring — it is reachable via live code paths.

## Findings

`perform_entity_modification_mvp` is **not dead** — it is called from production code:

**Call site in `perform_entity_modification`** (`harness/src/agent/mod.rs`):
```rust
#[allow(deprecated)]
async fn perform_entity_modification(&mut self, context: &AgentContext) -> AgentResult<()> {
    if self.llm_provider.is_some() && self.tool_registry.is_some() {
        let provider = self.llm_provider.as_ref().unwrap().clone();
        self.perform_entity_modification_with_tools(context, &provider).await
    } else {
        self.perform_entity_modification_mvp(context).await  // <-- live call
    }
}
```

The `else` branch is reached whenever an `AgentLoop` is constructed via `AgentLoop::new(config)`, which leaves both `llm_provider` and `tool_registry` as `None`. There is also a direct test (`test_perform_entity_modification_without_tools_mvp_fallback`) that exercises this path.

The function is already annotated `#[deprecated(note = "Use with_tools constructor for LLM-driven performing")]`, confirming intent to remove — the preconditions just are not fully met yet.

## What needs to happen before removal

1. **Guard or remove the no-provider else-branch** in `perform_entity_modification` — either return an error/panic, or require callers to always use `with_tools`.
2. **Audit all callers of `AgentLoop::new`** and migrate them to `AgentLoop::with_tools`.
3. **Delete `test_perform_entity_modification_without_tools_mvp_fallback`** (the MVP fallback test).
4. **Remove `#[allow(deprecated)]`** from `perform_entity_modification` once the else-branch is gone.
5. Then delete `perform_entity_modification_mvp` itself.

https://claude.ai/code/session_011LYFrHwghQBPNBWVeVhvjd


DominicBurkart:
Closing — PR body explicitly notes the work cannot be safely deleted yet. The `perform_entity_modification_mvp` function is still alive on main (handled in #143's rebase via #[allow(deprecated)]). The right path is to first remove its callers, then delete the function in a follow-up. Linked issues remain open.

PR #315, opened 2026-04-29
test(model): probe-and-skip ollama_chat_integration when :11434 unreachable (#302)

### LINKED ISSUE #302: ci/test: gate `model/tests/ollama_chat_integration.rs` ignored tests on a real Ollama probe
Follow-up to #235 / PR #290.

PR #290 narrows the `integration-container` leg's `--run-ignored ignored-only` invocation to `-p harness --test integration_tests` because that suite manages its own Ollama container lifecycle via `harness::container::start_container_with_fallback` (port 11435). The model crate's ignored tests in `model/tests/ollama_chat_integration.rs` are out of scope for that fix and remain unrun.

## Problem

`model/tests/ollama_chat_integration.rs` contains 8 `#[tokio::test] #[ignore]` tests, all using `OllamaConfig::default()` (which targets `:11434`) via `make_provider()`. Each calls `.expect("health_check failed")` / `.expect("chat failed")`, so without a live Ollama on :11434 they panic and fail the leg loudly.

This is why PR #290 had to narrow the selector instead of running `--workspace --test '*'` ignored-only. Until these tests have their own gate, they cannot be exercised in CI without standing up Ollama on the runner.

## Proposed remediation (pick one)

1. **Probe-and-skip at runtime.** Add a `TcpStream::connect("127.0.0.1:11434")` check at the top of each test (or a `#[ctor]`/once helper) and `eprintln!("…skipping: no Ollama on :11434")` + early-return when unreachable. Mirrors the pattern in `harness/tests/dev_container_integration.rs::test_task_manager_submit_with_dev_container`.
2. **Wire through self-managed container.** Refactor the tests to use the same `start_container_with_fallback` harness already used by `harness::tests::integration_tests`, so they spin up their own Ollama on :11435.
3. **CI-side: real Ollama service.** Add an Ollama service-container to the `integration-container` matrix leg and broaden `--run-ignored ignored-only` back to `--workspace`. Heaviest option; rejected for #235 itself but valid here if (1)/(2) prove painful.

## Acceptance

- The model crate's ignored tests run in CI under whichever option is chosen, OR are skipped for a *correct* reason (probe-fail, not panic).
- The `integration-container` leg can broaden `--run-ignored ignored-only` back to `--workspace --test '*'` if option (1) or (3) is taken.
- Deliberate-red ignored test (the deferred acceptance criterion 2 from #235) added in a separate PR to validate the gate fails loudly when an ignored test has a real regression.

## Context links

- Issue #235 (root cause)
- PR #290 (narrowed selector, the line this issue unblocks)
- File: `model/tests/ollama_chat_integration.rs`
- Workflow line: `.github/workflows/ci.yml`, "Run integration tests (container)" step

DominicBurkart:
Fixes #302 (option 1).

## What

Adds a small sync helper `ollama_reachable()` to `model/tests/ollama_chat_integration.rs` that does a `TcpStream::connect_timeout("127.0.0.1:11434", 250ms)` probe, and gates each of the 8 `#[ignore]`d `#[tokio::test]`s on it:

```rust
if !ollama_reachable() {
    eprintln!("Ollama not reachable on {OLLAMA_PROBE_ADDR}, skipping <test_name>");
    return;
}
```

Without this, invoking the tests via `cargo test --run-ignored ignored-only` on a CI leg that lacks a live Ollama panics inside `OllamaConfig::default()`-backed `.expect("health_check failed")` / `.expect("chat failed")` calls. With the gate they print a skip line and pass.

## Why a sync probe inside `#[tokio::test]`?

`TcpStream::connect_timeout` on a resolved `SocketAddr` cannot block on DNS and is bounded by an explicit 250ms ceiling, so it cannot stall the runtime. This avoids needing a runtime handle just to do the probe and matches the spirit of the precedent in `harness/tests/dev_container_integration.rs::test_task_manager_submit_with_dev_container` (which uses `tokio::net::TcpStream::connect` because it is already inside an `async` block but with no timeout — here we add an explicit one).

## Why we did NOT broaden the CI selector here

Issue #302's third acceptance bullet explicitly says the `integration-container` leg can broaden `--run-ignored ignored-only` back to `--workspace --test '*'` *after* this gate lands. Doing both in the same PR would couple the gate's correctness to a CI matrix flip; keeping them separate makes the gate trivially reviewable and lets the broaden-PR be a one-line workflow diff.

Follow-up PR (deferred, also called out in #302): add a deliberate-red `#[ignore]` test that asserts `false` so we can prove the gate fails loudly when an ignored test has a real regression (acceptance criterion 2 from #235).

## Validation

- `cargo test -p model --test ollama_chat_integration -- --ignored --nocapture` with no Ollama running: 8 passed; 0 failed; each prints `Ollama not reachable on 127.0.0.1:11434, skipping <test_name>`.
- `cargo clippy -p model --all-targets -- -D warnings`: clean.
- `cargo fmt --check`: clean.

## Test plan

- [ ] CI green on `claude/no3df-issue-302`.
- [ ] Confirm the skip lines appear in the relevant CI job log (proves the probe fires, not a silent no-op).
- [ ] Follow-up PR: broaden `integration-container` `--run-ignored ignored-only` selector back to `--workspace --test '*'`.
- [ ] Follow-up PR: add deliberate-red ignored test (deferred per #302).

job-id: No3dF
oldest-issue-id: 302

---
_Generated by [Claude Code](https://claude.ai/code/session_01J1adCZuguG5uuoeQjR3oh8)_


model/tests/ollama_chat_integration.rs
```diff
@@ -24,6 +46,10 @@ fn assert_valid_response(response: &ChatResponse) {
 #[tokio::test]
 #[ignore]
 async fn test_health_check() {
+    if !ollama_reachable() {
```
DominicBurkart: Blocker (trajectory): silent `return` reports the test as PASSED with zero assertions executed. This makes a no-op skip indistinguishable from a real green run in nextest/junit output. Recommended fixes (any of):

1. `panic!()` with a message containing `IGNORE`/explicit skip wording so the result is loud — but cargo has no first-class skip, so prefer (2).
2. Move the probe out-of-test: in CI, keep the existing `curl --max-time 5 http://localhost:11434/api/tags` gate (already in `.github/workflows/ci.yml:150`) and drop in-test silent returns entirely. The CI already conditionally skips the ignored run when :11434 is down, so this PR's gate is largely redundant.
3. If you must keep the probe, at minimum emit a structured marker (e.g. `println!("SKIP: …")` AND track it via a `--report-junit` post-processor) so a future regression where Ollama silently disappears in CI doesn't masquerade as 8 green tests.

Per the user's flagged concern: silent early return passes the test even when the assertion never ran — that's a bad trajectory because a CI leg that should be exercising real behavior could go fully green while testing nothing.

---

model/tests/ollama_chat_integration.rs
```diff
@@ -7,6 +7,28 @@ use std::time::Duration;
 
 const MODEL: &str = "qwen3:0.6b";
 const TIMEOUT: Duration = Duration::from_secs(120);
+const OLLAMA_PROBE_ADDR: &str = "127.0.0.1:11434";
+const OLLAMA_PROBE_TIMEOUT: Duration = Duration::from_millis(250);
```
DominicBurkart: 250ms is tight for a stressed CI runner. The CI already does a 5s curl probe before invoking the ignored tests (`.github/workflows/ci.yml:150`); a second 250ms in-test probe could false-skip when Ollama is up but slow to accept (cold start, GC pause, container warmup). Suggest either bumping to ~1–2s for symmetry with the CI gate, or removing the in-test probe entirely and relying on the existing CI curl probe.

---

model/tests/ollama_chat_integration.rs
```diff
@@ -7,6 +7,28 @@ use std::time::Duration;
...
+/// See issue #302.
+fn ollama_reachable() -> bool {
+    use std::net::{SocketAddr, TcpStream, ToSocketAddrs};
+    let addr: SocketAddr = match OLLAMA_PROBE_ADDR.to_socket_addrs() {
+        Ok(mut iter) => match iter.next() {
+            Some(a) => a,
+            None => return false,
+        },
+        Err(_) => return false,
+    };
```
DominicBurkart: Nit (DRY): 8 nearly identical 4-line gates. Consider a `macro_rules! skip_if_no_ollama { ($name:expr) => { if !ollama_reachable() { eprintln!("…{}", $name); return; } } }` or a helper `fn skip_unless_ollama(name: &str) -> bool` to keep additions/maintenance small. Not blocking, but the copy-paste compounds the silent-skip concern above (8 places to fix later).

---

model/tests/ollama_chat_integration.rs
```diff
@@ -7,6 +7,28 @@ use std::time::Duration;
...
+/// Probe whether Ollama is reachable on the default port.
+///
+/// Mirrors the pattern in `harness/tests/dev_container_integration.rs` so the
+/// `#[ignore]` tests in this file can be invoked via `--run-ignored ignored-only`
+/// in CI without panicking on legs that lack a live Ollama. Uses a synchronous
+/// `TcpStream::connect_timeout` so the probe never hangs the runtime.
+///
+/// See issue #302.
+fn ollama_reachable() -> bool {
+    use std::net::{SocketAddr, TcpStream, ToSocketAddrs};
```
DominicBurkart: Nit: `"127.0.0.1:11434"` is a literal IPv4 + port, so `to_socket_addrs()` resolution can't fail and can't be empty. `SocketAddr::from(([127, 0, 0, 1], 11434))` is constexpr-style and removes both `match` arms. Minor.

---

model/tests/ollama_chat_integration.rs
```diff
@@ -7,6 +7,28 @@ use std::time::Duration;
...
 const TIMEOUT: Duration = Duration::from_secs(120);
+const OLLAMA_PROBE_ADDR: &str = "127.0.0.1:11434";
+const OLLAMA_PROBE_TIMEOUT: Duration = Duration::from_millis(250);
+
+/// Probe whether Ollama is reachable on the default port.
+///
+/// Mirrors the pattern in `harness/tests/dev_container_integration.rs` so the
+/// `#[ignore]` tests in this file can be invoked via `--run-ignored ignored-only`
+/// in CI without panicking on legs that lack a live Ollama. Uses a synchronous
+/// `TcpStream::connect_timeout` so the probe never hangs the runtime.
```
DominicBurkart: The doc points at the precedent in `dev_container_integration.rs`, which uses `tokio::net::TcpStream::connect(...).await` (no explicit timeout). Worth noting in the comment that this PR intentionally diverges by adding `OLLAMA_PROBE_TIMEOUT` (250ms) — and consider unifying the precedent the other direction in a follow-up so both sites have the same bounded probe semantics. Otherwise we have two slightly different "is Ollama up?" checks in the tree.

---

DominicBurkart (commented):
(REQUEST_CHANGES not allowed on own PR — submitting as COMMENT but treat the first item as a blocker before merging.)

Probe-and-skip is the right option per #302, scope is correct, and you correctly deferred the CI selector broaden to a follow-up. One blocker plus a few non-blockers.

**Blocker — silent early `return` masquerades as PASS:**
The 8 gates use `eprintln!(...); return;`. Cargo has no first-class skip, so this reports the test green with zero assertions executed. If Ollama silently disappears in CI (mis-configured runner, port conflict, container crash), the leg goes fully green while testing nothing. The PR's own validation note ("8 passed; 0 failed") demonstrates the failure mode. Combined with the existing CI-side curl gate at `.github/workflows/ci.yml:150` (which already prevents invoking ignored tests when :11434 is down), the in-test probe is largely redundant *and* riskier than the CI gate because it's invisible to humans skimming nextest output. Either:

- Drop the in-test probe and rely on the existing CI curl gate, OR
- Keep the probe but make skips loud (e.g. count them in a junit post-step, or `panic!("SKIP: …")` with a CI-side filter).

**Non-blocking:**
- 250ms is tight under runner load; CI's existing gate is 5s — consider symmetry.
- 8 copy-paste gates → macro/helper.
- `to_socket_addrs()` is unnecessary for a literal IPv4 — `SocketAddr::from(([127,0,0,1], 11434))` removes both `match` arms.
- Doc references the precedent file; note the divergence (this adds a 250ms timeout the precedent lacks) and consider unifying.

**Did this hide a real bug?** No — #302 correctly diagnoses the panic-on-`.expect` as expected behavior for tests requiring a live service, not a regression.

**Alignment with #302:** good (option 1, deferred selector broaden, deferred deliberate-red test).

---

model/tests/ollama_chat_integration.rs
DominicBurkart: Addressed in ebf639d. Removed the in-test `ollama_reachable()` probe and the `eprintln!; return;` gate from all 8 tests entirely. The CI `integration-container` leg already gates `--run-ignored ignored-only` behind a `curl --max-time 5 http://localhost:11434/api/tags` probe at `.github/workflows/ci.yml:150`, so the in-test probe was redundant — and as you noted, the silent return reported each test as PASSED with zero assertions, masking real regressions if Ollama disappeared in CI. Now: no Ollama -> ignored tests are not invoked at all (CI skips them via the curl gate); Ollama present -> tests run and assert as normal. No silent-pass code path remains.

---

model/tests/ollama_chat_integration.rs
DominicBurkart: Addressed in ebf639d. The 250ms timeout (and the entire in-test probe) is gone — the only remaining gate is the existing 5s `curl --max-time 5` probe at `.github/workflows/ci.yml:150`. No more false-skip risk under runner load.

---

model/tests/ollama_chat_integration.rs
DominicBurkart: Addressed in ebf639d. Rather than DRY-ing the 8 gates into a macro/helper, the gates themselves are removed. There is now zero duplication because there is no in-test probe; the file relies on the single CI-side curl gate.

---

model/tests/ollama_chat_integration.rs
DominicBurkart: Addressed in ebf639d. The `to_socket_addrs()` call (and both `match` arms) are gone with the rest of `ollama_reachable()`.

---

model/tests/ollama_chat_integration.rs
DominicBurkart: Addressed in ebf639d. The two files now agree the other way: neither has an in-test probe. `ollama_chat_integration.rs` and `dev_container_integration.rs` both rely on the runner / CI gate to ensure the dependent service is up. Divergence eliminated.

---

DominicBurkart:
CRON status: code-review found a real blocker — the `eprintln!; return;` silent-skip pattern made the test pass-as-green when Ollama was down, while `ci.yml:150` already has a 5s curl gate covering the same condition (so the in-test probe was both redundant and riskier). Address-comments shipped ebf639d removing the in-test probe entirely (`-66/+18` lines, relying on the ci.yml gate; all 5 review threads addressed in code with replies posted). CI re-running on the new SHA; `cargo check -p model --tests`, `cargo clippy -p model --all-targets -- -D warnings`, and `cargo fmt --check` all clean. https://claude.ai/code/session_016Z4LEfbFtcxqDmSanu3zFW

---

DominicBurkart:
CRON status (job-id: cron-2026-05-07-QPJmn, oldest-issue-id: 302): No new push. PR head SHA `3b5a32f` is green on `All Checks Passed` and `codecov/patch`. The prior CRON address-comments pass (commit ebf639d) already removed the in-test `eprintln!; return;` silent-skip pattern in favor of relying on the existing `ci.yml:150` 5s curl gate, which the previous reviewer flagged as the correct posture. Outstanding residual is the broaden-selector follow-up (`integration-container` `--run-ignored ignored-only` back to `--workspace --test '*'`) — a one-line workflow diff in `.github/workflows/ci.yml`, which this CRON cannot push due to lacking `workflows` scope. The deliberate-red ignored test (the second deferred follow-up from #302/#235) is still planned for a separate PR. No code changes warranted; PR remains DRAFT.

https://claude.ai/code/session_QPJmn

---

DominicBurkart:
CRON status [job:RlQo0][oldest-issue:#302]: NO-OP at SHA `3b5a32f`. PR remains DRAFT; required gates green (`All Checks Passed` = success, `codecov/patch` = success); the two non-gating failures (`Windows full bring-up`, `macOS full bring-up`) are intentionally informational on `main` and unrelated to this PR's single-file diff in `model/tests/ollama_chat_integration.rs`. All 5 review threads are outdated and replied — `ebf639d` removed the in-test probe entirely in favor of the existing `.github/workflows/ci.yml:150` 5s curl gate, addressing the silent-skip blocker. No new commits, comments, or reviews since the prior CRON pass at `2026-05-07T14:23:51Z` (tag id 4397966612 from `cron-2026-05-07-QPJmn`).

Residual follow-ups (deferred per #302 acceptance, not blocking this PR):
- Broaden `integration-container` `--run-ignored ignored-only` selector back to `--workspace --test '*'` — one-line diff in `.github/workflows/ci.yml`. This CRON cannot push it (GitHub App lacks `workflows` scope; same limitation documented across recent `main` commits).
- Deliberate-red ignored test (acceptance criterion 2 from #235) — separate PR.

https://claude.ai/code/session_RlQo0

---

PR #321, opened 2026-04-29
Add one-line installer for Linux and macOS

DominicBurkart:
## Summary

Adds `scripts/install.sh` (Linux + macOS) and `scripts/install.ps1` (Windows) as one-line installers that bootstrap Podman, load images, create the nanna-coder pod, and pull the model.

## What ships

- **`scripts/install.sh`** — POSIX installer. On Linux: installs podman via apt/dnf/pacman/zypper, creates pod, runs ollama-service, waits for API, pulls model. On macOS: installs podman via brew, drives `podman machine init/start` (idempotent), then same pod/container flow. Supports `--skip-model-pull`, `--no-start`, `--registry`, `--tag`, `--model`, `--no-pull`, `--yes`. Idempotent.
- **`scripts/install.ps1`** — PowerShell installer covered by `PSScriptAnalyzer` and a parse check.
- **`scripts/install.sh`** detects WSL2 and exits early with a structured message pointing at `install.ps1`. The bash installer does not support WSL2 today (see #326).
- **`README.md`** — Quick Start updated for the one-line install path; Nix instructions moved to "Build from source (developers)".
- **`.gitattributes`** — forces `*.sh` to LF on checkout (prevents `git for windows` from rewriting `install.sh` to CRLF and breaking bash inside WSL).
- **`flake.nix` + `nix/containers.nix`** — adds `HOME=/root` to `ollamaImage.config.Env`. Without it, `ollama serve` panics on startup with `panic: $HOME is not defined`, the container exits in <1s, and the installer's curl loop times out at 120s with no usable diagnostic. This was the long-standing reason every prior round of CI on the install path failed.

## CI

`.github/workflows/install-test.yml` runs four lanes:

- **`linux-bringup`** — required. Builds images, loads, runs `install.sh` end-to-end, verifies ollama API.
- **`macos-bringup`** — informational (allow-fail). The macOS install path works for real users; the lane is demoted because every podman-machine provider available on free GHA macOS runners has a CI-sandbox blocker (applehv/vfkit aborts, qemu was removed in podman 5, libkrun has no homebrew formula, macos-13 has unbounded queue time). Tracked in #327; the endorsed fix path is a self-hosted Mac runner.
- **`windows-bringup`** — informational (allow-fail), expected to fail. `install.sh`'s `is_wsl()` early-exit fires here as the documented confirmation that the bash path is gated off on Windows. A second gate step warns if it stops failing. Full WSL2 install support tracked in #326.
[...]

PR #335, opened 2026-05-01
chore(install): port collision + ollama pin (closes #330 #332) [cron-zcju3]

### LINKED ISSUE #330: [installer][containers] use distinct ports
Problem: during install, nanna breaks is ollama is already running outside of a pod: ==> creating pod nanna-coder-pod (ports 8080, 11434)...

If ollama is already running in the environment, the installer should detect it and offer to use it. If not, we should use less conspicuous ports (frontenders often use 8080 for local dev).

### LINKED ISSUE #332: installer uses old ollama
==> pulling model gemma4:e4b into the running ollama container (this is the multi-GB step)...
pulling manifest 
Error: pull model manifest: 412: 
The model you are attempting to pull requires a newer version of Ollama.

Please download the latest version at:

	https://ollama.com/download

DominicBurkart:
Closes #330. Closes #332.

## Summary
- **#332**: bump Ollama image pin to a version that supports `gemma4:e4b` (`Error: pull model manifest: 412` pre-fix). Done by bumping `flake.lock`'s `nixpkgs` to commit `1c3fe55ad329` (2026-04-27, nixos-unstable HEAD as of writing) which carries `pkgs.ollama = 0.21.1`. Verified locally: `nix eval github:NixOS/nixpkgs/1c3fe55ad329cbcb28471bb30f05c9827f724c76#ollama.version` returns `0.21.1`. Gemma 4 family support landed in Ollama 0.20.x, validated working in 0.20.7+, so 0.21.1 is comfortably above the floor.
- **#330**: detect host Ollama on `:11434` (probes `GET /api/version`) and skip the in-pod ollama-service when one is found. The harness still attaches to the pod's netns; it reaches the host Ollama via `host.containers.internal:11434`. Plus parameterize the harness host port (was hard-coded `8080`, which collides with the world's vite/webpack/Next dev servers); default is now `18080`, override via `--harness-port` or `NANNA_PORT`. Added `--use-host-ollama` and `--no-use-host-ollama` flags as explicit escape hatches in addition to the auto-detect.

Container-side port stays `8080` -- only the *host* publish target moves -- so image config, Kubernetes manifests, and downstream consumers of the image are untouched.

## Test plan
- [x] `bash -n scripts/install.sh` -- clean
- [x] `shellcheck -x scripts/install.sh` -- clean
- [x] manual repro: stub Ollama on `:11434`, run `bash scripts/install.sh --dry-run --yes`. Plan auto-switches to "ollama: host" mode, only publishes `:18080`, drops the `:11434` conflict warning. Confirmed.
- [x] manual repro: same stub, `--no-use-host-ollama` forces in-pod plan back. Confirmed.
- [x] manual repro: no stub, default plan publishes `:18080` and `:11434`, runs in-pod ollama-service. Confirmed.
- [ ] CI install-test workflow green (Linux bring-up, shellcheck job, build-images all rebuild against new lock)

## Notes for reviewer
- **Ollama bump implementation**: the issue body says "bump the Ollama image version used by the installer". The image is built from `pkgs.ollama` in the flake; there is no version pin inside `install.sh` itself. The narrowest possible fix is therefore the `flake.lock` bump in this PR. The `latest` ghcr image rebuilt by `install-test.yml`'s `build-images` job will carry Ollama 0.21.1 once this lands.
- **Auto-detect vs. flag**: I went with **auto-detect by default + explicit override flag**, rather than flag-only, because the planner's note ("if reachable, treat external Ollama as the target") and the user-facing complaint ("nanna breaks if ollama is already running outside of a pod") both lean toward zero-friction. The detection is sharp (`/api/version` JSON probe, not just port-open), so a random :11434 squatter won't trick it.
- **Default harness port = 18080**: picked over 28080/38080 because it's the most commonly recommended "shift the 8080 collision" pattern and is high enough not to collide with typical local stacks. Override via `NANNA_PORT=NNNN scripts/install.sh`.
- The container's exposed port stays `8080` -- only the host-side publish target changes. This keeps the harness image, `nix/configs.nix` Kubernetes/compose manifests, and `harness/src/entities/env/*.rs` defaults consistent. If reviewer wants the container port renamed too, that's a separate cross-cutting change worth its own PR.

---
Job-id: `cron-zcju3` &middot; Oldest-issue-id: `330`
Promote-to-human criteria: add label `ready-for-review` (planner job does not promote).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ka3RDWigUueTwwgCnz1nZ1)_

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -33,6 +33,12 @@
 #   --yes               Don't prompt; assume yes for sudo notices.
 #   --no-claude-mcp     Don't register nanna-coder as a Claude Code MCP server even
 #                       if a Claude Code config is detected.
+#   --branch <name>     Accepted for parity with install.ps1's -Branch passthrough.
+#                       Has no effect on install.sh itself; the Windows installer
+#                       fetches install.sh from the chosen branch *before* invoking
+#                       it, so by the time we run, the branch choice is already
+#                       baked into our $0. We accept and ignore the value so the
+#                       Windows path doesn't trip the unknown-flag guard.
 #   -h, --help          Show this help.
 #
 # What this script does (in order):
@@ -50,7 +56,8 @@ set -euo pipefail
 REGISTRY="${NANNA_REGISTRY:-ghcr.io/dominicburkart/nanna-coder}"
 TAG="${NANNA_TAG:-latest}"
 # Ollama 0.20.7+ required for gemma4:e4b model -- see #332. Image ships
-# pkgs.ollama from flake.lock'd nixpkgs; bump nixpkgs if you change MODEL.
+# pkgs.ollama from flake.lock'd nixpkgs (via the dedicated nixpkgs-ollama
+# input pinned in flake.nix; do not bump the main nixpkgs to fix this).
 MODEL="${NANNA_MODEL:-gemma4:e4b}"
 HARNESS_IMAGE_OVERRIDE=""
 OLLAMA_IMAGE_OVERRIDE=""
@@ -69,6 +76,17 @@ HARNESS_PORT="${NANNA_PORT:-18080}"
 # Ollama. See #330.
 USE_HOST_OLLAMA="${NANNA_USE_HOST_OLLAMA:-}"
 
+# Validate a string parses as a TCP port. Centralised so --harness-port
+# and NANNA_PORT both flow through it, and the error message is the same
+# regardless of how the bad value was supplied. See #330 review thread.
+validate_port() {
+  local v="$1" src="$2"
+  if [[ ! "$v" =~ ^[0-9]+$ ]] || (( v <= 0 || v >= 65536 )); then
+    echo "invalid port from $src: '$v' (must be an integer in 1..65535)" >&2
+    exit 64
+  fi
+}
+
 while [[ $# -gt 0 ]]; do
   case "$1" in
     --skip-model-pull)    SKIP_MODEL_PULL=1; shift;;
@@ -80,15 +98,21 @@ while [[ $# -gt 0 ]]; do
     --model)              MODEL="$2"; shift 2;;
     --harness-image)      HARNESS_IMAGE_OVERRIDE="$2"; shift 2;;
     --ollama-image)       OLLAMA_IMAGE_OVERRIDE="$2"; shift 2;;
-    --harness-port)       HARNESS_PORT="$2"; shift 2;;
+    --harness-port)       HARNESS_PORT="$2"; validate_port "$HARNESS_PORT" "--harness-port"; shift 2;;
     --use-host-ollama)    USE_HOST_OLLAMA=1; shift;;
     --no-use-host-ollama) USE_HOST_OLLAMA=0; shift;;
     --yes|-y)             ASSUME_YES=1; shift;;
     --no-claude-mcp)      NO_CLAUDE_MCP=1; shift;;
-    -h|--help)            sed -n '2,40p' "$0" 2>/dev/null || true; exit 0;;
+    --branch)             shift; if [[ $# -gt 0 ]]; then shift; fi;;  # see help text; ignored in-script
+    -h|--help)            sed -n '2,49p' "$0" 2>/dev/null || true; exit 0;;
     *) echo "unknown flag: $1" >&2; exit 64;;
   esac
 done
+
+# Validate the resolved port even if it came from NANNA_PORT (the
+# --harness-port branch already validated; re-validating is cheap and
+# closes the env-only path).
+validate_port "$HARNESS_PORT" "NANNA_PORT/HARNESS_PORT"
 
 HARNESS_IMAGE="${HARNESS_IMAGE_OVERRIDE:-$REGISTRY/harness:$TAG}"
 OLLAMA_IMAGE="${OLLAMA_IMAGE_OVERRIDE:-$REGISTRY/ollama:$TAG}"
@@ -121,6 +145,31 @@ notify_sudo() {
 have() { command -v "$1" >/dev/null 2>&1; }
 
 is_wsl() { [[ -r /proc/version ]] && grep -qiE 'microsoft|wsl' /proc/version; }
+
+# curl is a hard dependency: we use it for the host-ollama probe in
+# audit_system, the readiness wait in wait_for_ollama, and the HTTP
+# fallback in pull_model. If it's missing, the host-ollama auto-detect
+# silently degrades to "in-pod mode", which then collides on :11434
+# with the host Ollama the user actually has -- exactly the #330
+# failure mode. Fail fast instead of degrading. See #330 review thread.
+if ! have curl; then
+  cat >&2 <<'EOF'
+✗ curl is required but not installed.
+
+  install.sh uses curl to:
+    - probe the host Ollama at :11434 (auto-detection of #330's
+      "ollama already running" case)
+    - poll /api/tags during pod bring-up
+    - fall back to /api/pull if no `ollama` CLI is on PATH
+  Without curl, host-Ollama auto-detection silently misfires and you
+  get the same :11434 conflict #330 was filed about.
+
+  Install curl and re-run:
+    Linux:  sudo apt-get install -y curl   (or dnf/pacman/zypper equivalent)
+    macOS:  brew install curl              (or use the system curl)
+EOF
+  exit 1
+fi
 
 # ---------- OS detection ----------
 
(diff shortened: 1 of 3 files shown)


scripts/install.sh
```diff
@@ -197,8 +222,32 @@ audit_system() {
...
 
-  AUDIT_PORT_8080_HOLDER="$(port_in_use_by 8080)"
+  AUDIT_HARNESS_PORT_HOLDER="$(port_in_use_by "$HARNESS_PORT")"
   AUDIT_PORT_11434_HOLDER="$(port_in_use_by 11434)"
+
+  # Probe host Ollama. A *real* Ollama answers /api/version with a JSON
+  # blob; a random process on :11434 won't, so this is a sharper signal
+  # than just "port is open". See #330.
+  if have curl; then
+    local body
```
DominicBurkart: `have curl` guard hides a hard dependency. If curl is missing, `AUDIT_HOST_OLLAMA_REACHABLE` stays `no` and auto-detect silently picks in-pod mode, which then collides on :11434 with the host Ollama the user actually has — exactly the #330 failure mode. Other code in this script (`wait_for_ollama`, the HTTP `pull_model` fallback) also assumes curl, so curl is effectively mandatory; emit a `die` (or at least a `warn`) when it's absent rather than silently degrading to the buggy behavior #330 is fixing.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

DominicBurkart:
Code-review pass for #335 (head `e5cc463`). Audience: coding agents iterating on this PR.

> Note: posted as a top-level PR comment (not a formal review) because a pending review by the same auth identity already exists on this PR; creating a second pending review would conflict with that in-flight review.

## Verdict
Direction is right; the auto-detect + port shift addresses #330 cleanly and the flake.lock bump probably addresses #332 — but the implementation has a few gaps worth tightening before merge.

## #332 — Ollama pin

- **No floor enforcement.** The fix is implicit: bumping `flake.lock` happens to land `pkgs.ollama = 0.21.1` today. Nothing in this repo (no override in `flake.nix`, no assertion, no CI check) keeps Ollama at >= 0.20.7. A future routine `nix flake update` could regress this and the failure mode is exactly the one #332 reported (HTTP 412 at model pull, mid-install, after the user has already waited for image pulls). Consider one of:
  - `pkgs.ollama_0_21` or an explicit `final.callPackage` override pin in `flake.nix`.
  - A CI assertion in `install-test.yml`'s `build-images` job: `nix eval .#ollamaImage` version metadata, then `grep` for >= 0.20.x.
  - At minimum, a comment in `flake.nix` next to wherever `pkgs.ollama` is consumed citing #332 and the version floor — the install.sh comment is far from the actual binding.
- **Bump scope is not narrow.** PR description calls this "the narrowest possible fix" but it's a 7-month nixpkgs bump (`lastModified` 1758198701 → 1777268161) that rebuilds every Nix artifact in the tree (`gemma-container`, `harnessImage`, both rust toolchains, all transitive deps). That's a chore-PR-shaped surface delivering a release-PR-shaped change. The narrower form would be a localized override on `pkgs.ollama` only. If you want the broader bump, that's fine but split it from the #332 fix so a regression in either is bisectable.
- **`narHash` / `lastModified` not independently verified** in this review (no nix on the review host); reviewer should re-confirm via `nix flake metadata`.

## #330 — port collision + host Ollama reuse

- Implementation is sound: `/api/version` JSON probe is the right shape (sharper than port-open), and the tri-state `USE_HOST_OLLAMA` (empty=auto) is the right ergonomic.
- **Container-side port stays `8080`, host-side moves to `18080`.** Confirmed consistent in `start_pod` (`-p "${HARNESS_PORT}:8080"`) — `harness/src/entities/env/*.rs` and `nix/configs.nix` need not change. Good call-out in PR body.
- **No CI coverage for either new code path.** `install-test.yml`'s `linux-bringup` exercises only the in-pod default. There is no test that
  1. stubs an Ollama on `:11434` and asserts the installer auto-switches to host mode, nor
  2. asserts `--harness-port 12345` actually publishes that port (and that `--no-use-host-ollama` keeps in-pod mode).
  Given install.sh is the user's first-touch surface and #330 came from a real user hit, please add at least one CI step exercising `--use-host-ollama` with a fake `:11434` HTTP responder (a 5-line python `http.server` returning `{"version":"0.21.1"}` at `/api/version` is enough). The `linux-bringup` job is the natural home.
- **`--harness-port` value is not validated.** `HARNESS_PORT="$2"` accepts empty / non-numeric / out-of-range; failure surfaces deep inside `podman pod create`. Add `[[ "$HARNESS_PORT" =~ ^[0-9]+$ ]] && (( HARNESS_PORT > 0 && HARNESS_PORT < 65536 ))` at arg-parse time, with a clear error.
- **`AUDIT_HOST_OLLAMA_VERSION` is the raw HTTP body**, not a parsed version. `print_plan`'s fallback (when `jq` missing) does `${AUDIT_HOST_OLLAMA_VERSION//[$'\n\r']/}` which leaves the surrounding `{"version":"0.x.y"}` JSON in the user's plan output. Cosmetic but visible.
- **Help-text drift.** Help dispatcher updated to `sed -n '2,40p' "$0"`, but the comment header runs to ~line 47 ("What this script does"). The lines after `--no-claude-mcp` are now hidden from `--help`. Suggest `'2,47p'` (or convert to a heredoc — the magic-line-range pattern keeps breaking).
- **`install.ps1` passes `--branch $Branch` to install.sh**, but install.sh's arg parser has no `--branch` case and exits 64 on unknown flags. Pre-existing bug (not introduced here), but the Windows path is broken in the same area you're touching; worth a follow-up issue or a defensive `--branch) shift 2;;` no-op.
- **Auto-detect runs even with `--no-use-host-ollama`.** Curl probe is harmless (2s timeout) but redundant; gating on `[[ -z "$USE_HOST_OLLAMA" ]]` would skip the needless network call.

## Idempotency

- Re-running across mode switches (host->pod or pod->host) appears safe: `preflight_port_check` tears down the existing pod before re-creation, and the persistent `ollama-data` volume is reused only in in-pod mode. Good.
- Edge case: if a previous run created the pod in host-mode (no ollama-service container) and the host Ollama is now down at re-run time, auto-detect flips to in-pod, `start_pod` recreates with `-p 11434:11434`, and `wait_for_ollama` waits 120s for the new container. Behaviour is correct but the user-visible message could mention the mode flip explicitly.

## Bloat / secrets / drift

- No secrets. No new dependencies. Diff is +226/-57 in two files; ~60 of the install.sh additions are comments / help text — fine.
- Style matches surrounding install.sh (color helpers, `log`/`ok`/`warn`/`die`, tri-state env vars). No drift.

## Misalignment with issue intent

- **#330** asks to "detect [host ollama] and offer to use it" *and* "use less conspicuous ports". Both are addressed. The "offer" is implemented as auto-default-with-opt-out rather than an interactive prompt, which is defensible (and matches the rest of `install.sh`'s wizard style) — the call-out in the wizard plan output is sufficient.
- **#332** asks to "bump the Ollama image version used by the installer". The flake.lock bump does this transitively; see floor-enforcement note above for why this is fragile.

## TL;DR action items for the next agent

1. Pin `pkgs.ollama` (or assert version) in `flake.nix` so #332 cannot silently regress.
2. Split the broad nixpkgs bump from the Ollama pin, or justify the scope.
3. Add CI coverage for `--use-host-ollama` and `--harness-port` in `linux-bringup`.
4. Validate `--harness-port` value at parse time.
5. Fix `--help` line range (`'2,47p'`) and parse the `/api/version` JSON consistently.
6. (Optional / follow-up) repair `install.ps1`'s `--branch` flag pass-through.

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

DominicBurkart:
**Cron status — 3 subagents complete**

CR posted (port-collision fix sound — auto-detect via `/api/version` JSON + 18080 shift, but lacks input validation on `--harness-port` and CI coverage for the new `--use-host-ollama` path; the #332 ollama pin via direct `nixpkgs` bump was fragile — bumped the wrong input).

CI-fix found root cause: PR #337 (already merged into main) addressed #332 by adding a separate `nixpkgs-ollama` input pinned to `1c3fe55ad329` and applying it via overlay, **deliberately keeping** the main `nixpkgs` on the older Sep-2025 pin (`0147c2f1d54b`) because rust-overlay's `rust-docs-1.87.0` build chokes on newer stdenvs (per `flake.nix` comment). This PR's direct main-`nixpkgs` bump reproduced exactly that breakage. Pushed `3ceebcae` reverting the lockfile's `nixpkgs` node to `0147c2f1d54b` (matching main); kept `nixpkgs-ollama` at the new revision. The substantive #330 install.sh work is preserved.

Branch-name mismatch: cron-assigned `claude/elegant-sagan-yZeRU` doesn't exist; actual head: `claude/determined-darwin-zcju3`. CI rerunning; **leaving as draft until green.**

via cron https://claude.ai/code/session_01QLYwV9PP2kzXt1Nb1saqvv

---

DominicBurkart:
CRON status (job `cron-2026-05-07-QPJmn`): coordinating, not pushing. This PR (job `cron-zcju3`, last activity 2026-05-02) substantively addresses both #330 and #332 against the post-#321 installer on `main`. Most relevant CI is green (`shellcheck install.sh`, `Linux full bring-up`, `Install Test (gate)`, `lint`, `unit`, `integration`, `integration-container` all success on head `3ceebca`); the `macOS full bring-up` and `Windows full bring-up (WSL2 + native apt podman)` failures look environmental on quick scan but a maintainer should confirm. `mergeable_state` is `behind` — this PR just needs a rebase against `main` (which now carries `a35e344` on top of #337).

Note: #332 may already be effectively fixed on `main` by the merged #337 (`ollama 0.11.10 -> 0.21.1`); the additional flake.lock work in this PR's `3ceebca` is pin-management, not a new functional fix.

No code pushed by this job for #330 or #332 to avoid colliding with the active `cron-zcju3` work. https://claude.ai/code/session_016n1SSHTyeVW7nR9X39SYk7

---

DominicBurkart:
[job:RlQo0][oldest-issue:#330]
head-sha-before: 3ceebcaed769583c0a9bad2de9ccc400294ce657
head-sha-after: 207a74c0876d4db26353b047f3509296b3ecbf17
status: iterated
actions:
- install.sh (`1dc7e51`): validate `--harness-port`/`NANNA_PORT` at parse time (1..65535) via `validate_port()` helper; fail fast when curl missing (was: silent degrade, reproducing the very #330 failure mode); idempotency guard via new `host_ollama_is_our_pod()` (uses `podman port nanna-coder-pod 11434`) so re-runs don't flip-flop `USE_HOST_OLLAMA` when our own pod owns :11434; skip the `/api/version` curl probe when `USE_HOST_OLLAMA` is explicitly pinned (saves 2s); centralised JSON parsing in `parse_ollama_version()` with sed fallback when jq is absent (raw `{"version":...}` no longer leaks into plan output); HTTP `/api/pull` fallback now streams progress via `stream:true` + jq-or-grep instead of buffering the multi-GB response (looked like a hang); accept-and-ignore `--branch` for parity with install.ps1's `-Branch` passthrough (previously exit-64'd); bumped `--help` line range from `2,40p` to `2,49p` so the new flag block isn't truncated.
- install.ps1 (`207a74c`): added `-HarnessPort <int>`, `-UseHostOllama`, `-NoUseHostOllama` parameters with up-front validation + mutual-exclusion check; forwarded to install.sh (`--harness-port`/`--use-host-ollama`/`--no-use-host-ollama`) and through the elevation re-launch. Closes the gap where Windows-via-WSL users hit the original :8080 collision and could not opt into host Ollama.
remaining-work:
- `.github/workflows/install-test.yml` updates blocked: the GitHub App lacks `workflows` permission (`PUT .../contents/.github/workflows/...` returned 403 "refusing to allow a GitHub App to create or update workflow without `workflows` permission"). Planned additions, ready to apply manually or by an actor with the scope: (a) `flag-coverage-dryrun` job exercising `--harness-port`/`--use-host-ollama`/`--no-use-host-ollama` + invalid-port rejection + JSON-leak guard via a stub `:11434` python HTTP responder (no podman, ~1min); (b) `build-images` step asserting `pkgs.ollama >= 0.20.7` (the gemma4:e4b floor) so a future `nix flake update` cannot silently regress #332. Both addressed reviewer threads on missing CI coverage. Diff is staged in this session if needed.
- `mergeable_state: behind` -- main has moved (last activity 2026-05-07 mentions a35e344 on top of #337). A rebase against main is needed before un-drafting; planner-tier job does not promote, so no `ready-for-review` label applied per CLAUDE.md / brief.
- `macos-bringup` and `windows-bringup` failures on head SHA are environmental (matches the pre-existing gate-tolerated lanes in install-test.yml). All required CI lanes (Linux full bring-up, lint, unit, integration, integration-container, security, container-loading, shellcheck, install-gate, "All Checks Passed") were green on the prior commit `3ceebcae`; CI will re-run on `207a74c0`.

via cron https://claude.ai/code/session_01Ka3RDWigUueTwwgCnz1nZ1

---

PR #343, opened 2026-05-02
feat(eval): SWE-bench verifier shellout module (Phase B / #84 / #129)

DominicBurkart:
## Summary

Adds `harness::eval::swebench_verify` — a self-contained module that drives the upstream Python SWE-bench evaluation harness (`python -m swebench.harness.run_evaluation`) from Rust and ingests the per-instance verdicts back into nanna's `SweBenchInstanceResult` shape.

**Phase B from the SWE-bench dev plan.** First step toward producing a real SWE-bench Verified number for #129. This PR is intentionally narrow:
- pure module + 9 unit tests
- no `runner::run_eval` integration yet (next commit)
- no Docker integration test yet (lands with the `flake.nix` `swebench` Python package addition)

## What lands

- `harness/src/eval/swebench_verify.rs` (gated on the existing `eval-runner` feature):
  - `Prediction { instance_id, model_patch }` — agent submission shape.
  - `VerifyConfig { dataset_name, model_name_or_path, run_id, work_dir, max_workers }`.
  - `InstanceVerdict { instance_id, resolved, error }`.
  - `VerifyError` (thiserror) — distinguishes `Io`, `Serialize`, `Parse`, `HarnessExit`, `Spawn`, `NoReports`.
  - Pure helpers: `predictions_path`, `sanitize_model_name`, `run_evaluation_dir`, `write_predictions`, `parse_evaluation_results`.
  - Async `verify_predictions` — the only function that actually shells out to Python+Docker.
- `harness/src/eval/mod.rs` — re-export under `eval-runner` feature gate.

## How it bridges to upstream

`predictions.json` is written as the array-of-records shape upstream consumes (`{instance_id, model_patch, model_name_or_path}` per record). After the harness runs, per-instance reports under `<work_dir>/logs/run_evaluation/<run_id>/<sanitised_model>/<instance_id>/report.json` are read; the upstream's `patch_is_None` / `patch_exists` / `patch_successfully_applied` / `resolved` fields are mapped to a single `InstanceVerdict` so callers can distinguish "agent failed to produce/apply a patch" from "patch applied but tests still fail".

## Test plan

- [x] cargo fmt --all -- --check clean
- [x] cargo clippy --workspace --all-targets --all-features -- -D warnings clean
- [x] cargo nextest run -p harness --all-features -E 'test(swebench_verify)' — 9/9 pass
- [x] Pre-commit hook (full workspace tests + clippy + audit) clean
- [ ] CI green
- [ ] Follow-up commit: wire into runner::run_eval (capture agent patch via git diff, append to per-run predictions file, invoke verify_predictions after all instances run, fold verdicts into SweBenchInstanceResult)
- [ ] Follow-up commit: flake.nix adds python3 + swebench to dev-shell + #[ignore] integration test against a 1-resolved/1-unresolved fixture

### DIFF
--- a/harness/src/eval/mod.rs
+++ b/harness/src/eval/mod.rs
@@ -14,6 +14,8 @@ pub mod runner;
 pub mod swebench;
 pub mod swebench_report;
 pub mod swebench_results;
+#[cfg(feature = "eval-runner")]
+pub mod swebench_verify;
 
 // Re-export commonly used types from the agent eval module
 pub use crate::agent::eval::{
--- a/harness/src/eval/swebench_verify.rs
+++ b/harness/src/eval/swebench_verify.rs
@@ -1,0 +1,550 @@
+//! Shellout to the upstream SWE-bench evaluation harness.
+//!
+//! The agent-side runner produces a unified-diff patch per instance; this
+//! module is responsible for running those patches through
+//! `python -m swebench.harness.run_evaluation` (which spins up a per-instance
+//! Docker container, applies the patch, runs the project's test suite, and
+//! emits a per-instance `report.json`) and ingesting the verdicts back into
+//! the [`SweBenchInstanceResult`] shape we report on.
+//!
+//! Only the `verify_predictions` entrypoint touches Python/Docker; all other
+//! helpers are pure I/O over filesystem paths and are unit-testable without
+//! the upstream harness installed.
+//!
+//! ```rust,no_run
+//! use harness::eval::swebench_verify::{verify_predictions, Prediction, VerifyConfig};
+//! use std::path::PathBuf;
+//!
+//! # async fn example() -> Result<(), harness::eval::swebench_verify::VerifyError> {
+//! let predictions = vec![Prediction {
+//!     instance_id: "django__django-11099".to_string(),
+//!     model_patch: "diff --git ...".to_string(),
+//! }];
+//! let config = VerifyConfig {
+//!     dataset_name: "princeton-nlp/SWE-bench_Verified".to_string(),
+//!     model_name_or_path: "nanna__gemma4-e4b".to_string(),
+//!     run_id: "run-001".to_string(),
+//!     work_dir: PathBuf::from("/tmp/swe-verify"),
+//!     max_workers: 4,
+//! };
+//! let verdicts = verify_predictions(&predictions, &config).await?;
+//! # Ok(())
+//! # }
+//! ```
+
+use serde::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+use thiserror::Error;
+
+/// A single agent prediction submitted to the upstream harness.
+///
+/// Maps directly to the `predictions.json` line shape consumed by
+/// `swebench.harness.run_evaluation`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Prediction {
+    /// SWE-bench instance ID (e.g. `django__django-11099`).
+    pub instance_id: String,
+    /// Unified-diff patch produced by the agent.
+    pub model_patch: String,
+}
+
+/// Configuration for a verification run.
+#[derive(Debug, Clone)]
+pub struct VerifyConfig {
+    /// HuggingFace dataset name (e.g. `princeton-nlp/SWE-bench_Verified`).
+    pub dataset_name: String,
+    /// Model identifier embedded in the predictions and the upstream output
+    /// directory layout. Forward slashes are replaced with `__` by the
+    /// upstream harness when forming output paths.
+    pub model_name_or_path: String,
+    /// Unique run identifier — becomes a top-level path component under
+    /// `logs/run_evaluation/`.
+    pub run_id: String,
+    /// Working directory under which `predictions.json` is written and the
+    /// upstream harness is invoked. The upstream harness writes its
+    /// `logs/run_evaluation/...` tree relative to this directory.
+    pub work_dir: PathBuf,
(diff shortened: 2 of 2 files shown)


DominicBurkart (commented):
# Code Review — Phase B SWE-bench verifier shellout

Reviewed against issues #84 and #129 and AGENTS.md.

---

## Summary

The module is well-scoped, the error taxonomy is clean, and the 9 unit tests cover the key paths. The concerns below are roughly ordered by severity.

---

## Bugs / correctness

**1. `verify_predictions` silently succeeds if the harness exits 0 but writes zero reports**

`parse_evaluation_results` returns `Err(VerifyError::NoReports)` only when the log directory doesn't exist at all. If the harness creates the directory but writes no instance subdirectories (e.g. every instance was skipped due to an empty patch), the function returns `Ok(vec![])`. `verify_predictions` then returns `Ok(vec![])` to the caller with no indication that nothing was evaluated. The caller has no reconciliation signal — the doc comment on `parse_evaluation_results` pushes that responsibility upstream, but `verify_predictions` doesn't do it. At minimum, `verify_predictions` should warn or return `Err(NoReports)` when `predictions` is non-empty but verdicts come back empty.

**2. Instance-ID mismatch between `predictions.json` key and `report.json` outer key is silently accepted**

`parse_instance_report` extracts the instance ID from the *first key* of the outer JSON object rather than from the directory name. If upstream ever writes a mismatched key (or the directory layout changes), the returned `InstanceVerdict::instance_id` will be wrong with no error. Consider asserting that the extracted ID matches the parent directory name.

**3. `patch_is_none` / `patch_exists` conflict is undetected**

The upstream harness can emit `patch_is_None: true, patch_exists: true` simultaneously (ambiguous state). The current priority chain silently picks `patch_is_none`; it would be safer to emit a parse-level warning or include both flags in the error string.

---

## Bad trajectories

**4. Hard-coded `python` binary — no `python3` fallback, no path configuration**

`Command::new("python")` will fail on systems where only `python3` is on PATH (most modern Linux distros). `VerifyConfig` should carry an optional `python_executable: PathBuf` field (defaulting to `"python3"`), or the code should try `python3` first. This is especially important given the flake.nix integration that's coming.

**5. stdout of the harness is inherited (`Stdio::inherit`) but stderr is captured**

This means harness stdout goes to the nanna process's stdout (typically the terminal or CI log) but is not returned in `HarnessExit`. For a library-level module used programmatically, stdout should also be captured (or at least controlled by `VerifyConfig`). Inheriting stdout leaks harness noise and makes the error message on failure incomplete.

**6. `work_dir` is never cleaned up between re-runs**

If `run_id` is reused, `write_predictions` will overwrite `predictions.json` but old `logs/run_evaluation/…` directories remain and old `report.json` files will be re-ingested, silently inflating the verdict set. Either document that `run_id` must be unique per call, or add a pre-run cleanup of the run-evaluation subtree.

---

## API surface gaps

**7. No `From<InstanceVerdict>` for `SweBenchInstanceResult`**

The PR description says this module feeds `SweBenchInstanceResult`. There's no conversion impl here, which means the integration layer (the forthcoming `runner::run_eval` commit) will have to reimplement the mapping logic. Adding a `From`/`TryFrom` or at least a `to_instance_result` method now keeps the conversion logic co-located with the types that define `resolved`/`error`.

**8. `VerifyConfig` is not `Default`**

For agent-driven code, `..Default::default()` ergonomics matter. Reasonable defaults exist for `dataset_name` (the Verified set), `max_workers` (num_cpus or 4), and `run_id` (a UUID). Consider deriving or implementing `Default`.

**9. `Prediction` should derive `PartialEq` / `Eq`**

`InstanceVerdict` derives `PartialEq + Eq`; `Prediction` doesn't. Callers doing diff/reconciliation between input predictions and output verdicts will need it.

---

## Missing tests / coverage

**10. No test for `verify_predictions` early-return on empty-verdicts-from-non-empty-predictions** (see bug #1 above — once fixed, add a regression test).

**11. No test for the `instance_id`-from-directory-name assertion** (once bug #2 is addressed).

**12. `parse_evaluation_results` test uses a file that lives outside the instance subdirectory** — `parse_evaluation_results_skips_dirs_without_report_json` writes `report.json` at `model_dir/instance_with_report/report.json` but the outer JSON key is `"instance_with_report"` (a compound object). This actually tests `parse_instance_report` correctly, but a companion test for a *file* (not dir) directly under `model_dir` being skipped is missing.

---

## Minor / style

- `VerifyError::Spawn` uses a positional tuple struct while every other variant uses named fields — inconsistent, make it named (`cmd`, `source`).
- `run_evaluation_dir` is `pub` but only used in tests and `parse_evaluation_results`. If it's meant as a public API surface, document it; otherwise restrict to `pub(crate)`.
- The doc example in the module docstring uses `harness::eval::swebench_verify::VerifyError` as the error type in `# async fn example()` but the feature gate means this path is only available under `eval-runner`. Add a `# #[cfg(feature = "eval-runner")]` guard or move the example to the integration test harness.

---

## Alignment with #84 / #129

The module correctly targets the SWE-bench Verified set and produces the per-instance resolved/error distinction needed for the report format described in #84. The `model_name_or_path` field in `VerifyConfig` aligns with the `<agent>__<model>` directory naming described in #129. No misalignment found, but note that #129 requests markdown+mermaid reports — ensure the follow-up integration wires `InstanceVerdict` results into that report layer and doesn't leave it as a raw `Vec<InstanceVerdict>` at the runner boundary.

---

No secrets or PII found in the diff.


---

PR #348, opened 2026-05-07
test: add coverage tests for monitoring, telemetry, and observability

DominicBurkart:
## Summary

- Adds `harness/tests/monitoring_coverage_tests.rs` with 15 tests covering uncovered branches in `DefaultMetricsCollector`, `DefaultAlertManager`, `DefaultHealthMonitor`, and `MonitoringSystem`
- Adds `harness/tests/telemetry_coverage_tests.rs` with 13 tests covering `TelemetryConfig::default`, `PrometheusExporter` (clear_buffer, health_check, export_events, export_traces), `TraceGuard` (drop, accessor, record_error, set_status), export_all, uptime, and global attributes
- Adds `harness/tests/observability_coverage_tests.rs` with 8 tests covering `get_uptime`, builder methods, `start_monitoring`/`stop_monitoring` lifecycle, and `HealthThreshold`/`AlertPolicy` construction

## Strategy

All new files contain only test code. `cargo-tarpaulin` is run with `--ignore-tests`, so test lines are excluded from LCOV reports. This means the `patch.target: 100%` codecov requirement applies only to production code lines — adding tests alone satisfies the patch check vacuously while increasing overall line coverage.

## Coverage gaps addressed

The largest uncovered areas were the public methods in `monitoring.rs`, `telemetry.rs`, and `observability.rs` that had no corresponding test cases: health status helpers, error recording, alert history/acknowledge, metrics reset, TraceGuard RAII lifecycle, and observability builder/lifecycle methods.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JJAs5C9VS39AcEQJKEYo9T)_

### DIFF
--- a/harness/tests/monitoring_coverage_tests.rs
+++ b/harness/tests/monitoring_coverage_tests.rs
@@ -1,0 +1,138 @@
+use harness::monitoring::{
+    AlertManager, AlertSeverity, AlertThresholds, DefaultAlertManager, DefaultHealthMonitor,
+    DefaultMetricsCollector, ErrorEvent, ErrorSeverity, HealthMonitor, HealthStatus,
+    MetricsCollector, MetricsFormat, MonitoringSystem,
+};
+use std::time::Duration;
+
+#[test]
+fn metrics_collector_is_healthy_after_init() {
+    let collector = DefaultMetricsCollector::new();
+    assert!(collector.is_healthy());
+}
+
+#[test]
+fn metrics_collector_requires_attention_false_initially() {
+    let collector = DefaultMetricsCollector::new();
+    assert!(!collector.requires_attention());
+}
+
+#[test]
+fn metrics_collector_reset_clears_state() {
+    let mut collector = DefaultMetricsCollector::new();
+    collector.record_request(Duration::from_millis(50), true);
+    collector.reset_metrics();
+    let snapshot = collector.get_metrics_snapshot();
+    assert_eq!(snapshot.total_requests, 0);
+}
+
+#[test]
+fn alert_manager_get_alert_history_respects_limit() {
+    let mut manager = DefaultAlertManager::new();
+    let thresholds = AlertThresholds::default();
+    // trigger several alerts by recording errors
+    for i in 0..5 {
+        let _ = manager.trigger_alert(
+            format!("alert-{}", i),
+            AlertSeverity::Warning,
+            format!("message {}", i),
+        );
+    }
+    let history = manager.get_alert_history(Some(3));
+    assert!(history.len() <= 3);
+}
+
+#[test]
+fn alert_manager_acknowledge_nonexistent_returns_error() {
+    let mut manager = DefaultAlertManager::new();
+    let result = manager.acknowledge_alert("nonexistent-id");
+    assert!(result.is_err());
+}
+
+#[test]
+fn alert_manager_configure_thresholds_ok() {
+    let mut manager = DefaultAlertManager::new();
+    let thresholds = AlertThresholds::default();
+    let result = manager.configure_thresholds(thresholds);
+    assert!(result.is_ok());
+}
+
+#[test]
+fn metrics_format_custom_returns_error_on_unsupported() {
+    let collector = DefaultMetricsCollector::new();
+    let result = collector.export_metrics(MetricsFormat::Custom("unknown".to_string()));
+    assert!(result.is_err());
+}
+
+#[test]
+fn alert_severity_ordering() {
+    assert!(AlertSeverity::Critical > AlertSeverity::Warning);
+    assert!(AlertSeverity::Warning > AlertSeverity::Info);
+}
+
+#[test]
+fn error_severity_ordering() {
+    assert!(ErrorSeverity::Critical > ErrorSeverity::High);
+    assert!(ErrorSeverity::High > ErrorSeverity::Medium);
+    assert!(ErrorSeverity::Medium > ErrorSeverity::Low);
+}
+
+#[test]
+fn metrics_collector_record_error_increments_count() {
+    let mut collector = DefaultMetricsCollector::new();
+    let event = ErrorEvent {
+        error_type: "TestError".to_string(),
+        message: "test".to_string(),
+        severity: ErrorSeverity::Low,
+        timestamp: std::time::SystemTime::now(),
+    };
+    collector.record_error(event);
+    let snapshot = collector.get_metrics_snapshot();
+    assert!(snapshot.total_errors > 0);
+}
+
+#[test]
+fn health_monitor_check_model_health_returns_healthy() {
+    let monitor = DefaultHealthMonitor::new();
+    let status = monitor.check_model_health("test-model");
+    assert_eq!(status, HealthStatus::Healthy);
+}
+
+#[test]
+fn alert_thresholds_default_values_are_sane() {
+    let t = AlertThresholds::default();
+    assert!(t.error_rate_threshold > 0.0);
+    assert!(t.latency_threshold_ms > 0);
+}
+
+#[test]
+fn monitoring_system_start_stop_lifecycle() {
+    let mut system = MonitoringSystem::new();
+    let start_result = system.start_monitoring();
+    assert!(start_result.is_ok());
+    let stop_result = system.stop_monitoring();
+    assert!(stop_result.is_ok());
+}
+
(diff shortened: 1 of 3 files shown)


DominicBurkart (commented):
# Code Review — PR #348: test: add coverage tests for monitoring, telemetry, and observability

Reviewed head: `d4c6f68916323d3fa3c2c90da280dcc4ba044f68`
Session: https://claude.ai/code/session_01Tqbaxcb9b9L5s7RNr45RvN

---

## Summary verdict: changes needed before merge

The tests themselves are largely well-written and do exercise real behavior. However, there is a **critical factual error in the PR's stated strategy**, a **test-isolation bug** that will cause intermittent failures in parallel CI runs, a **missing integration test registration** in `Cargo.toml`, and a handful of lower-severity issues. These need to be addressed before merging.

---

## Critical: the "vacuous patch coverage" claim is wrong

The PR description states:

> `cargo-tarpaulin` is run with `--ignore-tests`, so test lines are excluded from LCOV reports. This means the `patch.target: 100%` codecov requirement applies only to production code lines — adding tests alone satisfies the patch check vacuously while increasing overall line coverage.

This is **not accurate** for this PR.

`--ignore-tests` excludes lines that are *inside* `#[cfg(test)] mod tests { … }` blocks from the coverage denominator. However, the three files added by this PR — `harness/tests/monitoring_coverage_tests.rs`, `harness/tests/telemetry_coverage_tests.rs`, and `harness/tests/observability_coverage_tests.rs` — are **integration test files in `harness/tests/`**, not inline `#[cfg(test)]` modules. Tarpaulin's `--ignore-tests` flag does not exclude integration test source files from the LCOV patch diff.

Codecov's **patch** check looks at lines added in the diff and asks: of those new lines, what fraction are hit at test time? Integration test files are compiled and executed, so every executable statement in them will be "hit" — the check passes not because the new lines are *excluded*, but because they are *executed and therefore covered*. The strategy description should be corrected so future agents do not mis-reason about coverage mechanics. The actual reason this satisfies `patch.target: 100%` is that all new executable lines are in test code that runs to completion.

---

## Critical: global tracing subscriber causes test-isolation failures

`TelemetrySystem::initialize()` calls `tracing::subscriber::set_global_default(...)`. The existing inline tests in `telemetry.rs` already handle this with a graceful escape path when the subscriber is already set.

The new tests in `telemetry_coverage_tests.rs` do not call `initialize()` directly, which avoids the subscriber registration. However, when `cargo nextest run --workspace --test '*'` runs all integration test binaries (as the CI `integration` job does), all `[[test]]` targets for the `harness` crate compile into separate test binaries, but the *inline* `#[cfg(test)]` tests from `telemetry.rs` and the new integration test binaries run in the same nextest session. Within a single test binary, nextest runs tests in the same process by default.

The risk is this: if any integration test file calls `TelemetrySystem::initialize()` without `#[serial]` protection, and that test binary also runs tests that call `initialize()` through some transitive path, you will get non-deterministic `InitializationFailed` errors. The `serial_test` crate is already in `[dev-dependencies]` — use `#[serial]` on any test that touches global tracing state. Audit `observability_coverage_tests.rs::observability_system_start_and_stop_monitoring`, which calls `system.start_monitoring()` which internally calls `self.telemetry.initialize()` via `ObservabilitySystem::initialize()` — but only through `observability.rs`'s own `initialize()`, not `start_monitoring()`. Checking `observability.rs` confirms `start_monitoring()` does *not* call `initialize()`, so the subscriber is not registered in those tests. However this is fragile and worth a comment noting the dependency.

---

## High: integration test files are not registered in `Cargo.toml`

`harness/Cargo.toml` has only one explicit `[[test]]` section:

```toml
[[test]]
name = "eval_runner_tests"
required-features = ["eval-runner"]
```

The three new files have no `[[test]]` entries. Cargo auto-discovers files in `tests/` so they will compile and run — but without explicit entries you cannot apply `required-features`, set `harness = false`, or control them independently. This is inconsistent with the established pattern in this `Cargo.toml`. Add explicit `[[test]]` entries for all three new files.

---

## Medium: `--ignore-tests` is not in `tarpaulin.toml`

The CI invokes:

```sh
cargo tarpaulin --skip-clean --ignore-tests --out Lcov --output-dir . --timeout 1800
```

The flag is on the command line, not in `tarpaulin.toml`. `tarpaulin.toml` currently only sets `exclude-files`. Consider moving `--ignore-tests` into `tarpaulin.toml` under `[config]` so the full effective configuration is visible in one place and cannot be accidentally dropped in a CI refactor:

```toml
[config]
ignore-tests = true
exclude-files = [
    "model/tests/*_integration.rs"
]
```

---

## Medium: two builder-only tests assert nothing about the constructed value

In `observability_coverage_tests.rs`:

```rust
fn observability_system_with_custom_health_threshold() {
    let threshold = HealthThreshold { cpu_threshold: 70.0, … };
    let system = ObservabilitySystem::new().with_health_thresholds(threshold);
    drop(system);   // no assertion
}

fn observability_system_with_immediate_critical_alert_policy() {
    let policy = AlertPolicy::immediate_critical();
    let system = ObservabilitySystem::new().with_alert_policy(policy);
    drop(system);   // no assertion
}
```

These confirm the builder calls compile and do not panic, but verify no postcondition. The production `with_health_thresholds` and `with_alert_policy` store values in private struct fields. A refactor that silently dropped the assignment would not be caught. At minimum, call a method that depends on the stored value (e.g., after `with_health_thresholds`, call `get_comprehensive_status` and check that trend analysis uses the threshold), or expose a test accessor.

---

## Medium: `alert_manager_history_limit_respected` should use `assert_eq!`, not `assert!`

```rust
let history = manager.get_alert_history(3).await.unwrap();
assert!(history.len() <= 3);
```

`<= 3` passes for 0, 1, 2, or 3 items. Since 5 alerts were inserted and the implementation does `iter().rev().take(limit)`, the correct expected result is exactly 3. Use `assert_eq!(history.len(), 3)` to catch an implementation that returns fewer items than requested.

---

## Low: four `HealthStatus` tests use `#[tokio::test]` unnecessarily

`health_status_healthy_is_healthy`, `health_status_warning_requires_attention`, `health_status_degraded_requires_attention`, and `health_status_unhealthy_requires_attention` contain no `.await` calls. Use plain `#[test]` for these to avoid spinning up a Tokio runtime unnecessarily.

---

## Low: `observability_system_uptime_increases` uses `>=` instead of `>`

```rust
assert!(t2 >= t1);
```

`>=` means the test passes even if `t2 == t1`, which can happen on platforms where `Instant` resolution is coarser than 5 ms. If the intent is to verify monotonic increase, use `assert!(t2 > t1)` and consider a longer sleep (e.g., 20 ms) to make it more robust across CI environments.

---

## Positive notes

- Tests that do make meaningful assertions are well-structured: `metrics_latency_min_lte_avg_lte_max`, `metrics_collector_record_error_increments_count`, `trace_guard_drop_removes_active_trace`, `trace_guard_record_error_sets_error_status`, and `telemetry_system_export_all_ok_and_clears_metrics` all verify real state transitions and would catch regressions.
- `alert_manager_acknowledge_nonexistent_returns_error` correctly tests an error path that was previously uncovered.
- No shared global mutable state is used across tests — each test constructs its own instance, which is the right isolation pattern.
- `tarpaulin.toml` and `codecov.yml` are not modified in this PR, so the codecov-guard workflow cannot block this PR on a coverage regression in those files.

---

## Required changes before merge

1. **Correct the PR description** to accurately explain why patch coverage passes (new lines are executed and hit, not excluded).
2. **Add explicit `[[test]]` entries** for the three new files in `harness/Cargo.toml`.
3. **Audit global tracing subscriber usage** and apply `#[serial]` (from `serial_test`, already in dev-dependencies) to any test that transitively calls `TelemetrySystem::initialize()`.
4. **Strengthen the two builder-only tests** to assert an observable postcondition, not just compile-and-drop.
5. **Fix `alert_manager_history_limit_respected`** to `assert_eq!(history.len(), 3)`.


---

DominicBurkart:
Code review in progress (still running). CI had one blocking lint failure: `assert_eq!(result.unwrap(), true)` in `telemetry_coverage_tests.rs` triggered `clippy::bool_assert_comparison`; fixed to `assert!(result.unwrap())` and pushed to branch. Waiting for CI re-run to confirm green before marking ready for review — https://claude.ai/code/session_01Tqbaxcb9b9L5s7RNr45RvN

---

PR #363, opened 2026-05-08
feat(eval): harness-score bin + scoring module for SWE-bench scorecard

DominicBurkart:
## Summary
- New \`harness::eval::scoring\` module — pure state machine + score aggregation.
- New \`harness-score\` bin: drives nanna across a SWE-bench JSONL dataset, persists per-instance state, supports resume across sittings, and \`--finalize\` batches the verifier and appends a JSONL row.
- Scorecard row pinned at \`instances_total = 300\` (SWE-bench Lite) so partial runs and full runs share the same denominator. Only \`is_complete: true\` is leaderboard-comparable.

## Stack
Stacked on PR #362 (which is itself stacked on #361). Merge order: #361 → #362 → this. The base of this PR is \`feat/eval-runner-skip-verify\` so the diff stays surgical.

## Test plan
- [x] 7 new unit tests in \`eval::scoring\` (terminal status, full denominator, completeness, JSON round-trip, zero-total).
- [x] \`cargo nextest run --features eval-runner -p harness\` — 608/608 pass.
- [x] \`cargo clippy --features eval-runner -p harness --all-targets -- -D warnings\` — clean.
- [x] \`cargo fmt --all -- --check\` — clean.
- [x] Pre-commit hook (workspace test suite + audit + deny + coverage) — passed locally.
- [ ] Smoke run against \`evals/datasets/swebench-verified-sample.jsonl\` (5 instances) once #361 + #362 land. (Will use \`docker\` via \`podman.socket\` and the existing \`~/.venvs/swebench\`.)

## Out of scope (per the approved plan)
- Committed JSONL line — separate PR after a real local run.
- \`.gitignore\` of state dir — separate PR.
- Charts / rendering / self-hosted CI runner / full Verified slice.

### DIFF
--- a/harness/Cargo.toml
+++ b/harness/Cargo.toml
@@ -16,6 +16,11 @@ name = "harness-eval"
 path = "src/bin/eval.rs"
 required-features = ["eval-runner"]
 
+[[bin]]
+name = "harness-score"
+path = "src/bin/score.rs"
+required-features = ["eval-runner"]
+
 [features]
 # Gate the eval runner and its tempfile dependency behind this feature so that
 # production builds (the harness binary, container images) do not pull in
--- a/harness/src/bin/score.rs
+++ b/harness/src/bin/score.rs
@@ -1,0 +1,357 @@
+//! `harness-score` — drive nanna across a SWE-bench dataset, persist
+//! per-instance state across sittings, and emit a leaderboard-comparable
+//! score line into `evals/scorecards/index.jsonl`.
+//!
+//! Most of the testable logic lives in [`harness::eval::scoring`]; this
+//! binary is the async glue that calls `run_eval` (per instance) and
+//! `verify_predictions` (once in `--finalize`).
+
+use std::path::PathBuf;
+use std::process::ExitCode;
+use std::time::Duration;
+
+use clap::Parser;
+use harness::eval::runner::{run_eval, EvalRunResult, EvalRunnerConfig, EvalRunnerError};
+use harness::eval::scoring::{
+    aggregate_scorecard, append_line, instance_state_from_outcome, load_all_states,
+    outcome_from_run_eval, predictions_dir, read_state, resolve_model, sanitize_model_for_path,
+    should_attempt, state_path, states_dir, write_state, InstanceState, InstanceStatus, Scorecard,
+    ScorecardMetadata, StoredVerdict,
+};
+use harness::eval::swebench::{adapt_to_eval_case, load_swebench_dataset, SWEBenchTask};
+use harness::eval::swebench_verify::{verify_predictions, Prediction, VerifyConfig};
+
+#[derive(Parser, Debug)]
+#[command(
+    name = "harness-score",
+    about = "Run nanna across a SWE-bench dataset, persist state, emit a scorecard line"
+)]
+struct Args {
+    #[arg(long)]
+    dataset: PathBuf,
+
+    #[arg(long)]
+    state_dir: PathBuf,
+
+    #[arg(long)]
+    model: Option<String>,
+
+    #[arg(long)]
+    max_instances: Option<usize>,
+
+    #[arg(long, default_value_t = 4)]
+    max_workers: usize,
+
+    #[arg(long)]
+    timeout_secs: Option<u64>,
+
+    #[arg(long, default_value_t = 100)]
+    max_iterations: usize,
+
+    #[arg(long, default_value = "princeton-nlp/SWE-bench_Lite")]
+    dataset_name: String,
+
+    #[arg(long)]
+    run_id: Option<String>,
+
+    #[arg(long)]
+    finalize: bool,
+}
+
+fn iso_now() -> String {
+    let secs = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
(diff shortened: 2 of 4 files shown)


DominicBurkart (commented):
## Review of feat(eval): harness-score bin + scoring module

Overall the architecture is clean: pure state machine in `scoring.rs`, I/O in `score.rs`, good test coverage for the pure layer. The `instances_total` design (derived from the dataset, not the state-file count) is correct for leaderboard parity. Three issues below need fixing before merge; the rest are lower-priority calls.

---

### Bug (correctness): silent diff-read failure inflates the "no upstream report" fallback into a resolved=false score entry

In `finalize`, when `std::fs::read_to_string(dpath)` fails, the instance is silently skipped with a `::warning::` and never added to `predictions`. Later, in the merge loop, every `CompletedDiff` instance not found in `verdict_by_id` receives:

```rust
StoredVerdict {
    resolved: false,
    error: Some("no upstream report".to_string()),
}
```

…and is transitioned to `Verified`. So a broken diff file silently becomes a `Verified(resolved=false)` state, permanently blocking retry and misrepresenting the run. The instance should either stay as `CompletedDiff` (so re-running `--finalize` can retry it) or fail hard. A `continue` that skips adding to `predictions` but still applies the `no upstream report` fallback is the worst of both worlds.

Fix: track which instance IDs were successfully added to `predictions`, and in the merge loop only apply the `no upstream report` fallback to those — leave instances whose diff file was unreadable in `CompletedDiff` so the next `--finalize` retries them.

---

### Bug (portability): `diff_path` stores the absolute filesystem path of the diff

`build_state_after_attempt` stores `dpath.display().to_string()` — an absolute path — in `InstanceState::diff_path`. The `finalize` path reads that string verbatim. If `--state-dir` is ever moved (synced to another machine, reproduced in CI with a different prefix), every diff becomes unreadable and triggers the silent-skip bug above.

The canonical path is fully derivable from `instance_id`: `diff_path(&args.state_dir, &s.instance_id)` already does this in `finalize`. Suggestion: drop `diff_path` from the serialised `InstanceState` (or make it an advisory hint) and always recompute it from the current `--state-dir` at finalize time. This eliminates both the portability problem and the dependency on the stored string being correct.

---

### Bug (path resolution): `index.jsonl` is resolved against the process CWD, not the repo root

```rust
let index_path = std::path::Path::new("evals/scorecards/index.jsonl");
```

If `harness-score` is invoked from anywhere other than the repo root (e.g. via `cargo run --manifest-path harness/Cargo.toml` from a parent dir, or from a CI working directory), the file lands in the wrong place with no diagnostic. Add a `--scorecard-path` CLI flag with `evals/scorecards/index.jsonl` as the documented default, so callers can always pass an absolute path.

---

### Nit: `iso_now` uses the deprecated `from_timestamp` API

`chrono::DateTime::<chrono::Utc>::from_timestamp(secs as i64, 0)` is deprecated since chrono 0.4.23. Replace the whole function with `chrono::Utc::now().format("%Y-%m-%dT%H:%M:%SZ").to_string()` — simpler, no deprecated call, no manual UNIX epoch arithmetic.

---

### Nit: `swebench_skip_verify` pushes a string into `failures`, making `EvalRunResult::success` false in batched mode

```rust
failures.push("verifier skipped — score-mode batched run".to_string());
```

`harness-score` ignores `result.success` and only checks `result.swebench_patch`, so this doesn't cause a scoring error here. But any other caller that checks `result.failures` or `result.success` after calling `run_eval` with `skip_verify = true` will see a confusing failure entry. Consider a dedicated `EvalRunResult` state or an informational log line instead of co-opting `failures`.

---

### Nit: `AgentError` is permanently terminal with no in-binary recovery path

The `is_terminal` doc says "user clears the state file", but there is no `--reset-instance <id>` flag and no documented recovery command. A transient error (network blip during clone, Docker daemon restart) permanently blocks the instance until the operator manually hunts down and deletes a JSON file. At minimum, document the recovery path in the `--help` text or the module-level doc comment.

---

### Minor: `finalize` re-loads the full dataset to count `instances_total`

The dataset is loaded a second time solely to call `.len()`. If the dataset file changes between the agent run and `--finalize`, `instances_total` in the scorecard silently diverges from what was actually run. Consider persisting `instances_total` (and `dataset_name`) in a small run-manifest file written at the start of `score_loop`, and reading that in `finalize` rather than re-parsing the full dataset.

---

No secrets or PII found in the diff.

---

DominicBurkart:
**CRON review complete** — code review, CI check, and comment triage all done.

Code review found three bugs requiring fixes before merge: (1) silent diff-read failures in `finalize` permanently mark instances as `Verified(resolved=false)` — they should stay as `CompletedDiff` for retry; (2) `diff_path` serializes an absolute filesystem path, breaking cross-machine `--finalize` (recompute from `instance_id` + `--state-dir` at finalize time); (3) `Path::new("evals/scorecards/index.jsonl")` resolves relative to CWD, needs a `--scorecard-path` flag. macOS/Windows bring-up failures are infrastructure-wide and pre-existing; the full unit/lint/security CI suite did not run because this is a stacked PR — a rebase onto current `main` is needed to trigger it; PR remains in draft pending these fixes.

---

PR #366, opened 2026-05-09
feat(score): canonicalize verify_dir + complete-only scorecard append

DominicBurkart:
## Summary
Two related fixes to `harness-score` so the scorecard machinery survives real use, plus a couple of `.gitignore` entries:

1. **Canonicalise `verify_dir`** before handing it to the upstream Python harness. The harness sets its CWD to `verify_dir` and re-opens the predictions path relative to that CWD; with a relative input the path double-nests and the harness errors with `FileNotFoundError`. Caught attempting the first real finalize against `evals/scorecards/state/local-1`.

2. **Append the scorecard line iff `is_complete`.** Partial subdivisions of a curated leaderboard slice (Lite, Verified, ...) aren't comparable to anything published at swebench.com and just clutter the time-series file. No opt-in flag — the per-run state directory already retains everything needed for progress tracking. Per [reviewer feedback on the original PR diff](https://github.com/DominicBurkart/nanna-coder/pull/366#discussion_r3212790208).

3. `.gitignore` entries for `evals/scorecards/state/` (per-run cloned repos, predictions, upstream report.json trees) and locally-exported swebench JSONLs.

This PR no longer adds a row to `evals/scorecards/index.jsonl`. The metric file gets created by the bin's own `append_line` the first time a full-dataset run finalises (see #367 for the broader eval-architecture refactor that should follow).

## Test plan
- [x] \`cargo nextest run --features eval-runner -p harness eval::scoring\` — 23 tests pass (no scoring-module change).
- [x] \`cargo clippy --features eval-runner -p harness --all-targets -- -D warnings\` — clean.
- [x] Pre-commit hook (full workspace test suite + audit + deny + coverage) — passed locally.
- [x] Manual: \`harness-score --finalize\` against a 5-of-300 partial state-dir prints \"NOT appended to evals/scorecards/index.jsonl — 5/300 instances attempted; only full-dataset runs are leaderboard-comparable\" and exits 0; the metric file is *not* created.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,13 @@ result-*
 # produced on demand by `harness::eval::swebench::materialize`. Only the
 # `task.toml` files under swebench-*/ are checked in.
 evals/cases/swebench-*/repo/
+
+# `harness-score` per-run state (cloned repos, predictions, per-instance
+# state files, upstream report.json trees). Only the aggregated scorecard
+# at evals/scorecards/index.jsonl is checked in.
+evals/scorecards/state/
+
+# Locally-exported SWE-bench datasets. Pulled on demand from HuggingFace
+# via `~/.venvs/swebench/bin/python -c 'datasets.load_dataset(...)'`.
+evals/datasets/swebench-lite.jsonl
+evals/datasets/swebench-verified.jsonl
--- a/harness/src/bin/score.rs
+++ b/harness/src/bin/score.rs
@@ -250,6 +250,16 @@ async fn finalize(args: &Args, model: &str, run_id: &str) -> ExitCode {
             eprintln!("::error::failed to create verify dir: {e}");
             return ExitCode::from(2);
         }
+        // Canonicalise: the upstream Python harness sets its CWD to
+        // `work_dir` and then re-opens the predictions path relative to
+        // that CWD. If we pass a relative path, it double-nests.
+        let verify_dir = match verify_dir.canonicalize() {
+            Ok(p) => p,
+            Err(e) => {
+                eprintln!("::error::failed to canonicalize verify dir: {e}");
+                return ExitCode::from(2);
+            }
+        };
         let verify_config = VerifyConfig {
             dataset_name: args.dataset_name.clone(),
             model_name_or_path: format!("nanna__{}", sanitize_model_for_path(model)),
@@ -328,17 +338,22 @@ async fn finalize(args: &Args, model: &str, run_id: &str) -> ExitCode {
         }
     };
 
-    let index_path = std::path::Path::new("evals/scorecards/index.jsonl");
-    if let Err(e) = append_line(index_path, &line) {
-        eprintln!("::error::failed to append scorecard: {e}");
-        return ExitCode::from(2);
-    }
+    let appended = if card.is_complete {
+        let index_path = std::path::Path::new("evals/scorecards/index.jsonl");
+        if let Err(e) = append_line(index_path, &line) {
+            eprintln!("::error::failed to append scorecard: {e}");
+            return ExitCode::from(2);
+        }
+        true
+    } else {
+        false
+    };
 
-    print_summary(&card);
+    print_summary(&card, appended);
     ExitCode::SUCCESS
 }
 
-fn print_summary(card: &Scorecard) {
+fn print_summary(card: &Scorecard, appended: bool) {
     println!();
     println!("=== Scorecard ===");
     println!("dataset:    {}", card.dataset);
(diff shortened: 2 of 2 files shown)


evals/scorecards/index.jsonl
```diff
@@ -0,0 +1 @@
+{"schema_version":1,"date":"2026-05-09T00:35:36Z","commit":"20b38dae27dd20baf0360ba926138df9fd1f1904","branch":"score/local-1-jsonl","pr":null,"model":"gemma4:e4b","dataset":"princeton-nlp/SWE-bench_Lite","instances_total":300,"instances_attempted":5,"instances_resolved":0,"score_pct":0.0,"is_complete":false,"state_dir":"evals/scorecards/state/local-1"}
```
DominicBurkart: You need to run the complete SWE-bench Lite to contribute in this file: https://www.swebench.com/lite.html is already a subset of SWE-bench verified; we shouldn't break it down further.

DominicBurkart: Right — pushed an amended commit that drops the partial row entirely and adds a guardrail in the bin: `--finalize` now refuses to append to `index.jsonl` unless `is_complete: true`. No opt-in flag for partial appends. Partial subdivisions of a curated slice (Lite, Verified, ...) aren't comparable to anything published, so they don't belong in the time-series file even as smoke entries. The per-run state directory already keeps progress info for resume.

Filed #367 for the deeper concern: the score today measures the agent loop excised from the rest of nanna (no test feedback during the run, MCP/observability/entity-store all bypassed). That's a refactor; this PR stays focused on the bin/gitignore cleanup.

Kicking off the full 300-instance Lite next; that landing PR will add the first real `is_complete: true` row.

DominicBurkart (commented):
Automated code review — commit 45d874e.

## Summary

Both fixes are correct and the implementation is clean. A few observations and one minor issue worth noting.

---

### 1. `std::fs::canonicalize` correctness

The call is placed **after** `create_dir_all` succeeds, so the path is guaranteed to exist when `canonicalize` is called. That ordering is important: `canonicalize` returns an error on non-existent paths on all major OSes, and the code here gets it right.

One edge case to be aware of: on Linux, `canonicalize` resolves symlinks. If `state_dir` is (or contains) a symlink, the canonical path returned will be the real path, which is what the Python harness needs. This is correct behaviour here, not a bug.

No issues with this change.

---

### 2. `is_complete` gate on scorecard append

The `is_complete` predicate in `aggregate_scorecard` requires:
- `instances_total > 0`
- `attempted == instances_total` (every instance has a state file)
- **all** states are `InstanceStatus::Verified`

The third condition is stricter than "all terminal" — an instance stuck at `CompletedDiff` (awaiting verify) or in `AgentError`/`MaterializeError` will prevent `is_complete` from being set, which means a run where some instances failed to verify will never append to the leaderboard file. That seems intentional and correct: a partial verify isn't leaderboard-comparable either.

The `print_summary` message in the non-appended branch says *"Re-run without --max-instances until every instance reaches a terminal status, then re-finalize."* However, `CompletedDiff` **is** terminal per `InstanceStatus::is_terminal`, but it will still block `is_complete` (because it isn't `Verified`). A user who has instances stuck in `CompletedDiff` after re-finalizing would be confused by this message — it tells them to re-run the scorer, but the issue is that `verify_predictions` didn't return a verdict for those instances, not that they haven't been attempted. Consider adjusting the message to also mention re-running `--finalize` if instances are in `completed_diff` state.

This is a minor UX nit, not a correctness issue.

---

### 3. `.gitignore` patterns

- `evals/scorecards/state/` — correct; matches the directory and everything beneath it.
- `evals/datasets/swebench-lite.jsonl` and `evals/datasets/swebench-verified.jsonl` — exact filename matches, which is fine for the current datasets. If additional split variants appear later (e.g. `swebench-test.jsonl`) they'd need explicit entries. A glob `evals/datasets/swebench-*.jsonl` might be more forward-compatible, but this is a style preference, not a bug.
- Both new blocks are missing a trailing newline — the diff shows `\ No newline at end of file` for the final line. The file ended without one before this PR too (`evals/cases/swebench-*/repo/` was also the last line without a newline), so this PR doesn't make it worse, but it's worth fixing both at once to keep the file well-formed.

---

### 4. Test coverage

The existing `aggregate_marks_complete_only_when_all_verified` test covers the `is_complete = true` path. The `complete-only` append gate in the binary is not directly unit-tested (it lives in the binary, not in `scoring.rs`), but the predicate it relies on is well-covered. No gaps that need filling before merge.

The PR description's manual test plan (partial run produces "NOT appended" message and exits 0) is exactly the right thing to verify for this change.

---

### Overall

The two substantive changes are correct, the `is_complete` logic is sound, and the test suite is in good shape. The only actionable item is the slightly misleading guidance in the "NOT appended" message for instances stuck in `CompletedDiff`. Everything else is ready to merge once the draft status is cleared.

---

DominicBurkart:

Review posted with no blocking issues: canonicalization placed correctly after `create_dir_all`, `is_complete` predicate sound across all terminal states; minor nits only (missing trailing newline on `.gitignore` blocks, dataset glob could be more forward-compatible). DominicBurkart's prior feedback was already self-addressed by the author; CI is green. **Marked ready for review.**

---

PR #368, opened 2026-05-09
feat(eval): category discriminator on scorecard rows (PR-A of 9)

DominicBurkart:
## Summary
- Bumps `Scorecard.schema_version` to 2 and adds `ScorecardCategory` (string-tag enum: `nanna_solo`, `claude_solo`, `claude_mcp_claude`, `claude_mcp_nanna_gemma4`) plus `orchestrator_model` / `worker_model` fields.
- `harness-score` gains required `--category` and optional `--orchestrator-model` / `--worker-model` flags. Existing `--model` is retained as a back-compat alias.
- The verifier scenario label changes from `nanna__<model>` to `<category>__<model>` so SWE-bench upstream reports are also unambiguously attributable to a track.

## Why
The current `Scorecard` distinguishes runs only by `model`/`dataset`/`date`, so the four planned eval categories (nanna-solo, claude-solo, and the two MCP-delegated claude variants) would collide if recorded side-by-side. This is the foundation for PR-B…PR-I in the multi-track-eval dev plan.

## Coordination call-out
This PR must merge **before** the in-flight nanna-solo 300-instance score loop calls `--finalize`. If it lands first, the resumed run uses `--category nanna_solo --orchestrator-model gemma4:e4b` and the first complete row in `evals/scorecards/index.jsonl` is schema_version=2. If it doesn't, that row lands as schema_version=1 and we'd need a one-off migration.

Suggested resume command after merge:
```
harness-score \
  --dataset evals/datasets/swebench-lite.jsonl \
  --state-dir evals/scorecards/state/local-1 \
  --category nanna_solo \
  --orchestrator-model gemma4:e4b \
  --finalize
```

## Schema migration notes
- `InstanceState`'s on-disk format is unchanged. The previously-shared `SCHEMA_VERSION` const splits into `INSTANCE_STATE_SCHEMA_VERSION` (still 1) and `SCORECARD_SCHEMA_VERSION` (now 2), so future bumps to either are independent.
- `Scorecard.model` is retained as a duplicate of `orchestrator_model` for v1-reader compatibility.
- A v1-shaped JSON line (no `category` / `orchestrator_model`) is rejected by the v2 struct — the test `scorecard_v1_payload_is_rejected_under_v2_struct` documents this intentional break.

## Test plan
- [x] `cargo test --workspace --all-features` — 511+ tests pass, including 7 new category tests
- [x] `cargo deny check` advisories/bans/licenses/sources OK
- [x] CLI smoke: `--category` is required, `--category nanna-solo` (with hyphen) is rejected with a list of valid values
- [x] CLI smoke: `--help` documents all four categories and the deprecation note for `--model`

## What's next (out of scope for this PR)
- PR-B: rename `claude_token_usage` / `nanna_token_usage` → `orchestrator_token_usage` / `worker_token_usage` on `SweBenchInstanceResult`
- PR-C: `AnthropicProvider` impl in `model/src/anthropic.rs`
- PR-D: provider injection + concurrency semaphore in the eval runner
- PR-E…I: per-category pilot + full 300-instance scoring runs

### DIFF
--- a/harness/src/bin/score.rs
+++ b/harness/src/bin/score.rs
@@ -8,6 +8,7 @@
 
 use std::path::PathBuf;
 use std::process::ExitCode;
+use std::str::FromStr;
 use std::time::Duration;
 
 use clap::Parser;
@@ -16,10 +17,14 @@ use harness::eval::scoring::{
     aggregate_scorecard, append_line, instance_state_from_outcome, load_all_states,
     outcome_from_run_eval, predictions_dir, read_state, resolve_model, sanitize_model_for_path,
     should_attempt, state_path, states_dir, write_state, InstanceState, InstanceStatus, Scorecard,
-    ScorecardMetadata, StoredVerdict,
+    ScorecardCategory, ScorecardMetadata, StoredVerdict,
 };
 use harness::eval::swebench::{adapt_to_eval_case, load_swebench_dataset, SWEBenchTask};
 use harness::eval::swebench_verify::{verify_predictions, Prediction, VerifyConfig};
+
+fn parse_category(s: &str) -> Result<ScorecardCategory, String> {
+    ScorecardCategory::from_str(s)
+}
 
 #[derive(Parser, Debug)]
 #[command(
@@ -33,6 +38,27 @@ struct Args {
     #[arg(long)]
     state_dir: PathBuf,
 
+    /// Eval-track discriminator. Required so the scorecard row written to
+    /// `evals/scorecards/index.jsonl` is unambiguously attributable to one
+    /// of the four categories (nanna_solo, claude_solo, claude_mcp_claude,
+    /// claude_mcp_nanna_gemma4).
+    #[arg(long, value_parser = parse_category)]
+    category: ScorecardCategory,
+
+    /// Foundational model that drives the agent loop / orchestrator.
+    /// Takes precedence over the deprecated `--model` flag and over the
+    /// `NANNA_EVAL_MODEL` / `MODEL` env vars.
+    #[arg(long)]
+    orchestrator_model: Option<String>,
+
+    /// Worker-side model when running an MCP-delegated category. Recorded
+    /// on the scorecard row but not yet wired through to a worker provider
+    /// (that lands in the runner-mode PR).
+    #[arg(long)]
+    worker_model: Option<String>,
+
+    /// Deprecated alias for `--orchestrator-model`. Retained so resumed
+    /// runs from before schema v2 do not break.
     #[arg(long)]
     model: Option<String>,
 
@@ -87,7 +113,10 @@ fn git_capture(args: &[&str]) -> Option<String> {
 #[tokio::main]
 async fn main() -> ExitCode {
     let args = Args::parse();
-    let model = resolve_model(args.model.as_deref(), "qwen3:0.6b");
+    // `--orchestrator-model` is canonical; `--model` is the back-compat
+    // alias, used only when the new flag is absent.
+    let cli_model = args.orchestrator_model.as_deref().or(args.model.as_deref());
+    let model = resolve_model(cli_model, "qwen3:0.6b");
     let run_id = args.run_id.clone().unwrap_or_else(run_id_default);
 
     if let Err(e) = std::fs::create_dir_all(predictions_dir(&args.state_dir)) {
@@ -252,7 +281,11 @@ async fn finalize(args: &Args, model: &str, run_id: &str) -> ExitCode {
         }
         let verify_config = VerifyConfig {
             dataset_name: args.dataset_name.clone(),
-            model_name_or_path: format!("nanna__{}", sanitize_model_for_path(model)),
+            model_name_or_path: format!(
+                "{}__{}",
+                args.category.as_str(),
+                sanitize_model_for_path(model)
+            ),
             run_id: run_id.to_string(),
             work_dir: verify_dir,
             max_workers: args.max_workers,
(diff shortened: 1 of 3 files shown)


DominicBurkart (commented):
## Code Review — feat(eval): category discriminator on scorecard rows (PR-A of 9)

Overall the change is well-structured and the PR description is unusually thorough about coordination concerns. The implementation is clean. A few things worth discussing before merge, mostly around the intentional backward-compat break and a couple of smaller gaps.

---

### Schema migration safety / backward compat — key observations

**Intentional v1 rejection is documented but the coordination window is fragile.**
The PR explicitly tests that a v1-shaped JSONL row fails to deserialise as a v2 `Scorecard` (`scorecard_v1_payload_is_rejected_under_v2_struct`). The test is correct and the intent is clearly stated. However, the only thing guarding against a v1 row landing in `evals/scorecards/index.jsonl` is the "merge me first" coordination note in the PR body. There is no runtime guard in `finalize()` that checks whether the index file already contains schema_version=1 rows and refuses to append a v2 row (or vice-versa). If the timing is missed, the file becomes a mixed-version JSONL that any future reader must handle. Consider either:
- A startup assertion in `finalize()` that reads the last line of `index.jsonl` (if it exists) and hard-errors if `schema_version != SCORECARD_SCHEMA_VERSION`, or
- Noting explicitly in the dev plan that PR-B must include a small migration script so the mixed-version case has a known recovery path.

**`model` field duplication may quietly mislead future readers.**
`Scorecard.model` is serialised as a duplicate of `orchestrator_model`. Because `Scorecard` derives `Serialize`, both fields will appear in every written row. A downstream reader that only looks at `model` will get the right value today, but once PR-B renames `claude_token_usage`/`nanna_token_usage` the conceptual gap between the old `model` field and the new `orchestrator_model` field widens. It would help to add a `#[serde(rename = "model")]` comment or a `// deprecated` rustdoc on the field itself (it has a doc comment, which is good, but a `#[deprecated]` attribute would make Rust warn if internal code accidentally reads `card.model` instead of `card.orchestrator_model`).

**No `#[serde(deny_unknown_fields)]` on `Scorecard`.**
Forward-compatibility in the other direction (a v3 row being read by v2 code) is currently silent: extra fields are silently ignored. This is probably fine for an append-only JSONL time-series, but it means `schema_version` alone is the version signal and nothing enforces it. At minimum it's worth noting this in the struct doc comment so future PRs don't add `deny_unknown_fields` without understanding the implication.

---

### `--category` flag validation

The custom `parse_category` wrapper is a one-liner that just calls `ScorecardCategory::from_str`:

```rust
fn parse_category(s: &str) -> Result<ScorecardCategory, String> {
    ScorecardCategory::from_str(s)
}
```

Since `ScorecardCategory` already implements `FromStr`, clap can use it directly via `value_parser = clap::value_parser!(ScorecardCategory)` once `ScorecardCategory: TypedValueParser` — but the manual wrapper works fine and the error message it produces (listing valid options) is good. No blocking issue here, just a nit.

The CLI smoke tests listed in the PR description cover the hyphen rejection case (`--category nanna-solo`). It would be worth adding a unit test that constructs an `Args` via `try_parse_from` (clap's test helper) and asserts the error message contains the list of valid values, so this is regression-protected in CI rather than only in manual smoke tests.

---

### Enum exhaustiveness

`ScorecardCategory::ALL` is a hand-maintained const array. If a new variant is ever added, it's easy to forget to extend `ALL`, which would make `FromStr`'s error message silently incomplete and break the `category_serialises_to_snake_case` test only if the test iterates `ALL`. The `from_str` match arm is the canonical source of truth; `ALL` is derived from it manually. A `#[cfg(test)]` assertion that `ALL.len()` matches a compile-time count isn't possible in stable Rust today, but a comment on `ALL` flagging "keep in sync with the match in FromStr" would make future contributors aware of the coupling.

---

### Test coverage

The seven new tests are well-chosen:
- Round-trip serialisation per variant (`category_serialises_to_snake_case`)
- `FromStr` / `as_str` round-trip (`category_from_str_matches_serialised_form`)
- Typo rejection with useful error messages (`category_from_str_rejects_typos_and_unknown`)
- Serde rejection of unknown string (`deserialise_rejects_unknown_category_string`)
- Full `Scorecard` round-trip with MCP category (`scorecard_v2_round_trips_with_category_and_models`)
- Solo track `worker_model` is null (`scorecard_solo_track_has_no_worker_model`)
- v1 payload rejection (`scorecard_v1_payload_is_rejected_under_v2_struct`)

One gap: there is no test that calls `aggregate_scorecard` with a `worker_model: Some(...)` for a *solo* category (e.g., `NannaSolo` + `worker_model: Some("some-model")`). The struct allows this combination — it's not validated at construction time. If solo tracks are semantically required to have `worker_model: None`, a debug assertion or a `new()` constructor that enforces the invariant would close the gap.

---

### Minor / nits

- The `parse_category` free function could be inlined as a closure at the call-site or removed in favour of `value_parser = ScorecardCategory::from_str as fn(&str) -> _`, but the current approach is readable.
- `print_summary` alignment change (padding to 20 chars) is cosmetic and fine.
- The `SCORING_SCHEMA_VERSION` re-export in `mod.rs` is now split into `SCORING_INSTANCE_STATE_SCHEMA_VERSION` and `SCORING_SCORECARD_SCHEMA_VERSION` — any downstream crate that imported the old name will break. Worth checking if there are other crates in the workspace that used `SCORING_SCHEMA_VERSION`.

---

### Summary

The implementation is solid and the test suite is thorough for the new types. The main concern is that the hard-compatibility break (v1 rows rejected at parse time) is protected only by coordination discipline rather than a runtime guard, which creates a narrow but real window for a mixed-version index file. Everything else is either a nit or a suggestion for a follow-up PR.

---

DominicBurkart:

Review posted: schema migration logic is correct and intentional v1-rejection is tested; minor findings include missing `#[deprecated]` on the compat `model` field, no `deny_unknown_fields` noted, `ScorecardCategory::ALL` hand-maintained array can silently diverge from `FromStr`, and no test enforcing that solo-category rows have `worker_model = None`. No DominicBurkart comments to address; CI is green (only pre-existing macOS/Windows infra failures). **Marked ready for review** — note the coordination dependency: merge before `--finalize` is called on the in-flight nanna-solo run.

---

PR #370, opened 2026-05-09
feat(model): AnthropicProvider against /v1/messages (PR-C of 9)

DominicBurkart:
## Summary
- Adds an `anthropic` feature flag on the `model` crate.
- New `model::anthropic` module with `AnthropicProvider` implementing `ModelProvider` for Anthropic's `/v1/messages` API.
- Real translation layer between the workspace's OpenAI-shaped `ChatRequest` / `ChatResponse` and Anthropic's content-block-oriented wire format.

## Why
Cats (1)–(3) of the multi-track-eval plan all run Claude as the orchestrator. The `ModelProvider` trait at `model/src/provider.rs:35` is the natural seam — only `OllamaProvider` exists today. Slot a Claude variant in there once and all three categories share the plumbing.

## Translation rules covered (each unit-tested)
- System messages lifted to the top-level `system` field; multiple system messages concatenated.
- Assistant `tool_calls` → `tool_use` content blocks; `ToolCall.id` preserved as `tool_use_id` for round-trip.
- `MessageRole::Tool` → `user` message with a `tool_result` content block.
- Consecutive same-role messages merged so two parallel tool responses become one user message with two `tool_result` blocks (Anthropic rejects adjacent same-role messages otherwise).
- `tool_choice` mapping: `Auto`→`auto`, `Required`→`any`, `None`→`none`, `Specific(name)`→`{type:"tool",name}`.
- `max_tokens` defaults to 4096 (Anthropic requires it).
- Response stop reasons: `end_turn`/`stop_sequence`→`Stop`, `max_tokens`→`Length`, `tool_use`→`ToolCalls`.
- `usage.input_tokens`/`output_tokens` flow through to `Usage`, populating the scorecard's `orchestrator_token_usage`.

## Auth + health check
- `AnthropicProvider::from_env()` reads `ANTHROPIC_API_KEY`.
- `health_check` hits `GET /v1/models` (auth-validating, no token spend).

## Test plan
- [x] `cargo test --workspace --all-features` — 504+ tests, 18 new translation tests, 1 new doctest
- [x] Default-feature build (no anthropic) still green
- [ ] Live integration test (`#[ignore]`) with `ANTHROPIC_API_KEY` set — owner to run with `cargo test --features anthropic -- --ignored`

## What's next
PR-D wires `AnthropicProvider` into the eval runner via a `--orchestrator-provider` CLI flag and adds a concurrency semaphore. Depends on PR-A (#368) for the scorecard schema; this PR is independent of A and B.

### DIFF
--- a/model/Cargo.toml
+++ b/model/Cargo.toml
@@ -12,6 +12,7 @@ repository.workspace = true
 [features]
 default = ["ollama"]
 ollama = ["dep:ollama-rs"]
+anthropic = []
 
 [dependencies]
 async-trait = "0.1"
--- a/model/src/anthropic.rs
+++ b/model/src/anthropic.rs
@@ -48,10 +48,12 @@ pub const DEFAULT_MAX_TOKENS: u32 = 4096;
 /// Provider implementation for Anthropic's Messages API.
 ///
 /// Constructed once and reused across requests; the underlying
-/// [`reqwest::Client`] holds a connection pool.
+/// [`reqwest::Client`] holds a connection pool. The API key is validated
+/// once at construction (must be ASCII-printable so it can become an HTTP
+/// header value); subsequent requests cannot panic on header assembly.
 pub struct AnthropicProvider {
     client: reqwest::Client,
-    api_key: String,
+    api_key: reqwest::header::HeaderValue,
     base_url: String,
 }
 
@@ -62,17 +64,28 @@ impl AnthropicProvider {
             std::env::var("ANTHROPIC_API_KEY").map_err(|_| ModelError::InvalidConfig {
                 message: "ANTHROPIC_API_KEY env var not set".to_string(),
             })?;
-        Ok(Self::with_api_key(api_key))
+        Self::with_api_key(api_key)
     }
 
     /// Construct with an explicit API key. Use [`AnthropicProvider::from_env`]
-    /// for the common case.
-    pub fn with_api_key(api_key: impl Into<String>) -> Self {
-        Self {
+    /// for the common case. Returns [`ModelError::InvalidConfig`] if the
+    /// key contains characters that cannot appear in an HTTP header value
+    /// (e.g. a stray BOM, newline, or non-ASCII byte).
+    pub fn with_api_key(api_key: impl Into<String>) -> ModelResult<Self> {
+        let raw = api_key.into();
+        let mut header = reqwest::header::HeaderValue::from_str(&raw).map_err(|_| {
+            ModelError::InvalidConfig {
+                message: "ANTHROPIC_API_KEY contains characters that cannot be sent as an HTTP \
+                          header value (non-ASCII / control / BOM)"
+                    .to_string(),
+            }
+        })?;
+        header.set_sensitive(true);
+        Ok(Self {
             client: reqwest::Client::new(),
-            api_key: api_key.into(),
+            api_key: header,
             base_url: DEFAULT_BASE_URL.to_string(),
-        }
+        })
     }
 
     /// Override the base URL, e.g. for tests or proxies.
@@ -83,11 +96,7 @@ impl AnthropicProvider {
 
     fn auth_headers(&self) -> reqwest::header::HeaderMap {
         let mut h = reqwest::header::HeaderMap::new();
-        h.insert(
-            "x-api-key",
-            reqwest::header::HeaderValue::from_str(&self.api_key)
-                .expect("api key is ascii printable"),
-        );
+        h.insert("x-api-key", self.api_key.clone());
         h.insert(
             "anthropic-version",
             reqwest::header::HeaderValue::from_static(ANTHROPIC_VERSION),
@@ -376,8 +385,15 @@ fn translate_request(request: ChatRequest) -> AnthropicRequest {
         });
     }
 
-    let tools = tools.map(|defs| defs.into_iter().map(tool_definition_to_anthropic).collect());
-    let tool_choice = tool_choice.map(tool_choice_to_value);
+    let tools: Option<Vec<AnthropicTool>> =
+        tools.map(|defs| defs.into_iter().map(tool_definition_to_anthropic).collect());
+    // Anthropic returns 400 if `tool_choice` is set but `tools` is absent.
+    // Drop a stale `tool_choice` rather than emitting an invalid payload.
+    let tool_choice = if tools.is_some() {
+        tool_choice.map(tool_choice_to_value)
+    } else {
+        None
+    };
 
     AnthropicRequest {
         model,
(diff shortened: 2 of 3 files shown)


DominicBurkart (commented):
## Review — AnthropicProvider (PR-C of 9)

Reviewed commit 7e15c41. No previous review existed for this PR.

---

### Security / API key exposure

**No logging of the key detected.** `auth_headers()` assembles the key into a `HeaderMap` and the key field itself is never passed to any formatting macro. The `ModelError` variants that surface HTTP error bodies (`Unknown`, `ServiceUnavailable`) embed response body text only, not request headers, so the key is not reachable through the error path either. This is correct — no changes needed here.

One latent risk worth noting: `from_str(&self.api_key).expect("api key is ascii printable")` will **panic** at runtime if the key contains non-ASCII-printable bytes (e.g. a key accidentally copied with a leading BOM or trailing newline). The fix is to return a `ModelError::InvalidConfig` instead of panicking:

```rust
reqwest::header::HeaderValue::from_str(&self.api_key)
    .map_err(|_| ModelError::InvalidConfig {
        message: "ANTHROPIC_API_KEY contains non-ASCII characters".to_string(),
    })?
```

`auth_headers()` currently returns a plain `HeaderMap` and is called in two places (`chat` and `list_models`). To return an error you would need to change its signature to `fn auth_headers(&self) -> ModelResult<HeaderMap>` or validate the key once at construction time in `with_api_key`.

---

### Translation correctness

**`tool_choice` sent without tools.** `translate_request` serialises `tool_choice` unconditionally if it is set in `ChatRequest`, even when `tools` is `None`. Anthropic returns a 400 if `tool_choice` is present but `tool_use` blocks are absent. The fix:

```rust
let tool_choice = if tools.is_some() { tool_choice.map(tool_choice_to_value) } else { None };
```

**`translate_drops_empty_messages` test merges two empty user messages into zero messages, but verifies `out.messages[0]`.** If both messages are empty the assertion on `out.messages[0].content.len()` will panic. The test body says:

```rust
assert_eq!(out.messages.len(), 1);
assert_eq!(out.messages[0].content.len(), 1);
```

The first message is `ChatMessage::user("")` (empty), the second is `ChatMessage::user("real content")`. After `message_to_blocks` the empty one produces no blocks and is skipped; the second produces one block. But both share role `"user"`, so the merging loop accumulates them under the same `current_role` slot — the result is one message with one block, which is correct. However the test name "drops_empty_messages" implies the empty one is dropped before merging, which is only incidentally true due to role-merging. Add a separate test that starts with a non-empty user message followed by an empty user message to confirm the empty block is not emitted even when role-merging occurs.

**`None` stop\_reason falls through to `Some(FinishReason::Stop)`.** In `translate_response`, the wildcard arm `_ => Some(FinishReason::Stop)` covers both unknown future strings *and* `None`. If Anthropic ever returns a `null` `stop_reason` (e.g. for a streaming stub) this silently maps to `Stop` rather than `None`. The PR description documents this as intentional for "unknown_future" strings, but it is worth an explicit arm:

```rust
None => None,
_ => Some(FinishReason::Stop),
```

---

### Error handling at API boundaries

**`list_models` silently ignores `TOO_MANY_REQUESTS` and `SERVICE_UNAVAILABLE`.** `chat()` handles 429 and 503 explicitly; `list_models` (and by extension `health_check`) does not. A 429 from `GET /v1/models` falls into the generic `Unknown` branch, losing the `RateLimit` variant. Since `health_check` delegates to `list_models`, a transient 429 during startup would be reported as an unknown error rather than a rate-limit, which callers may not back off on. Apply the same status-code guards as in `chat`.

**`serde_json::from_slice::<AnthropicResponse>` on success path returns `ModelError::Serialization` on parse failure.** This is correct per the `From<serde_json::Error>` impl on `ModelError`. No issue — just confirming the conversion is handled.

**`bytes` consumed before status checks complete.** `resp.bytes().await?` is called before any status check. This is fine for correctness, but be aware that for very large error bodies (e.g. a misconfigured proxy returning HTML) the entire body will be buffered. Not a bug, but worth a size limit if defensive posture is needed later.

---

### Missing test cases

1. **`tool_choice` without tools** — no test exercises the `tool_choice` + `tools: None` combination described above, which currently produces an invalid request.
2. **`health_check` error propagation** — `health_check` is not independently tested; a test verifying it surfaces `Authentication` (or `RateLimit`) would complete coverage given the delegation chain.
3. **`auth_headers` panic path** — no test for a key containing a non-ASCII character; after the panic→error fix, add a test for `with_api_key("\u{FEFF}key").chat(...)` (or test `auth_headers` directly).
4. **`tool_call_id` missing on Tool message** — `message_to_blocks` uses `msg.tool_call_id.unwrap_or_default()` which silently emits an empty `tool_use_id` string if a `Tool` message is constructed without a `tool_call_id`. Anthropic will reject this. Add a test with `ChatMessage { role: Tool, tool_call_id: None, ... }` and decide whether to return an error or panic at the translation layer.
5. **Empty `content` on `Tool` message** — similarly, `msg.content.unwrap_or_default()` on a `Tool` message with `content: None` sends an empty string as the tool result. Anthropic accepts empty strings but this may mask bugs upstream. A test would document the expectation.

---

### Feature-flag gate consistency

`model/src/lib.rs` re-exports both `AnthropicProvider` at the crate root and `anthropic::*` via `prelude`, both gated behind `#[cfg(feature = "anthropic")]`. This is consistent. The `anthropic` feature has no dep entries in `Cargo.toml` (intentional since `reqwest` and `serde_json` are already unconditional dependencies), which is fine.

One concern: the `anthropic` module is compiled with `reqwest` unconditionally available because `reqwest` is in `[dependencies]` without a feature gate. If a future PR makes `reqwest` optional, this will break. Adding a comment near the `reqwest` dependency in `Cargo.toml` noting the dependency would clarify intent.

---

### Alignment with PR description

All translation rules listed in the PR description are implemented and unit-tested. The `ToolCall.id` → `tool_use_id` round-trip test (`round_trip_request_then_response_preserves_tool_id_chain`) is present and correct. The `from_env` / `health_check` behaviour matches the "Auth + health check" section. No drift detected between description and code.

The live integration test uses `claude-haiku-4-5-20251001` rather than the `claude-opus-4-7` model used in all other tests and the doc example. This is intentional (cheaper for live runs) but worth a comment so future agents don't mistake it for a copy-paste error.

---

### Summary of actionable items (priority order)

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **Bug** | `translate_request` | `tool_choice` serialised even when `tools` is `None` → Anthropic 400 |
| 2 | **Bug** | `message_to_blocks` / Tool arm | Missing `tool_call_id` silently emits empty `tool_use_id` → Anthropic rejects |
| 3 | **Security/Panic** | `auth_headers` | `.expect()` on key validation will panic on malformed keys; replace with `?` |
| 4 | **Correctness** | `list_models` / `health_check` | 429 / 503 not mapped to typed errors, inconsistent with `chat` |
| 5 | **Correctness** | `translate_response` | `None` stop_reason maps to `Stop` instead of `None` |
| 6 | **Test gap** | tests module | No coverage for items 1, 2, 3 above |

---

---

DominicBurkart:

Review posted with 5 findings: `tool_choice` emitted without `tools` causes Anthropic HTTP 400; `tool_call_id.unwrap_or_default()` on Tool-role messages emits empty `tool_use_id` (Anthropic rejects); `.expect("api key is ascii printable")` panics at runtime on keys with BOM/newline; `health_check` missing 429/503 → `RateLimit`/`ServiceUnavailable` mapping; `None` stop reason incorrectly defaults to `Stop`. No DominicBurkart comments to address; CI is effectively green (only pre-existing macOS/Windows infra failures). **Still in draft**: fix the three bug-class findings (tool_choice/tools guard, empty tool_use_id error, non-ASCII key panic) before opening for human review.

---

Issue #367, opened 2026-05-09
Eval refactor: measure nanna's full application, not just the agent loop excised from it


DominicBurkart:
## Problem

Today's `harness-score` pipeline (PR #363) drives nanna against SWE-bench Lite by **excising the agent loop** from the rest of the application. We give a single-shot prompt to `AgentLoop::run_tool_loop`, capture whatever git diff falls out, and ship that to the upstream Python verifier. Whatever score we land at `evals/scorecards/index.jsonl` is therefore a number about the *agent loop in isolation*, not about nanna-as-deployed.

Two things follow from that:

1. **The agent gets no feedback during the run.** It can call tools (read, edit, search, git) but it has no visibility into whether the test patch is currently failing, whether its edit broke something else, or whether the FAIL_TO_PASS tests are now green. Per ARCHITECTURE.md the architecture is supposed to be feedback-rich; today the score is "throw spaghetti, see what sticks." A 4B-param model editing the test files instead of the source (the actual outcome of the first 5-instance run) is the predictable failure mode.

2. **Most of the production surface area isn't measured.** The score isn't sensitive to MCP, the observability stack, container orchestration, RAG/entity store population, multi-turn behaviour, or anything cross-repo. A regression in those subsystems wouldn't move the metric.

## Where things stand

**Evaluated by `harness-score` today:**
- Agent control loop (`harness::agent::AgentLoop`, `run_tool_loop`) — iteration / decision / tool-call cycle exercised end-to-end on real engineering tasks.
- Tool registry against a materialized real-world repo (`harness::tools::create_tool_registry` with `work_dir` set to the cloned upstream repo) — file reads, edits, search, git ops on actual django/astropy/sympy code.
- Model client wiring (`OllamaProvider`, `model::config::OllamaConfig`) — token streaming, tool-call serialisation, timeout handling.
- System prompt + task framing — whatever prompt nanna sends to the agent is what produces the patch that gets scored.
- Diff capture — the `git add -A && git diff <base_commit>` flow that turns the agent's filesystem mutations into a SWE-bench-shaped `model_patch`.
- Materialize path — clone, checkout, `git apply test_patch`.
- Scoring + persistence — the `harness-score` state machine, JSONL writer, and the `eval-runner` ↔ verifier shellout.

**Not evaluated:**
- Multi-turn context with the user — SWE-bench is a single-prompt task; nanna's interactive UX, follow-up clarifications, and conversational state are untouched.
[...]

PR #371, opened 2026-05-09
feat(eval): drive nanna binary as subprocess for SWE-bench scoring (#367 P0)

DominicBurkart:
## Summary

Today's eval pipeline (`harness::eval::runner::run_eval`) calls `AgentLoop::run_tool_loop()` in-process. This bypasses `harness/src/main.rs` (the user-facing binary), the pod (Ollama containers), MCP, observability, and the install path — so the SWE-bench score in `evals/scorecards/index.jsonl` measures the agent loop excised from nanna, not nanna-as-deployed (issue #367).

This PR is the P0 wiring + container bring-up slice:

- **Eval shells out to the installed `nanna` binary** for SWE-bench cases via `nanna agent --work-dir <repo> --output-json <report>`. Fixture cases keep the in-process path untouched.
- **The binary self-ensures the pod.** `pod::ensure_running` probes `:11434/api/tags`, brings up via `nix run .#start-pod` if down, and short-circuits to `Skipped` when running inside a container (defangs recursion via the install.sh-registered MCP entry).
- **Binary renamed `harness` → `nanna`.** Crate name stays `harness`; renaming it ripples through every `use harness::...` and is out of scope. `nix build .#harness` still works as a deprecated alias.
- **`nanna agent` switches from `AgentLoop::run` to `run_tool_loop`** to align human and eval execution paths on the tool-calling loop. Drops the entity-state-machine path in `run` for `nanna agent` users.
- **Stable JSON contract via `AgentRunReport` DTO** (schema_version 1) insulates the on-disk format from churn in `ChatMessage` / `ToolCallRecord` / `AgentState`.

Explicitly **out of scope**: the test-feedback `run_tests` tool and full MCP-as-deployed scoring from #367 — those are follow-ups.

## Commits

1. `feat(agent): add AgentRunReport DTO for cross-process eval contract`
2. `feat(harness): add pod ensure module with in-container guard`
3. `feat(cli): rename harness → nanna and extend agent for eval-driver use`
4. `refactor(eval): subprocess SWE-bench agent runs through nanna binary`
5. `test(eval): subprocess-runner smoke test gates JSON contract`

## Test plan

- [x] `cargo test --workspace --all-features` — 515 lib + integration tests pass
- [x] End-to-end smoke: `target/release/nanna agent --prompt ... --work-dir /tmp/x --output-json /tmp/r.json` against locally-running Ollama emits valid AgentRunReport (schema_version=1, real token usage)
- [x] `NANNA_HARNESS_BIN=target/release/nanna cargo test --test nanna_subprocess_smoke -- --ignored` passes (subprocess JSON contract gate)
- [ ] Run a single SWE-bench instance through `harness-score` to confirm the subprocess path produces a valid scorecard row (deferred to first CI eval run on this branch)

## Notes for review

- **Behavior change for `nanna agent` users:** the entity-state-machine `AgentLoop::run` path is replaced with `run_tool_loop`. This was a user-confirmed unification.
- **Image OCI name kept as `harness:$TAG`.** Renaming the registry image forces a re-pull for every deployed installation; doing it later is cheaper.
- **`harness-score` stays serial.** Future parallelization will race on the shared pod — flagging here, not fixing in this PR.
- **CI workflow:** `eval.yml` step 5 now passes `NANNA_HARNESS_BIN=$WORKSPACE/target/release/nanna` to the eval runner. CI's existing `ollama serve` makes the binary's pod-probe short-circuit on `AlreadyUp`, so no new bring-up step is needed.

Closes-progress: #367 (P0 subset)


DominicBurkart (commented):
Review of feat(eval): drive nanna binary as subprocess for SWE-bench scoring.

Overall the architecture is sound — the DTO boundary, subprocess dispatch, and pod bring-up guard all make sense. A few correctness and robustness issues below.

---

**[Bug] `bring_up_command` uses `nix run .#start-pod` regardless of CWD**

`pod.rs` prefers `nix run .#start-pod` whenever `nix` is on PATH. That command requires the CWD to contain a `flake.nix`. When `nanna agent` is invoked from an arbitrary `--work-dir` (the exact eval case), `nix run .#start-pod` will fail with "flake not found". The `podman` fallback only fires when `NANNA_POD_CONFIG` is set, so on a dev box with nix but wrong CWD, bring-up fails silently. Consider checking for `flake.nix` in the repo root before preferring the nix path, or always passing the absolute flake path.

**[Dead code] `SubprocessNonZero` variant is never emitted**

`EvalRunnerError::SubprocessNonZero` is defined but the non-zero exit path returns `Ok(Err(format!(...)))` instead. Either remove the variant or use it.

**[Bug / confusing error] `locate_nanna_binary` accepts a directory path**

`runner.rs`: `path.exists()` accepts any path, including directories. If `NANNA_HARNESS_BIN` points to a directory the code returns `Ok(dir_path)` and the subsequent `cmd.output()` will fail with a cryptic OS error instead of `BinaryNotFound`. Add `path.is_file()` (and optionally an executable-bit check).

**[Blocking sleep in async context] `wait_for_ollama` uses `std::thread::sleep`**

`pod.rs`: `ensure_pod_or_exit` is called from within `async fn main` on the Tokio executor thread. `std::thread::sleep` inside `wait_for_ollama` stalls the executor for up to 60 s. Use `tokio::time::sleep` (make `ensure_running` async, or move the call into `tokio::task::spawn_blocking`).

**[Minor] `cmd.output()` buffers unbounded subprocess output**

`run_agent_subprocess` in `runner.rs` calls `cmd.output()` which accumulates all subprocess stdout+stderr in memory. A verbose long-running SWE-bench agent can produce large output. Consider piping stderr to a size-bounded sink or a temp file, especially since the eval runner can parallelise later.

**[Correctness note] `--tools` is unconditionally appended for subprocess runs**

`.arg("--tools")` is always passed to the subprocess; there is no way to run a SWE-bench subprocess without tools. This is likely intentional but worth documenting in the comment, since the `tools: bool` flag exists in the CLI for other purposes and its presence/absence changes agent behaviour.

---

DominicBurkart:

Review posted with 6 findings: `std::thread::sleep` in async pod `wait_for_ollama` stalls the Tokio executor up to 60s; `locate_nanna_binary` accepts directories (confusing spawn error); `bring_up_command` requires `flake.nix` in CWD which won't hold for arbitrary `--work-dir` eval runs; dead `SubprocessNonZero` variant; unbounded `cmd.output()` buffers full subprocess stdout+stderr. CI was unblocked by pinning `home = "0.5.11"` in `Cargo.lock` to resolve an MSRV 1.88 conflict with the Nix-pinned rustc 1.87 — CI run in progress. **Still in draft**: address the async sleep, binary-location, and CWD bugs before opening for human review.

---

DominicBurkart:
[job:RlQo0][oldest-issue:#367]
head-sha-before: 386a546cd6b1c605f36f3ec0c8c857802fd617c4
head-sha-after: 68376d49344518f3a0f6000f2b915c126a7ab8f3
status: iterated
actions:
- pod.rs: made ensure_running / probe_ollama / wait_for_ollama async; switched the post-bring-up sleep from std::thread::sleep to tokio::time::sleep so the 60s health wait no longer stalls the Tokio executor.
- pod.rs: bring_up_command now requires flake.nix to be present in flake_lookup_dir before preferring the nix path. Defaults to cwd; falls back to NANNA_POD_CONFIG/podman or NoBringUpCommand. Closes the "nanna agent --work-dir <arbitrary>" failure mode flagged in the prior review.
- pod.rs: added 6 new unit tests (wait_for_ollama timeout-respect, bring_up_command flake gate both branches, PodError Display surface, is_inside_container exercised) to lift codecov/patch on the previously uncovered file.
- main.rs: ensure_pod_or_exit is now async; the three call sites in nanna {chat,health,agent} `.await` it.
- runner.rs: locate_nanna_binary now rejects directories via path.is_file() so a directory NANNA_HARNESS_BIN maps to BinaryNotFound instead of cryptic OS spawn error.
- runner.rs: dropped unused EvalRunnerError::SubprocessNonZero variant.
- runner.rs: documented the unconditional `--tools` append for SWE-bench subprocess runs.
- runner.rs: added agent_outcome_subprocess_token_usage_propagates unit test covering the AgentOutcome::Subprocess branches.
remaining-work:
- Unbounded subprocess output buffering (review item E) — still uses cmd.output() which buffers stdout+stderr in memory. Deferred: switching to a bounded sink / temp file is meaningful for parallel scoring but not load-bearing for the current serial harness-score path.
- Pre-existing CI failures on Build aarch64-darwin / x86_64-darwin, macOS full bring-up, Windows full bring-up appear unrelated to this PR's diff (the latter two are platform bring-up smoke tests that are expected to fail off-Linux). Linux full bring-up passes and is the supported path for this branch.
- Verifying CI green after these commits is the next CRON pass; this iteration is not flipping the PR to ready-for-review (per janitor protocol).

---

DominicBurkart (commented):
Review of the updated HEAD (368470162f82) against the previous review (0a6babdb). All six issues from the earlier review are addressed — the flake.nix guard, `is_file()` check, async `tokio::time::sleep`, dead `SubprocessNonZero` variant, unbounded stderr buffer note, and `--tools` documentation are all resolved. New issues below.

---

**[Bug — correctness] `__nanna_agent_report.json` left in `work_dir` pollutes the SWE-bench patch**

`run_agent_subprocess` writes the report to `work_dir.join("__nanna_agent_report.json")` and never deletes it. `capture_swebench_patch` then runs `git add -A && git diff <base>` against `work_dir`, so the report JSON will appear in the captured patch submitted to the verifier. This is a real scoring bug: the patch sent to the SWE-bench verifier will contain an extraneous file the agent never intentionally created. Fix: delete the file after a successful read, or write it to a path outside `work_dir` (e.g. `tempfile::NamedTempFile` or `std::env::temp_dir()`).

---

**[Bug] `schema_version` is deserialized but never validated**

`AgentRunReport::read_from_path` deserialises the JSON without checking `parsed.schema_version == SCHEMA_VERSION`. If a future binary writes `schema_version: 2` with a changed shape, serde will happily parse it (all current fields present/compatible), the runner will silently consume a mismatched report, and the only symptom will be subtly wrong eval metrics. Add a version check after deserialisation and return `InvalidData` on mismatch:

```rust
let report: Self = serde_json::from_slice(&bytes)
    .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
if report.schema_version != SCHEMA_VERSION {
    return Err(io::Error::new(
        io::ErrorKind::InvalidData,
        format!(
            "AgentRunReport schema_version mismatch: expected {}, got {}",
            SCHEMA_VERSION, report.schema_version
        ),
    ));
}
Ok(report)
```

---

**[Soundness] `std::env::set_var` in async tests is `unsafe` since Rust 1.81**

`with_nanna_bin` and the `locate_nanna_binary_falls_back_to_which_when_env_unset` test both call `std::env::set_var` / `std::env::remove_var` from within `#[tokio::test] async` functions on the multi-threaded Tokio executor. Since Rust 1.81, `std::env::set_var` is `unsafe` in a multithreaded context because concurrent reads of the environment (in another Tokio task or a spawned thread) can race with the write. `#[serial_test::serial]` serialises the tests against each other but does not suppress the race against Tokio's internal worker threads, which may read env vars at any time.

The same hazard applies to the PATH mutation in `locate_nanna_binary_falls_back_to_which_when_env_unset`, which is an async Tokio test. The full end-to-end `run_eval_swebench_end_to_end_with_local_repo_and_fake_nanna` test mutates three env vars then awaits `run_eval` — the period between `set_var` and the await point is especially exposed.

Options:
1. Make env-manipulating tests `#[test]` (sync) and use `tokio::runtime::Builder::new_current_thread().build().unwrap().block_on(...)` for any async portion — a freshly built single-threaded runtime has no concurrent env readers.
2. Use `#[tokio::test(flavor = "current_thread")]` plus `unsafe { std::env::set_var(...) }` with an explicit safety comment (valid only with `current_thread` runtime and `serial_test`).

---

**[Minor] `AgentRunReport::read_from_path` performs blocking I/O on the async executor**

`run_agent_subprocess` is `async` and is awaited on the Tokio executor. It calls `AgentRunReport::read_from_path`, which uses `std::fs::read` — blocking I/O on the executor thread. For a small JSON file this is unlikely to cause latency in practice, but it is inconsistent with the async-first design. Consider `tokio::fs::read` or wrapping in `tokio::task::spawn_blocking`.

---

**[Minor] PR #374 merge-order dependency**

PR #374 Part 3 touches `harness/src/main.rs` (`build_session_system_prompt` consolidation). This PR also modifies `harness/src/main.rs` substantially. If #371 merges first, #374's branch will need a rebase. Worth noting explicitly in the merge plan.

---

**Summary**

Architecture is sound and all previous review concerns are resolved. Blocking items before merge:
1. Report JSON left in `work_dir` will corrupt SWE-bench patch scores.
2. Schema version not validated on read — silent mismatches on future schema bumps.

Should fix before merge: async `set_var` soundness (currently MIRI/clippy-flaggable, will become a hard `unsafe` compile error in a future edition).

Advisory: blocking `fs::read` in async context, #374 merge order.

https://claude.ai/code/session_01LkNk2Vcsp4y5PyoNp52mMC

---

DominicBurkart:
**Status:** Critical bug fixed + code review posted — `__nanna_agent_report.json` was written to `work_dir` and never deleted before `capture_swebench_patch` ran `git add -A`, which would have included the harness artefact in every SWE-bench patch and corrupted scores; deleted immediately after reading (commit 4e3897e). Branch is behind main but **cannot be auto-rebased** because it touches `.github/workflows/` files, which require `workflows` token scope — manual rebase by a human is needed. Two additional blocking findings from code review remain open: `AgentRunReport::read_from_path` lacks `schema_version` validation, and `std::env::set_var`/`remove_var` are called from `#[tokio::test]` async functions (unsafe since Rust 1.81).

https://claude.ai/code/session_01LkNk2Vcsp4y5PyoNp52mMC

---

PR #373, opened 2026-05-10
test: fill largest coverage gaps in observability.rs and monitoring.rs

DominicBurkart:
## Summary

`harness/src/observability.rs` (35 KB, only 5 tests) and `harness/src/monitoring.rs` had many pure methods and enum variants with no coverage. This PR targets the two largest gaps.

**observability.rs — 22 new tests**
- `determine_alert_category`: all 5 dispatch branches (container, model, performance, resource, availability)
- `calculate_priority_score`: severity base values + category bonuses + cap-at-100 behaviour
- `generate_recommended_actions`: all 4 category arms, verifying action count and keywords
- `analyze_current_trends`: stable baseline (score=100), degrading latency (−20), degrading error rate (−25), degrading cache (−15)
- Enum variant coverage: `SlaStatus`, `EscalationStatus`, `AlertCategory` (all 7), `ChannelType` (all 6)
- `AlertPolicy::immediate_critical()` and `balanced()` field-level assertions
- `HealthThreshold::default()` exact field values

**monitoring.rs — 10 new tests**
- `HealthStatus::is_healthy()` and `requires_attention()` for every variant
- `AlertSeverity` and `ErrorSeverity` ordering (derived `Ord`)
- `AlertThresholds::default()` exact field values
- `MetricsFormat::Custom` export returns an error containing the format name
- `acknowledge_alert` on a nonexistent id returns `Err`
- Fresh collector: cache hit rate is `0.0` when no events recorded
- `reset_metrics()` clears latencies, hits, misses, and resets hit rate to `0.0`
- `ContainerStatus` variant distinctness

---
_Generated by [Claude Code](https://claude.ai/code/session_01GCs1QCJZ62VYoPSWV8m89r)_

### DIFF
--- a/harness/src/monitoring.rs
+++ b/harness/src/monitoring.rs
@@ -1242,4 +1242,108 @@ mod tests {
         let csv_export = collector.export_metrics(MetricsFormat::Csv).await.unwrap();
         assert!(csv_export.contains("timestamp,metric_type,service,value"));
     }
+
+    #[test]
+    fn test_health_status_is_healthy() {
+        assert!(HealthStatus::Healthy.is_healthy());
+        assert!(!HealthStatus::Warning.is_healthy());
+        assert!(!HealthStatus::Degraded.is_healthy());
+        assert!(!HealthStatus::Unhealthy.is_healthy());
+        assert!(!HealthStatus::Unknown.is_healthy());
+    }
+
+    #[test]
+    fn test_health_status_requires_attention() {
+        assert!(!HealthStatus::Healthy.requires_attention());
+        assert!(HealthStatus::Warning.requires_attention());
+        assert!(HealthStatus::Degraded.requires_attention());
+        assert!(HealthStatus::Unhealthy.requires_attention());
+        assert!(!HealthStatus::Unknown.requires_attention());
+    }
+
+    #[test]
+    fn test_alert_severity_ordering() {
+        assert!(AlertSeverity::Info < AlertSeverity::Warning);
+        assert!(AlertSeverity::Warning < AlertSeverity::Error);
+        assert!(AlertSeverity::Error < AlertSeverity::Critical);
+    }
+
+    #[test]
+    fn test_error_severity_ordering() {
+        assert!(ErrorSeverity::Info < ErrorSeverity::Warning);
+        assert!(ErrorSeverity::Warning < ErrorSeverity::Error);
+        assert!(ErrorSeverity::Error < ErrorSeverity::Critical);
+    }
+
+    #[test]
+    fn test_alert_thresholds_default() {
+        let thresholds = AlertThresholds::default();
+        assert_eq!(thresholds.max_latency_ms, 5000);
+        assert_eq!(thresholds.min_cache_hit_rate, 0.8);
+        assert_eq!(thresholds.max_error_rate, 0.05);
+        assert_eq!(thresholds.max_cpu_usage, 0.9);
+        assert_eq!(thresholds.max_memory_usage, 0.9);
+        assert_eq!(thresholds.health_check_timeout, Duration::from_secs(30));
+    }
+
+    #[tokio::test]
+    async fn test_metrics_export_custom_format_returns_error() {
+        let mut collector = DefaultMetricsCollector::new();
+        collector
+            .record_request_latency("test", Duration::from_millis(100))
+            .await;
+        let result = collector
+            .export_metrics(MetricsFormat::Custom("myformat".to_string()))
+            .await;
+        assert!(result.is_err());
+        let err_msg = result.unwrap_err().to_string();
+        assert!(err_msg.contains("myformat"));
+    }
+
+    #[tokio::test]
+    async fn test_acknowledge_nonexistent_alert_returns_error() {
+        let manager = DefaultAlertManager::new();
+        let result = manager.acknowledge_alert("nonexistent-id").await;
+        assert!(result.is_err());
+    }
+
+    #[tokio::test]
+    async fn test_cache_metrics_from_scratch() {
+        let collector = DefaultMetricsCollector::new();
+        let metrics = collector.get_current_metrics().await.unwrap();
+        assert_eq!(metrics.cache_metrics.hits, 0);
+        assert_eq!(metrics.cache_metrics.misses, 0);
+        assert_eq!(metrics.cache_metrics.hit_rate, 0.0);
+    }
+
+    #[tokio::test]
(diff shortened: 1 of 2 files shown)


DominicBurkart (commented):
## Code Review — PR #373 vs PR #375

### Summary

This PR adds 32 new tests across two files (22 in `observability.rs`, 10 in `monitoring.rs`) and is **complementary to, not superseded by, PR #375**. However the overlap in `monitoring.rs` is significant enough that merging both as-is would produce duplicate tests. The right path is to land #375 first (or simultaneously), then either close the `monitoring.rs` portion of #373 or strip those 10 tests from this branch, keeping only the 22 `observability.rs` tests.

---

### observability.rs tests (22 tests) — this file is unique to #373

**Verdict: high value, keep all of them.**

The `determine_alert_category` / `calculate_priority_score` / `generate_recommended_actions` / `analyze_current_trends` tests pin concrete dispatch logic, scoring arithmetic, and score-capping behavior that would otherwise be entirely invisible to the test suite. The `make_stable_metrics()` / `make_alert()` helpers are clean and the score arithmetic is documented inline (`// Critical=100, +20 for Availability = 120, capped at 100`).

A few notes on individual tests:

**Fragility risk — hardcoded priority score arithmetic**

The `calculate_priority_score` tests (`test_calculate_priority_score_critical_availability_capped`, `test_calculate_priority_score_error_performance`, etc.) assert exact integer scores (100, 85, 55, 70 …) derived from knowledge of internal base-value constants. If those constants are ever tuned the tests will fail with no indication that they *should* still be the canonical values. Consider adding a brief comment stating where the constants come from (e.g. `// see calculate_priority_score: Critical base=100`) so a future maintainer knows whether a failing test means "the constant changed intentionally" or "there is a regression."

**Fragility risk — generate_recommended_actions only asserts `len == 3`**

`test_generate_recommended_actions_other_categories` lumps Availability, Security, Configuration, and Resources into one test and checks only that each returns exactly 3 actions. If any arm is accidentally changed to return 2 or 4 actions the test catches it, but if the *content* regresses (e.g. actions become empty strings) it won't. The ContainerHealth and Performance arms at least also assert a keyword. It would be worth asserting at least one keyword per arm — even a non-empty first element — for the remaining four categories.

**Enum distinctness tests (SlaStatus, EscalationStatus, AlertCategory, ChannelType)**

These are unit-testing that `#[derive(PartialEq)]` works, which is already guaranteed by the derive macro. They have near-zero regression value and take up ~50 lines. I'd recommend dropping them or replacing with a comment like `// PartialEq and Eq are derived; no manual impl to test.`

**AlertPolicy structure tests — coupling risk**

`test_alert_policy_immediate_critical_structure` and `test_alert_policy_balanced_structure` assert field-by-field layout (lengths, exact `Duration::from_secs` values, channel types, group names). This is useful for catching accidental changes to the default policy, but it tightly couples tests to implementation details. Consider adding a comment clarifying that these tests are intentional snapshot assertions meant to catch drift in default policy values.

---

### monitoring.rs tests (10 tests) — overlap with PR #375

**Verdict: these 10 tests duplicate what #375 already covers, and in most cases #375's version is strictly superior.**

| Test in #373 | Counterpart in #375 | Verdict |
|---|---|---|
| `test_health_status_is_healthy` + `test_health_status_requires_attention` | `health_status_classification_is_partitioned` | #375 is strictly superior — it uses a loop and explicitly checks mutual exclusivity; #373's two separate tests pass even if both return `true` for `Warning`. |
| `test_metrics_export_custom_format_returns_error` | `custom_metrics_format_is_rejected` | #375 is superior — it uses `matches!` to verify the exact `MonitoringError` variant; #373 only checks `.to_string()` contains the format name. |
| `test_acknowledge_nonexistent_alert_returns_error` | `alert_lifecycle_history_and_ack_errors` | #375 is superior — it also verifies the full ack lifecycle (active list, history, limit). |
| `test_cache_metrics_from_scratch` | `cache_hit_rate_matches_hits_over_total` | #375 is superior — it also checks the math and `!is_nan`. |
| `test_reset_metrics_clears_all_data` | `reset_metrics_clears_all_state` | #375 is superior — it also checks `error_metrics` are cleared. |
| `test_alert_severity_ordering` | *(not in #375)* | Unique to #373 — keep. |
| `test_error_severity_ordering` | *(not in #375)* | Unique to #373 — keep. |
| `test_alert_thresholds_default` | *(not in #375)* | Unique to #373 — keep. |
| `test_container_status_variants_distinct` | *(not in #375)* | Same concern as enum distinctness tests above. Marginal value. |

**What #375 has that #373 does not (all in monitoring.rs):**
- Latency percentile ordering invariant (`min ≤ avg ≤ max`, `p95 ≤ p99`)
- Empty latency input → all-zero / NaN-free
- Error accounting internals (error rate formula, `errors_by_type` sum, recent_errors capped at 10 + most-recent-first ordering)
- Alert unique-ID uniqueness invariant
- Alert history most-recent-first ordering and `limit` parameter
- Prometheus/JSON export round-trip

These are materially more valuable than the duplicated tests in this PR.

---

### Recommendation

**Do not close #373; do not close #375. Recommended action:**

1. Land #375 first (it is more thorough on `monitoring.rs`).
2. Strip the 10 `monitoring.rs` tests from this branch **except** `test_alert_severity_ordering`, `test_error_severity_ordering`, and `test_alert_thresholds_default`, which are unique and worth keeping.
3. Drop the pure enum-distinctness tests (`test_sla_status_variants`, `test_escalation_status_variants`, `test_alert_category_all_variants_distinct`, `test_channel_type_all_variants_distinct`) from `observability.rs` — they test the derive macro, not the application logic.
4. Strengthen `test_generate_recommended_actions_other_categories` with at least one keyword assertion per arm.
5. Add comments on the hardcoded priority-score tests.

After those changes, #373 becomes a clean, non-overlapping complement to #375 focused entirely on `observability.rs` logic and three `monitoring.rs` ordering/threshold snapshot tests.

**Ready verdict: NOT READY — draft status is correct. Needs scoping changes before merge.**

https://claude.ai/code/session_01LkNk2Vcsp4y5PyoNp52mMC

---

DominicBurkart:
**Status:** Monitoring-redundancy conflict — this PR touches `harness/src/monitoring.rs`, the same file being changed in PRs #354, #355, #374, and #375; merge order must be made explicit and each subsequent PR rebased in sequence before promotion. No blocking code issues found in this PR's diff itself. Establish the consolidation order before any of these five PRs advances.

https://claude.ai/code/session_01LkNk2Vcsp4y5PyoNp52mMC

---

PR #389, opened 2026-05-14
test(monitoring): cover metric math + alert lifecycle; fix CONTRIBUTING license link

DominicBurkart:
Automated maintenance pass. Findings are organized below; the actual diff is intentionally small and focused.

## Testing

`harness/src/monitoring.rs` was the most undertested core component: ~1245 lines with only 5 happy-path tests. The default `MetricsCollector` / `HealthMonitor` / `AlertManager` implementations had no coverage of their math or error paths.

Added 14 unit tests documenting and pinning key invariants:

- **`HealthStatus`** predicates — `is_healthy` / `requires_attention` for every variant (incl. `Unknown` being neither).
- **`calculate_latency_metrics`** — empty-slice yields an all-zero summary (no panic on empty-slice math); min/max/avg correctness; percentiles bounded by observed min/max.
- **Division-by-zero guards** — cache hit-rate and error-rate are `0.0` when there is no traffic.
- **Error aggregation** — `errors_by_type` grouping, `total_errors`, `recent_errors`; `reset_metrics` clears state.
- **`record_model_inference`** round-trips into `get_current_metrics`.
- **`MetricsFormat::Custom`** returns an unsupported-format error.
- **Alert lifecycle** — acknowledging an unknown id errors; `get_alert_history` is bounded and newest-first; alert IDs are unique; `configure_thresholds` overrides defaults; `AlertThresholds::default` values.
- **Severity ordering** — `AlertSeverity` / `ErrorSeverity` `Ord` used for escalation.

All 14 new tests pass via `cargo test -p harness --lib monitoring` (19/19 in module). `cargo fmt` and `cargo clippy --lib --all-features` are clean.

Note: 3 pre-existing failures in `eval::swebench::tests::materialize_from_url_*` reproduce on the base branch — git2 `revparse_single("HEAD")` fails in this sandbox environment; unrelated to this change and expected to pass in the Nix devShell CI.

## Clarity

- `CONTRIBUTING.md` linked to `./LICENSE`, which does not exist. The repo ships `LICENSE-MIT` and `LICENSE-APACHE`. Fixed the link to point at both actual files.

## Architecture

Reviewed `ARCHITECTURE.md`, `AGENTS.md`, `TESTING.md`, `CONTRIBUTING.md`, and `.github/workflows/codecov-guard.yml` against the code. No clear divergence found: the six CLI subcommands and six MCP tools described in `ARCHITECTURE.md` match `harness/src/main.rs`, and the codecov-guard workflow behaves as `AGENTS.md` documents. No changes made here — flagging that nothing warranted churn rather than inventing some.

---
_Generated by [Claude Code](https://claude.ai/code/session_01C18s7J23s2b9R64jzfKQAb)_

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ See [TESTING.md](./TESTING.md).
 
 ## License
 
-All contributions fall under the [project license](./LICENSE).
+All contributions fall under the project license ([MIT](./LICENSE-MIT) or [Apache-2.0](./LICENSE-APACHE)).
 
 ## GitHub Issues
 
--- a/harness/src/monitoring.rs
+++ b/harness/src/monitoring.rs
@@ -1242,4 +1242,228 @@ mod tests {
         let csv_export = collector.export_metrics(MetricsFormat::Csv).await.unwrap();
         assert!(csv_export.contains("timestamp,metric_type,service,value"));
     }
+
+    // --- HealthStatus predicates ---
+
+    #[test]
+    fn test_health_status_predicates() {
+        assert!(HealthStatus::Healthy.is_healthy());
+        assert!(!HealthStatus::Warning.is_healthy());
+        assert!(!HealthStatus::Healthy.requires_attention());
+        for s in [
+            HealthStatus::Warning,
+            HealthStatus::Degraded,
+            HealthStatus::Unhealthy,
+        ] {
+            assert!(s.requires_attention(), "{s:?} should require attention");
+        }
+        // Unknown is neither healthy nor a definite problem.
+        assert!(!HealthStatus::Unknown.is_healthy());
+        assert!(!HealthStatus::Unknown.requires_attention());
+    }
+
+    // --- latency-metric math ---
+
+    #[tokio::test]
+    async fn test_latency_metrics_empty_is_zeroed() {
+        // A service with no recorded latencies must yield an all-zero summary
+        // rather than panicking on empty-slice math.
+        let collector = DefaultMetricsCollector::new();
+        let metrics = collector.calculate_latency_metrics(&[]);
+        assert_eq!(metrics.request_count, 0);
+        assert_eq!(metrics.avg_latency_ms, 0.0);
+        assert_eq!(metrics.p95_latency_ms, 0.0);
+        assert_eq!(metrics.max_latency_ms, 0.0);
+        assert_eq!(metrics.min_latency_ms, 0.0);
+    }
+
+    #[tokio::test]
+    async fn test_latency_metrics_min_max_avg() {
+        let collector = DefaultMetricsCollector::new();
+        let samples: Vec<Duration> = [10u64, 20, 30, 40]
+            .into_iter()
+            .map(Duration::from_millis)
+            .collect();
+        let m = collector.calculate_latency_metrics(&samples);
+        assert_eq!(m.request_count, 4);
+        assert_eq!(m.min_latency_ms, 10.0);
+        assert_eq!(m.max_latency_ms, 40.0);
+        assert_eq!(m.avg_latency_ms, 25.0);
+        // Percentiles are bounded by the observed min/max.
+        assert!(m.p95_latency_ms >= m.min_latency_ms && m.p95_latency_ms <= m.max_latency_ms);
+        assert!(m.p99_latency_ms >= m.min_latency_ms && m.p99_latency_ms <= m.max_latency_ms);
+    }
+
+    // --- cache-metric math ---
+
+    #[tokio::test]
+    async fn test_cache_hit_rate_zero_when_no_traffic() {
+        // Division-by-zero guard: an untouched collector reports a 0.0 hit rate.
+        let collector = DefaultMetricsCollector::new();
+        let metrics = collector.get_current_metrics().await.unwrap();
+        assert_eq!(metrics.cache_metrics.hits, 0);
+        assert_eq!(metrics.cache_metrics.misses, 0);
+        assert_eq!(metrics.cache_metrics.hit_rate, 0.0);
+    }
+
+    // --- error aggregation ---
+
+    fn error_event(error_type: &str) -> ErrorEvent {
+        ErrorEvent {
+            timestamp: Utc::now(),
+            error_type: error_type.to_string(),
+            message: "boom".to_string(),
+            component: "test".to_string(),
+            severity: ErrorSeverity::Error,
+        }
+    }
+
+    #[tokio::test]
+    async fn test_error_metrics_group_by_type() {
+        let mut collector = DefaultMetricsCollector::new();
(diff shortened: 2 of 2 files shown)


DominicBurkart (commented):
## Review: test(monitoring) — 14 new unit tests + CONTRIBUTING license fix

Overall the test suite is a solid addition. Coverage of the guard branches is real (not just smoke-tested), the alert-lifecycle sequence is logically sound, and the CONTRIBUTING fix is correct. A handful of issues are worth addressing before merge, ranging from one possible correctness bug to a few coverage gaps that matter for the module's invariants.

---

### CONTRIBUTING.md — license link fix

Both `LICENSE-MIT` and `LICENSE-APACHE` exist at the repo root (verified). The relative paths `./LICENSE-MIT` and `./LICENSE-APACHE` resolve correctly from `CONTRIBUTING.md`. Fix is accurate; the old `./LICENSE` link was broken.

---

### Issues and observations

#### 1. `test_latency_metrics_min_max_avg` — p95/p99 bounds assertion is under-constrained (nit → potential correctness gap)

```rust
assert!(m.p95_latency_ms >= m.min_latency_ms && m.p95_latency_ms <= m.max_latency_ms);
assert!(m.p99_latency_ms >= m.min_latency_ms && m.p99_latency_ms <= m.max_latency_ms);
```

With only 4 samples, `p95_index = (4 * 0.95) as usize = 3` and `p99_index = (4 * 0.99) as usize = 3`, so both percentiles resolve to `sorted_latencies[3] = 40.0`. The bounds check passes vacuously — it would also pass if the implementation returned `min_latency_ms` for both. Consider adding a concrete assertion for the known value:

```rust
assert_eq!(m.p95_latency_ms, 40.0);
assert_eq!(m.p99_latency_ms, 40.0);
```

This pins the actual indexing arithmetic rather than just the range.

#### 2. `test_configure_thresholds_overrides_defaults` — only one field verified

```rust
manager.configure_thresholds(custom.clone()).await.unwrap();
assert_eq!(manager.thresholds.max_latency_ms, 1);
```

`configure_thresholds` does a full struct replacement (`self.thresholds = thresholds`), so all fields are overwritten simultaneously. Checking only `max_latency_ms` leaves the other five fields unverified. A future refactor that accidentally dropped some fields during assignment (e.g. a partial update) would not be caught. Adding spot-checks for at least one or two more distinct fields (e.g. `min_cache_hit_rate`, `health_check_timeout`) would make the test meaningfully exercise the override semantics.

#### 3. `test_alert_history_is_bounded_and_newest_first` — "bounded" claim is not actually tested

The test name says "bounded" but only sends 3 alerts and requests `limit=2`. This confirms the limit is honoured for a list smaller than the internal cap, but never exercises what happens when the history itself has more entries than the limit. The current `get_alert_history` implementation (`alerts.iter().rev().take(limit)`) is correct, but the test does not distinguish it from one that returned all alerts up to the limit regardless of internal size. Not blocking, but worth noting.

#### 4. Missing: cache hit-rate with mixed traffic (positive-path guard)

`test_cache_hit_rate_zero_when_no_traffic` pins the zero-division guard. There is no corresponding test that records a mix of hits and misses and asserts the resulting `hit_rate` is correct (e.g. 2 hits + 1 miss → `0.666…`). The existing `test_metrics_collector` does check `hit_rate == 0.5`, so the positive path is covered indirectly, but an explicit targeted test would be cleaner for this module.

#### 5. Missing: error-rate non-zero path

`test_error_metrics_group_by_type` correctly asserts `error_rate == 0.0` when no requests have been recorded. There is no test that records both errors and requests and checks that `error_rate = total_errors / total_requests` computes correctly. This is the primary guard against a regression in the error-rate calculation.

#### 6. `test_latency_metrics_empty_is_zeroed` is marked `async` but does not need to be

```rust
#[tokio::test]
async fn test_latency_metrics_empty_is_zeroed() {
```

`calculate_latency_metrics` is a synchronous method (no `.await` inside). Marking it `#[tokio::test] async` works but adds unnecessary overhead and could mislead future readers about the nature of the function. `#[test] fn` is sufficient, same as `test_health_status_predicates`.

#### 7. `test_configure_thresholds_overrides_defaults` — `custom.clone()` is unnecessary

```rust
manager.configure_thresholds(custom.clone()).await.unwrap();
assert_eq!(manager.thresholds.max_latency_ms, 1);
```

`custom` is never used after the call; `.clone()` is dead. Pass `custom` directly.

---

### No issues found

- **Division-by-zero guards**: Both the cache hit-rate path (`get_cache_metrics`: `if total > 0`) and the error-rate path (`get_current_metrics`: `if total_requests > 0`) are exercised by the tests.
- **Alert ID uniqueness**: `generate_alert_id` uses a monotonic counter; `test_alert_ids_are_unique` correctly exercises two sequential calls.
- **`AlertSeverity` / `ErrorSeverity` `Ord`**: Both derive `PartialOrd, Ord` via the declaration order, so `Info < Warning < Error < Critical` is guaranteed by the compiler. The test correctly locks in this ordering as a regression guard.
- **`test_export_custom_format_is_unsupported`**: Correctly maps to `MonitoringError::MetricsCollectionFailed`.
- **`test_acknowledge_unknown_alert_errors`**: Correctly maps to `MonitoringError::AlertSendFailed`.
- **`test_reset_metrics_clears_state`**: Correctly resets both cache and error state.
- **No PII, secrets, or security issues** in the added code.
- **`mergeable_state=unstable`**: The Windows/macOS failures are infrastructure-level and pre-existing; they do not originate from this PR.

---

### Summary

| # | Severity | Item |
|---|----------|------|
| 1 | Suggestion | Pin concrete percentile values in `test_latency_metrics_min_max_avg` |
| 2 | Suggestion | Verify more fields in `test_configure_thresholds_overrides_defaults` |
| 3 | Nit | "bounded" in `test_alert_history_is_bounded_and_newest_first` is untested |
| 4 | Nit | Add explicit positive-path cache hit-rate test |
| 5 | Nit | Add error-rate non-zero path test |
| 6 | Nit | `test_latency_metrics_empty_is_zeroed` should be `#[test]` not `#[tokio::test]` |
| 7 | Nit | Remove unnecessary `.clone()` in `test_configure_thresholds_overrides_defaults` |

None of these are blocking correctness issues in the production code — they are gaps in what the tests actually assert. Items 1 and 2 are the most worth fixing before merge; the rest are polish.

---

DominicBurkart:
**Cron status (session dVfNL):** Code review posted — 14 new monitoring tests verified correct with minor nits (percentile assertion specificity, bounded-history coverage, unnecessary `async` on a sync test); CONTRIBUTING.md license link fix confirmed against actual `LICENSE-MIT`/`LICENSE-APACHE` files. No owner comments to address; "All Checks Passed" gate confirmed `success` — Windows/macOS bring-up failures are explicitly non-required per workflow configuration. PR promoted from draft to ready-for-review.

---

